PR #88, opened 2022-06-11
chore: add feature all to fix rust ls

tjjfvi:


harrysolovay:
@tjjfvi is this preferable to specifying `"rust-analyzer.cargo.features": ["all"]` in the `.vscode/settings.json`? I suppose this solves the problem for non-VSC devs, so the answer is likely yes. If so, please remove the setting from `.vscode/settings.json`.

tjjfvi:
@harrysolovay It is not an alternative to that – the setting of `["all"]` does not work without a corresponding feature `all` (at least on my machine)

harrysolovay (approved):

PR #91, opened 2022-06-11
docs: treatment of types

### LINKED ISSUE #74: Document Type Conversions Between Rust & JS Env
The work of #73 solidifies our approach to converting Rust-defined data types to their JS counterparts and vice versa. Developers will—in many cases—need to manually input data (with JS) that correctly encodes to the on-chain world’s expectations. This is difficult without first understanding our approach to conversion.

TODO: conversation documentation

harrysolovay:
Resolved #74.

Capi does [some very interesting conversions](https://github.com/paritytech/capi/blob/d108e2dda4e0e525b54a96f432f00b00f216d24d/frame_metadata/Codec.ts) between the Rust and JS env. While––ideally––most Rust&TS-familiar devs will be able to intuit the conversion, we should provide explicit documentation. This will also serve as a guide for determining how to craft interactions per-chain. This is important given that we DO NOT want to manually type out / ship with too many hard-coded factories (as they can vary between chains).


tjjfvi (approved):
Still rather WIP, but I think this is good to merge for now.

PR #111, opened 2022-06-14
chore: rework branded types

### LINKED ISSUE #96: Approach to Branded Types & RPC-related Validation
Do `U.HexString` and comparable branded types buy us much in terms of safety / DX?

harrysolovay:
Resolves #96 –– mostly taken from the exceptional recommendations of @tjjfvi

### DIFF
--- a/primitives/AccountId.ts
+++ b/primitives/AccountId.ts
@@ -1,0 +1,2 @@
+// Any need to encode size of 32 bytes?
+export class AccountId extends Uint8Array {}
--- a/rpc/messages.ts
+++ b/rpc/messages.ts
@@ -1,3 +1,4 @@
+import * as P from "../primitives/mod.ts";
 import { EnsureLookup } from "../util/mod.ts";
 import * as U from "../util/mod.ts";
 import * as T from "./types/mod.ts";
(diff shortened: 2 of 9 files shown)


primitives/AccountId.ts
```diff
@@ -0,0 +1,2 @@
+// Any need to encode size of 32 bytes?
+export class AccountId extends Uint8Array {}
```
tjjfvi, line 2: This should be a branded type, not a real run-time class. A class implies the validity of e.g. `instanceof` checks, which is not the case here.

rpc/messages.ts
```diff
@@ -231,6 +232,7 @@ export interface NotifMessageBase<Method extends MethodName, Result> extends Jso
 
 const _N: unique symbol = Symbol();
 type Subscription<NotificationResult = any> = { [_N]: NotificationResult };
+type SubscriptionId = U.Branded<string, typeof _N, Subscription>;
```
tjjfvi, line 235: Should this have a `NotificationResult` type parameter?

rpc/messages.ts
```diff
@@ -231,6 +232,7 @@ export interface NotifMessageBase<Method extends MethodName, Result> extends Jso
 
 const _N: unique symbol = Symbol();
 type Subscription<NotificationResult = any> = { [_N]: NotificationResult };
```
tjjfvi, line 234:
```suggestion
type Subscription<NotificationResult = unknown> = { [_N]: NotificationResult };
```

PR #113, opened 2022-06-15
chore: RPC-related Cleanup

harrysolovay:


effect/std/atoms/rpcClient.ts
```diff
@@ -2,14 +2,17 @@ import * as rpc from "../../../rpc/mod.ts";
 import { effector, EffectorArgs } from "../../impl/mod.ts";
 
 export interface RpcClientR<Beacon> {
-  rpc: rpc.RpcClientFactory<Beacon>;
+  rpc: rpc.RpcClientFactory<Beacon, rpc.RpcError>;
```
harrysolovay, line 5: Ignore the fact that the RPC errors aren't piped in through the top-level `rpcClient` effect. We'll get to this later (#46).

rpc/Base.ts
```diff
@@ -8,41 +9,84 @@ export type ListenerCb<IngressMessage_ extends M.IngressMessage = M.IngressMessa
...
-    super();
-  }
-}
+export abstract class RpcClient<RpcError extends E.RpcError> {
+  #nextId = 0;
+  listeners = new Map<ListenerCb, boolean>();
+  #errors: RpcError[] = [];
 
-export abstract class RpcClient {
+  /** Send a message to the RPC server without
```
tjjfvi, line 17: without what?

rpc/Base.ts
```diff
@@ -8,41 +9,84 @@ export type ListenerCb<IngressMessage_ extends M.IngressMessage = M.IngressMessa
 
 export type StopListening = () => void;
 
-export class RpcError extends Error {
-  constructor(readonly inner: M.ErrMessage) {
-    super();
-  }
-}
+export abstract class RpcClient<RpcError extends E.RpcError> {
```
tjjfvi, line 12: Why is this generic on the error?

harrysolovay: My thinking is that we'll want beacon-specific error types (web sockets can error out differently from Smoldot client instances).

PR #114, opened 2022-06-15
feat: subscription iteration

### LINKED ISSUE #10: Wrapping RPC Client with Async Generator
What is the best approach for getting an async iterator over incoming messages with an RPC client? Ideally, when the iteration is ended, we can trigger a client-specific callback (to close the connection, for example).

```ts
for await (const message of rpc.iter(client)) {
  // utilize
  if (isLastMessage(message)) {
    break;
  }
}

// `client` is no longer needed, so the connection is closed
```

harrysolovay:
Resolves #10.

Desired end state (of raw RPC usage, not the effect equivalent):

```ts
import * as C from "capi";

const subscription = await C.client(C.POLKADOT_RPC_URL);

const subscription = client.subscription("chain_subscribeAllHeads", []);

for await (const notif of subscription) {
  console.log(notif);
  if (someCondition(notif)) {
    break; // Closes subscription
  }
}
```

Currently working:

```ts
import * as C from "capi";

const client = await C.wsRpcClient(C.POLKADOT_RPC_URL);

const subscription = new C.Subscription(client, "chain_subscribeAllHeads", []);

for await (const notif of subscription) {
  console.log(notif);
  if (someCondition(notif)) {
    break; // Closes subscription
  }
}
```

### DIFF
--- a/examples/rpc/subscription_iter.ts
+++ b/examples/rpc/subscription_iter.ts
@@ -1,0 +1,7 @@
+import * as C from "../../mod.ts";
+
+const client = await C.wsRpcClient(C.POLKADOT_RPC_URL);
+const subscription = new C.Subscription(client, "chain_subscribeAllHeads", []);
+for await (const notif of subscription) {
+  console.log(notif);
+}
--- a/rpc/Subscription.ts
+++ b/rpc/Subscription.ts
@@ -1,0 +1,57 @@
+import { deferred } from "../_deps/async.ts";
+import { RpcClient, StopListening } from "./Base.ts";
+import { RpcError } from "./Error.ts";
+import * as M from "./messages.ts";
+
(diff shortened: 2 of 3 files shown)


harrysolovay:
Taking a different approach. RPC client subscriptions and corresponding async iterables will NOT be coupled. Instead, devs will be able to use a `C.iter` util to get a pushable iterable.

```ts
import * as C from "../mod.ts";

// ...

const iter = C.iter();

await chain
  .pallet("Balances")
  .extrinsic("transfer")
  .call({
    dest: alice.asPublicKeyBytes(),
    value: 12345n,
  })
  .signed(bob, bob.sign)
  .send(iter);
```

The type `T` of an `Iter<T>` can be refined by `.send(iter)`, which will (safely) assert the type `T` based on the notification value.

PR #116, opened 2022-06-17
chore: crypto util / wasm binding cleanup

harrysolovay:
- Begin doing some chain-specific codegen for a soon-to-be-developed ss58 conversion util
- Clean up how we expose/load wasm bindings
- Clean up the sr25519 utils (most of which are solely for testing)
- Bind to ss58 encoding/decoding implementation
- Simplify usage of these bindings in `transfer` example
- Fix type errors elsewhere
- `ErrorCtor` util


karl-kallavus (approved):
🍰

PR #118, opened 2022-06-18
chore: fix rpc message typings & clean up some tsdocs

harrysolovay:


frame_metadata/Codec.ts
```diff
@@ -6,7 +6,7 @@ export type DeriveCodec = (typeI: number) => $.Codec<unknown>;
 
 /**
  * All derived codecs for ZSTs will use this exact codec,
- * so `derivedCodec === $null` is true iff the type is a ZST.
```
tjjfvi, line 9: This was not a typo – `iff` in logic is an abbreviation for "if and only if", which is the most accurate description of the contract of `$null`.

util/types.ts
```diff
@@ -14,7 +11,9 @@ export type EnsureLookup<
   },
 > = Lookup;
 
-export type Flatten<T> = T extends Function ? T : { [K in keyof T]: T[K] };
+export type Flatten<T> = T extends (infer E)[] ? Flatten<E>[]
+  : T extends object ? { [K in keyof T]: Flatten<T[K]> }
+  : T;
```
tjjfvi, line 16: This breaks `Flatten` for tuples. I've recently adopted
```ts
export type Flatten<T> = [{ [K in keyof T]: Flatten<T[K]> }][0]
```
with the caveat that it should only be used to wrap things known to be simple objects or arrays

frame_metadata/Codec.ts
harrysolovay: Ahh, apologies

tjjfvi:
(the above posthumous review is here for convenience, and will be extracted later)

PR #129, opened 2022-06-20
chore: decouple rpc method types from clients

harrysolovay:
This PR allows users to supply their own method lookups to beacons. This paves the way for narrow RPC typings for different chains (all in the same env!).

```ts
const myBeacon = rpc.beacon<{
  someFn(a: string, b: number): string;
}>();

const client = await rpc.client(myBeacon);
assert(!(client instanceof Error));
const result = await client.call("state_getMetadata", []);
//                               ~~~~~~~~~~~~~~~~~~~
//                               ^
//                               Argument of type '"state_getMetadata"' is not assignable to parameter of type '"someFn"



console.log(result);
await client.close();
```


tjjfvi (commented):
Very nice, was going to suggest this

PR #131, opened 2022-06-20
chore: beacon-related tweaks

harrysolovay:


effect/std/metadata.ts
```diff
@@ -7,7 +7,7 @@ import { rpcCall } from "./rpcCall.ts";
 
 export const metadata = effector(
   "metadata",
-  (rpc: EffectorItem<rpc.AnyClient>, blockHash?: EffectorItem<U.HashHexString>) => {
+  (rpc: EffectorItem<any>, blockHash?: EffectorItem<U.HashHexString>) => {
```
harrysolovay, line 10: Effects are in need of a refactor, given the way that RPC types flow through the beacon into any/all calls.

known/mod.ts
```diff
@@ -1,15 +1,3 @@
-import { Branded } from "../util/mod.ts";
```
harrysolovay, line 1: Proper branded types are still a todo.

rpc/Base.ts
```diff
@@ -18,10 +18,10 @@ export type Subscription<NotificationResult = any> = { [_N]: NotificationResult
 
 export interface ClientProps<
   M extends AnyMethods,
-  Beacon,
+  DiscoveryValue,
```
harrysolovay, line 21: Zeroing in on correct terminology. Beacons contain discovery values (such as URLs and chain specs). Beacons are encoded with the type information necessary to constrain client usage.

harrysolovay, line 21: Zeroing in on correct terminology. Beacons contain discovery values (such as URLs and chain specs). Beacons are encoded with the type information necessary to constrain client usage.

Issue #140, opened 2022-06-27
Dual node-template versions (mac, linux)


wirednkod:
In order to be able to run a local node, at the moment exists a `node-template` inside test-util.
During some pair programming with 2 different OS (mac, linux), it was identified that the node-template that was executing on Mac would not run on Linux.

It seems like a good idea, to have 2 different versions of `node-template`, inside test util, named e.g.:
a) `node-template-linux` and b)`node-template-mac`, and create a new task that will detect the OS and run the one needed.

harrysolovay:
Currently, the repository contains the compiled `node-template` bin. This is not ideal. Perhaps a better approach is to extend [the underlaying configuration](https://github.com/paritytech/capi/blob/afdb68694f73b2893f314e114a0999addbc38347/util/substrateProcess.ts#L6-L12), such that a repository URL can be specified.

```ts
const process = await substrateProcess({
  repo: "git@github.com:substrate-developer-hub/substrate-node-template.git",
  dev: true,
});
```

Under the hood, this would clone the repo into a `~/.local/share/capi/cloned` and compile and execute it. Perhaps we should allow the user to specify the clone dest as well.

```ts
const process = await substrateProcess({
  repo: "git@github.com:substrate-developer-hub/substrate-node-template.git",
  dest: "~/Desktop/tmp",
  dev: true,
});
```

harrysolovay:
I did some experimentation in #142. Seems we should support several means of local node instantiation:

- Specifying repo (user, repo name, branch, directory into which to clone)
- Specifying project (path on disk (to Substrate project root))
- Specifying a compiled executable

In the mocked branch, the following clones, compiles and runs a node.

```ts
const process = await substrateNode({
  repo: "git@github.com:substrate-developer-hub/substrate-node-template.git",
  cloneDir: "./examples/node-template",
  ensureSynced: false, // we don't care about pulling new commits with each run
  recompile: false, // we don't care about recompiling with each run
});
```

This works, but it feels like bloat. Doesn't allow for the seamless experience we want:

```ts
const node = await test.node();

// connect to & use node here

node.close();
```

## Some Tentative Requirements

- **Avoid bundling many (or even any) artifacts in Capi** (let's eventually remove `test-util/node-template` entirely –– for now, it gives us development velocity)
- **Avoid depending on Docker/Nix/virtualization solutions**... although it would be quite convenient to simply specify an image and offload the machine-specific considerations

wirednkod:
[...]

PR #142, opened 2022-06-27
test: mocking out local node utils

harrysolovay:
Just playing with some ideas from #140. I expect to close this draft.


harrysolovay:
For now going to [point to the `polkadot` exe](https://github.com/paritytech/capi/blob/8c68b84525be7f617d874a82e4d15e21ffd3e779/test-util/node.ts#L25-L29) by default.

PR #194, opened 2022-08-30
feat: add SS58 Codec JS implementation

### LINKED ISSUE #173: Ss58 Codec
Per [request](https://github.com/denoland/deno_std/issues/2447) in Deno std, it appears that Base58 transcoding will soon make its way into the std lib. At this point, let's introduce a pure JS implementation of Ss58 transcoding (which will depend on the Base58 impl in std).

kratico:
Resolves #173 

Adds SS58 Codec JS implementation using
- Deno base58 codec implementation from https://raw.githubusercontent.com/denoland/deno_std/main/encoding/base58.t
- Blake2b hash implementation from [@noble/hashes](https://github.com/paulmillr/noble-hashes)

TODO
- [x] replace SS58 wasm implementation
- [x] rename `./ss58-js` to `./ss58`
- [x] validate allowed public key lengths with https://github.com/substrate-developer-hub/substrate-docs/blob/archive/main-mdx/v3/docs/07-advanced/h-ss58/index.mdx
- [x] add more unit tests

Reference SS58 Codec implementations
- https://github.com/shamilsan/ss58.org/blob/master/src/encoder.rs
- https://github.com/paritytech/substrate/blob/0ba251c9388452c879bfcca425ada66f1f9bc802/primitives/core/src/crypto.rs#L264
- https://github.com/polkadot-js/common/blob/master/packages/util-crypto/src/address/encode.ts
- https://github.com/polkadot-js/ss58/blob/master/index.js


ss58-js/consts.ts
```diff
@@ -0,0 +1,16 @@
+import ss58Registry from "../deps/ss58_registry.ts";
+
+export const ALLOWED_PUBLIC_KEY_LENGTHS = [32, 33];
```
kratico, line 3: Polkadot-JS allows more values 
See https://github.com/polkadot-js/common/blob/master/packages/util-crypto/src/address/defaults.ts#L7

ss58-js/constants.ts
```diff
@@ -0,0 +1,16 @@
+import ss58Registry from "../deps/ss58_registry.ts";
+
+export const ALLOWED_PUBLIC_KEY_LENGTHS = [32, 33];
```
kratico, line 3: Polkadot-JS allows more values 
See https://github.com/polkadot-js/common/blob/master/packages/util-crypto/src/address/defaults.ts#L7

ss58-js/mod.ts
```diff
@@ -0,0 +1,89 @@
...
+
+import {
+  ALLOWED_ADDRESS_LENGTHS,
+  ALLOWED_NETWORK_PREFIXES,
+  ALLOWED_PUBLIC_KEY_LENGTHS,
+  CHECKSUM_LENGTH,
+  SS58PRE,
+} from "./constants.ts";
+
+export const encode = (prefix: number, pubKey: string): string => {
```
harrysolovay, line 14: Let's change the signature

```diff
- (prefix: number, pubKey: string): string;
+ (prefix: number, pubKey: Uint8Array): Uint8Array;
```

Line 15 can then be deleted.

ss58-js/mod.ts
```diff
@@ -0,0 +1,89 @@
...
+  const pubKeyBytes = hex.decode(pubKey);
+
+  if (!ALLOWED_PUBLIC_KEY_LENGTHS.includes(pubKeyBytes.length)) {
```
harrysolovay, line 17: Not sure if placing these in constants buys a ton. Maybe just do the comparison?

```ts
if (pubKeyBytes.length === 32 || pubKeyBytes.length === 33) { /* misc */ }
```

ss58-js/mod.ts
```diff
@@ -0,0 +1,89 @@
...
+    throw new Error("Invalid public key length");
+  }
+
+  if (!ALLOWED_NETWORK_PREFIXES.includes(prefix)) {
```
harrysolovay, line 21: Not loving this lookup. Should prefix validation be optional/pluggable? Perhaps there could be an optional 3rd param that accepts a list of (to-be-treated-as-valid) prefixes.

```diff
- (prefix: number, pubKey: Uint8Array): Uint8Array;
+ (prefix: number, pubKey: Uint8Array, validPrefixes?: number[]): Uint8Array;
```

Thoughts?

ss58-js/mod.ts
```diff
@@ -0,0 +1,89 @@
...
+  const addrBytes = base58.decode(address);
+
+  if (!ALLOWED_ADDRESS_LENGTHS.includes(addrBytes.length)) {
+    throw new Error("Invalid address length");
+  }
+
+  const prefixLength = (addrBytes[0] as number) & 0b0100_0000 ? 2 : 1;
+
+  const prefix: number = prefixLength === 1
+    ? addrBytes[0] as number
```
harrysolovay, line 62:
```suggestion
    ? addrBytes[0]!
```

ss58-js/mod.ts
```diff
@@ -0,0 +1,89 @@
...
+    : (((addrBytes[0] as number) & 0b0011_1111) << 2) | ((addrBytes[1] as number) >> 6)
```
harrysolovay, line 63:
```suggestion
    : (((addrBytes[0]!) & 0b0011_1111) << 2) | ((addrBytes[1]!) >> 6)
```

ss58-js/mod.ts
```diff
@@ -0,0 +1,89 @@
...
+      | (((addrBytes[1] as number) & 0b0011_1111) << 8);
```
harrysolovay, line 64:
```suggestion
      | (((addrBytes[1]!) & 0b0011_1111) << 8);
```

ss58-js/mod.ts
```diff
@@ -0,0 +1,89 @@
...
+
```
harrysolovay, line 65: Note: we generally want to avoid these non-null assertions. Are you certain that values exist at these indices of `addrBytes`?

ss58-js/mod.ts
```diff
@@ -0,0 +1,89 @@
...
+    dkLen: 512 / 8,
+  });
+
+  hasher.update(SS58PRE);
+  hasher.update(new Uint8Array(addrBytes.buffer, 0, addrBytes.length - CHECKSUM_LENGTH));
+
+  const digest = hasher.digest();
+  const checksum = new Uint8Array(addrBytes.buffer, addrBytes.length - CHECKSUM_LENGTH);
+
+  if (digest[0] !== checksum[0] || digest[1] != checksum[1]) {
```
harrysolovay, line 76:
```suggestion
  if (digest[0] !== checksum[0] || digest[1] !== checksum[1]) {
```

ss58-js/mod.ts
```diff
@@ -0,0 +1,89 @@
...
+  ALLOWED_PUBLIC_KEY_LENGTHS,
+  CHECKSUM_LENGTH,
+  SS58PRE,
+} from "./constants.ts";
+
+export const encode = (prefix: number, pubKey: string): string => {
+  const pubKeyBytes = hex.decode(pubKey);
+
+  if (!ALLOWED_PUBLIC_KEY_LENGTHS.includes(pubKeyBytes.length)) {
+    throw new Error("Invalid public key length");
```
harrysolovay, line 18: Let's subtype errors.

```ts
import * as U from "../util/mod.ts";

export class InvalidPublicKeyLenError extends ErrorCtor("InvalidPublicKeyLenError") {}
```

This way, their names exist at the static type level.

```ts
const e = new InvalidPublicKeyLenError();

// type is "InvalidPublicKeyLenError"
e.name;

// Also, we can identify the instance type at runtime
if (e instanceof InvalidPublicKeyLenError) {
  // ...
}
```

Make sure to export these error subtypes as well.

Also, we may want to attach additional data.

```diff
import * as U from "../util/mod.ts";

- export class InvalidPublicKeyLenError extends ErrorCtor("InvalidPublicKeyLenError") {}
+ export class InvalidPublicKeyLenError extends ErrorCtor("InvalidPublicKeyLenError") {
+   constructor(readonly actualLen: number) {
+     super();
+   }
+ }
```

ss58-js/mod.ts
```diff
@@ -0,0 +1,89 @@
...
+  addrBytes.set(pubKeyBytes, prefixBytes.length);
+  addrBytes.set(checksum, prefixBytes.length + pubKeyBytes.length);
+
+  return base58.encode(addrBytes);
+};
+
+export const decode = (address: string): [number, string] => {
+  const addrBytes = base58.decode(address);
+
+  if (!ALLOWED_ADDRESS_LENGTHS.includes(addrBytes.length)) {
```
harrysolovay, line 55: Same as above. If you must extract these constants, I'd recommend doing so in this same file. And perhaps use a record/index instead of an array/includes method.

Bad

```ts
[32, 33].includes(actual);
```

Good

```ts
({ 32: true, 33: true } as Record<number, true>)[actual];
```

ss58-js/mod.test.ts
```diff
@@ -0,0 +1,111 @@
+import { assertEquals, assertThrows } from "../deps/std/testing/asserts.ts";
+import { decode, encode } from "./mod.ts";
+
+const ALICE_PUBLIC_KEY = "d43593c715fdd31c61141abd04a99fd6822c8558854ccde39a5684e7a56da27d";
```
harrysolovay, line 4: Use the test util instead.

```ts
import * as t from "../test-util/mod.ts";

t.alice.publicKey;
```

ss58-js/mod.test.ts
```diff
@@ -0,0 +1,111 @@
...
+
+[
+  {
+    name: "ss58.encode should encode 8-bit prefix polkadot address",
```
harrysolovay, line 8: Simplify test data?

```ts
(<[networkName: string, prefix: number][]>[
  ["Polkadot", 0],
  ["Aventus", 65],
  ...misc,
]).forEach(([networkName, ss58Format]) => {
  // ...
});
```

ss58-js/mod.test.ts
```diff
@@ -0,0 +1,111 @@
...
+  {
+    name: "ss58.encode should encode 16-bit prefix aventus address",
+    args: [65, ALICE_PUBLIC_KEY],
+    expected: "cLxkfNUiCYsb57YLhTJdNVKxUTB1VTpeygYZNhYuFc83KrFy7", // cspell:disable-line
+  },
+].forEach(({ name, args, expected }) => {
+  Deno.test({
+    name,
+    fn: () => {
+      // @ts-ignore
```
harrysolovay, line 26: Let's avoid ts-ignoring in situations where we don't have to.

Instead of having an `args` props, let's have `prefix` and `publicKey` props, which we can supply directly to the `encode` fn.

ss58-js/mod.test.ts
```diff
@@ -0,0 +1,111 @@
...
+  {
+    name: "ss58.encode should throw for an invalid network prefix",
+    args: [0b0000_1111_1111_1111, ALICE_PUBLIC_KEY],
+    errorMessage: "Invalid network prefix",
+  },
+].forEach(({ name, args, errorMessage }) => {
+  Deno.test({
+    name,
+    fn: () => {
+      // @ts-ignore
```
harrysolovay, line 53: Same thing here. Side note: I love the error testing!

ss58-js/mod.test.ts
```diff
@@ -0,0 +1,111 @@
...
+  {
+    name: "ss58.decode should decode 16-bit prefix aventus address",
+    args: ["cLxkfNUiCYsb57YLhTJdNVKxUTB1VTpeygYZNhYuFc83KrFy7"], // cspell:disable-line
+    expected: [65, ALICE_PUBLIC_KEY],
+  },
+].forEach(({ name, args, expected }) => {
+  Deno.test({
+    name,
+    fn: () => {
+      // @ts-ignore
```
harrysolovay, line 79: Any again

ss58-js/mod.test.ts
```diff
@@ -0,0 +1,111 @@
...
+    name: "ss58.decode should throw for an invalid checksum",
+    // Alice public key with a 0xff 0xff checksum
+    args: ["15oF4uVJwmo4TdGW7VfQxNLavjCXviqxT9S1MgbjMNHr6ccE"],
+    errorMessage: "Invalid address checksum",
+  },
+].forEach(({ name, args, errorMessage }) => {
+  Deno.test({
+    name,
+    fn: () => {
+      // @ts-ignore
```
harrysolovay, line 107: And again

ss58-js/constants.ts
harrysolovay: Interesting. Just opened #196.

harrysolovay (changes requested):
Nice first pass!

ss58-js/mod.ts
```diff
@@ -0,0 +1,89 @@
...
+    ? addrBytes[0] as number
+    : (((addrBytes[0] as number) & 0b0011_1111) << 2) | ((addrBytes[1] as number) >> 6)
+      | (((addrBytes[1] as number) & 0b0011_1111) << 8);
+
+  const hasher = blake2b.create({
+    dkLen: 512 / 8,
+  });
+
+  hasher.update(SS58PRE);
+  hasher.update(new Uint8Array(addrBytes.buffer, 0, addrBytes.length - CHECKSUM_LENGTH));
```
tjjfvi, line 71:
```suggestion
  hasher.update(addrBytes.subarray(0, addrBytes.length - CHECKSUM_LENGTH));
```

ss58-js/mod.ts
```diff
@@ -0,0 +1,89 @@
...
+
+  const digest = hasher.digest();
+  const checksum = new Uint8Array(addrBytes.buffer, addrBytes.length - CHECKSUM_LENGTH);
```
tjjfvi, line 74:
```suggestion
  const checksum = addrBytes.subarray(addrBytes.length - CHECKSUM_LENGTH);
```

ss58-js/mod.ts
```diff
@@ -0,0 +1,89 @@
...
+
+  if (digest[0] !== checksum[0] || digest[1] != checksum[1]) {
+    throw new Error("Invalid address checksum");
+  }
+
+  const pubKeyBytes = new Uint8Array(
+    addrBytes.buffer,
+    prefixLength,
+    addrBytes.length - prefixLength - CHECKSUM_LENGTH,
+  );
```
tjjfvi, line 84:
```suggestion
  const pubKeyBytes = addrBytes.subarray(
    prefixLength,
    addrBytes.length - CHECKSUM_LENGTH,
  );
```

ss58-js/mod.ts
```diff
@@ -0,0 +1,89 @@
...
+  const hasher = blake2b.create({
+    dkLen: 512 / 8,
+  });
+
+  hasher.update(SS58PRE);
+  hasher.update(prefixBytes);
+  hasher.update(pubKeyBytes);
+
+  const digest = hasher.digest();
+  const checksum = new Uint8Array(digest.buffer, 0, CHECKSUM_LENGTH);
```
tjjfvi, line 41:
```suggestion
  const checksum = digest.subarray(0, CHECKSUM_LENGTH);
```

tjjfvi (commented):
```ts
arr.subarray(start, end)
```
should be used instead of
```ts
new Uint8Array(arr.buffer, start, length)
```
as the latter produces incorrect results when `arr` is already a slice of the buffer.

(The correct version with `new Uint8Array` would be
```ts
new Uint8Array(arr.buffer, start + arr.byteOffset, length)
```
but it's better to use `.subarray` to reduce the likelihood of mistakes)

ss58-js/mod.ts
tjjfvi: Agreed – the raw logic should be decoupled from the string-encoding, so that raw `Uint8Array`s can be passed around internally

ss58-js/mod.ts
```diff
@@ -0,0 +1,89 @@
...
+  const addrBytes = new Uint8Array(prefixBytes.length + pubKeyBytes.length + CHECKSUM_LENGTH);
+
+  addrBytes.set(prefixBytes, 0);
+  addrBytes.set(pubKeyBytes, prefixBytes.length);
+  addrBytes.set(checksum, prefixBytes.length + pubKeyBytes.length);
+
+  return base58.encode(addrBytes);
+};
+
+export const decode = (address: string): [number, string] => {
```
tjjfvi, line 52: Like with `encode`, the raw logic should be decoupled from the string-encoding, so that raw Uint8Arrays can be passed around internally. Additionally, the tuple type should have labels for documentation purposes.
```suggestion
export const decode = (address: Uint8Array): [prefix: number, pubKey: Uint8Array] => {
```

ss58-js/mod.ts
kratico: Yeap, optional prefix validation sounds good

ss58-js/mod.ts
kratico: Perfect!
That will also do the check in O(1) time complexity

ss58-js/mod.ts
kratico: @harrysolovay @tjjfvi  if there is not need to encode it into a string, then, the base68 dependency is not needed for this, right?

kratico: I'm a bit confused on the suggested signatures for encode/decode
```
export const encode = (prefix: number, pubKey: Uint8Array): Uint8Array
export const decode = (address: Uint8Array): [prefix: number, pubKey: Uint8Array]
```

if `encode` returns bytes or `decode` receive bytes, base58 encoding/decoding is not used.

Note: I'm assuming that the SS58 address format needs to be base58 encoded as described [here](https://github.com/substrate-developer-hub/substrate-docs/blob/archive/main-mdx/v3/docs/07-advanced/h-ss58/index.mdx#basic-format) 
```
base58encode ( concat ( <address-type>, <address>, <checksum> ) )
```

ss58-js/mod.ts
kratico: yes because there is check for the address length

ss58-js/mod.test.ts
```diff
@@ -0,0 +1,132 @@
+import * as base58 from "../deps/std/encoding/base58.ts";
+import { assertEquals, assertThrows } from "../deps/std/testing/asserts.ts";
+import * as p from "../test-util/pairs.ts";
```
kratico, line 3: Doing
```
import * as t from "../test-util/mod.ts";
```

Output the following errors
```
error: TS2322 [ERROR]: Type 'Error' is not assignable to type 'SmoldotClient<Config_> | FailedToStartSmoldotError | FailedToAddChainError'.
  Property 'inner' is missing in type 'Error' but required in type 'FailedToAddChainError'.
    return smoldotInstance;
    ~~~~~~~~~~~~~~~~~~~~~~~
    at file:///Users/parity/ghq/github.com/paritytech/capi/rpc/providers/smoldot.ts:18:5

    'inner' is declared here.
      constructor(readonly inner: unknown) {
                  ~~~~~~~~~~~~~~~~~~~~~~~
        at file:///Users/parity/ghq/github.com/paritytech/capi/rpc/providers/smoldot.ts:90:15

TS7006 [ERROR]: Parameter 'response' implicitly has an 'any' type.
      jsonRpcCallback: (response) => {
                        ~~~~~~~~
    at file:///Users/parity/ghq/github.com/paritytech/capi/rpc/providers/smoldot.ts:25:25

TS2503 [ERROR]: Cannot find namespace 'smoldot'.
  #chain?: smoldot.Chain;
           ~~~~~~~
    at file:///Users/parity/ghq/github.com/paritytech/capi/rpc/providers/smoldot.ts:38:12

TS2503 [ERROR]: Cannot find namespace 'smoldot'.
const _state: { smoldotInstance?: smoldot.Client } = {};
                                  ~~~~~~~
    at file:///Users/parity/ghq/github.com/paritytech/capi/rpc/providers/smoldot.ts:74:35

TS2503 [ERROR]: Cannot find namespace 'smoldot'.
async function ensureInstance(): Promise<smoldot.Client | FailedToStartSmoldotError> {
                                         ~~~~~~~
    at file:///Users/parity/ghq/github.com/paritytech/capi/rpc/providers/smoldot.ts:76:42

Found 5 errors.
```

I'll investigate

cc @harrysolovay

tjjfvi: I can't reproduce those errors locally or [on CI](https://github.com/paritytech/capi/commit/ad78b161b9ab9e07440a1e7cafa9ba606bbedc07), so I think this is a local issue.

Can you try running with `--reload` and/or nuking `DENO_DIR` (from `deno info`)?

ss58-js/mod.ts
```diff
@@ -0,0 +1,104 @@
...
+
+export class InvalidPublicKeyLengthError extends U.ErrorCtor("InvalidPublicKeyLengthError") {}
+export class InvalidNetworkPrefixError extends U.ErrorCtor("InvalidNetworkPrefixError") {}
+export class InvalidAddressLengthError extends U.ErrorCtor("InvalidAddressError") {}
+export class InvalidAddressChecksumError extends U.ErrorCtor("InvalidAddressChecksumError") {}
+
+export const encode = (
+  prefix: number,
+  pubKey: Uint8Array,
+  validNetworkPrefixes?: number[],
```
tjjfvi, line 29:
```suggestion
  validNetworkPrefixes?: readonly number[],
```

tjjfvi (commented):
[Suggested changes to `ss58-js/mod.test.ts`](https://github.com/paritytech/capi/commit/133e1979c4c61fb92b75ccb731dcfb50e9f12236): I merged the successful `encode` and `decode` tests, as they use the same data, and I unrolled the other loops – though I appreciate the DRYness of the loops, the unrolled version is much shorter and easier to read.

kratico:
> [Suggested changes to `ss58-js/mod.test.ts`](https://github.com/paritytech/capi/commit/133e1979c4c61fb92b75ccb731dcfb50e9f12236): I merged the successful `encode` and `decode` tests, as they use the same data, and I unrolled the other loops – though I appreciate the DRYness of the loops, the unrolled version is much shorter and easier to read.

Thanks @tjjfvi , I like your suggestion and as you said it's easier to read

ss58-js/mod.ts
tjjfvi: I think we should rename `encode`/`decode` to `encodeRaw`/`decodeRaw` and add
```ts
export const encode = (prefix: number, pubKey: Uint8Array) => base58.encode(encodeRaw(prefix, pubkey));
export const decode = (address: string) => decodeRaw(base58.decode(address));
```

Most of the time, one will use `encode`/`decode`, but it is worthwhile to have `encodeRaw`/`decodeRaw` as options in the rare cases where one deals with the raw data.

kratico: makes sense 😄

ss58/mod.test.ts
```diff
@@ -1,15 +1,64 @@
...
+      [0, p.alice.publicKey],
+    ],
+    [
+      "substrate",
+      "5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY",
+      [42, p.alice.publicKey],
+    ],
+    [
+      "aventus",
+      "cLxkfNUiCYsb57YLhTJdNVKxUTB1VTpeygYZNhYuFc83KrFy7", // cspell:disable-line
```
harrysolovay, line 28: Let's avoid placing cspell-disabling comments in the source –– hopefully you can do this in your workspace settings?

ss58/mod.ts
```diff
@@ -1,17 +1,113 @@
...
-  return {
-    encode: instance.encode,
-    decode: instance.decode as Ss58["decode"],
-  };
-}
+import { blake2b } from "../deps/hashes.ts";
+import * as base58 from "../deps/std/encoding/base58.ts";
+import * as U from "../util/mod.ts";
+
+// SS58PRE = Uint8Array.from("SS58PRE".split("").map((c) => c.charCodeAt(0)));
```
harrysolovay, line 5: Is there a reason to keep this commented alternative around?

ss58/mod.ts
```diff
@@ -1,17 +1,113 @@
...
+export class InvalidPublicKeyLengthError extends U.ErrorCtor("InvalidPublicKeyLengthError") {}
+export class InvalidNetworkPrefixError extends U.ErrorCtor("InvalidNetworkPrefixError") {}
+export class InvalidAddressLengthError extends U.ErrorCtor("InvalidAddressError") {}
+export class InvalidAddressChecksumError extends U.ErrorCtor("InvalidAddressChecksumError") {}
+
+export const encode = (
+  prefix: number,
+  pubKey: Uint8Array,
+  validNetworkPrefixes?: readonly number[],
+) => base58.encode(encodeRaw(prefix, pubKey, validNetworkPrefixes));
```
harrysolovay, line 31: NICE

ss58/mod.test.ts
kratico: I'll add it to the cspell config

ss58/mod.ts
kratico: I left it to state that the Uint8Array from the next line is the `SS58PRE` prefix.
I'll remove the code and write a better comment

ss58/mod.ts
kratico: credits to @tjjfvi

tjjfvi (dismissed):
Looks good, just one nit: can you rename `deps/hashes.ts` to `deps/blake2b.ts`?

tjjfvi (approved):

ss58-js/mod.test.ts
kratico: My issue was that I was running the tests with
```
deno test ss58/
```
instead of 
```
deno task test ss58/
```
which runs
```
deno test -A --no-check=remote -L=info --ignore=target "ss58/"
```

harrysolovay:
One last thing: @kratico please remove lines 11 and 12 from `wasm.rs`. Upon making this change, this PR LGTM.

harrysolovay (approved):

deps/std/encoding/base58.ts
```diff
@@ -0,0 +1,3 @@
+// FIXME: update dependency when deno release base58 encoding
+// export * from "https://deno.land/std@0.xyz.0/encoding/base58.ts";
+export * from "https://raw.githubusercontent.com/denoland/deno_std/main/encoding/base58.ts";
```
harrysolovay, line 3:
```suggestion
export * from "https://deno.land/std@0.154.0/encoding/base58.ts";
```

harrysolovay: (and let's get rid of lines 1-2)

harrysolovay (dismissed):
Just in time!: there was just a new release of Deno std

tjjfvi (approved):

PR #197, opened 2022-08-31
chore: add `run:browser` task

tjjfvi:
Allows testing the various examples in the browser via e.g.
```sh
deno task run:browser examples/transfer.ts
```
The task file is a bit hacked together, but as it's an internal util, I think this is fine.

### DIFF
--- a/_tasks/.gitignore
+++ b/_tasks/.gitignore
@@ -1,0 +1,1 @@
+_transpiled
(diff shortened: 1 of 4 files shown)


tjjfvi:
To clarify:

Users who want to use capi from the browser will *not* use anything like `run_browser` – they will likely use the DNT-produced package along with a bundler of their choice.

The `run_browser` script is solely for *internal* testing, so that we can run the Deno example scripts in the browser without having to go through DNT and a bundler.

harrysolovay (approved):

PR #199, opened 2022-08-31
fix: storage key handling

### LINKED ISSUE #193: Dynamic key type and generation.
Current demos seem to indicate that the key generation has some assumptions about the number of components a key has, and how it is hashed. In general, I suggest decoupling storage read and key generation for future proofing. 

```
let key = C.key(config, "pallet", "storage", [other, stuff]);
let value  = C.read(key);
```

or something like that.

The important points to remember is: 

1. storage keys can typically be anything. We have some `well_known_keys` that are not hashed at all, and most others are hashed.
2. FRAME storage items have be composed of multiple hashes, and multiple hashers. We cannot make an assumption about the key being 2 or 3 parts only. For example, consider the new `StorageNMap`.

tjjfvi:
Fixes #193.

This changes `$storageKey` to deal with keys in a tuple form (`[]`/`[A]`/`[A,B,...]`), rather than a collapsed form (`null`/`A`/`[A,B,...]`). The collapsed form was never being used in user code, so this just resulted in unnecessary extra work.

It also fixes the bug behavior of `$storageKey` for multi-part keys described in <https://github.com/paritytech/capi/issues/193#issuecomment-1231675617>.

### DIFF
--- a/effect/atoms/StorageKey.ts
+++ b/effect/atoms/StorageKey.ts
@@ -4,7 +4,7 @@ import { atomFactory } from "../sys/Atom.ts";
 
 export const storageKey = atomFactory(
   "StorageKey",
-  ($storageKey: $.Codec<unknown>, ...keys: unknown[]) => {
-    return U.hex.encode($storageKey.encode(keys.length === 1 ? keys[0] : keys)) as U.HexString;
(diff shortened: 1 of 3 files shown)


harrysolovay (approved):
YES

PR #203, opened 2022-09-02
feat: js hashers

tjjfvi:
This removes the last of the wasm, replacing the Rust hashers with JS impls.

The `blake2` implementation is imported from `@noble/hashes`, which we already depended on.

The `twox` implementation is adapted from `@polkadot-js/util-crypto`'s implementation.


tjjfvi:
Obviously a bit worse perf-wise, but I still think this is worthwhile to simplify the repo.

We can look into optimizing the hashers later if this is relevant to perf.

harrysolovay (approved):

PR #208, opened 2022-09-05
chore: update devcontainer docker image

kratico:
Related #41 

Updates devcontainer docker image to install the following tools as `vscode` user
- `deno`
- `dprint`
- `node` (needed for `cspell`)
- `cspell`

**Developer test plan**

Using VSCode, start the project in a container
- open a `.ts` file, and try go to Class/Interface definitions in other files using `F12` (or `command + click`)
- edit a `.ts` file, add many line breaks, and validate that `dprint` with the VSCode extension formats the text
- open the VSCode terminal and run `cspell "**/*"`

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -9,7 +9,28 @@
       "**/.git/subtree-cache/**": true,
       "target/**": true
     },
-    "dprint.path": "/home/vscode/.dprint/bin/dprint"
+    "[jsonc]": {
+      "editor.defaultFormatter": "esbenp.prettier-vscode"
+    },
+    "[rust]": {
+      "editor.defaultFormatter": "rust-lang.rust-analyzer"
+    },
+    "[yaml]": {
+      "editor.defaultFormatter": "esbenp.prettier-vscode"
+    },
+    "[typescript]": {
+      "editor.defaultFormatter": "dprint.dprint"
+    },
+    "deno.codeLens.testArgs": ["--no-check=remote", "-A", "-L=info"],
+    "deno.config": "./deno.jsonc",
+    "deno.enable": true,
+    "deno.lint": true,
+    "editor.defaultFormatter": "dprint.dprint",
+    "editor.formatOnSave": true,
+    "editor.tabSize": 2,
+    "prettier.printWidth": 100,
+    "rust-analyzer.diagnostics.disabled": ["incorrect-ident-case"],
+    "rust-analyzer.cargo.features": ["all"]
   },
   "extensions": [
     "EditorConfig.EditorConfig",
@@ -23,6 +44,6 @@
     "streetsidesoftware.code-spell-checker",
     "vadimcn.vscode-lldb"
   ],
-  "postCreateCommand": "deno task bootstrap",
+  "postCreateCommand": "deno task star",
   "remoteUser": "vscode"
 }
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,13 +2,20 @@ FROM mcr.microsoft.com/vscode/devcontainers/rust:0-1
 
 RUN rustup target add wasm32-unknown-unknown
 
-ENV DENO_INSTALL=/usr/local
 # crates.io index update fails without this (https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=918854)
 ENV CARGO_NET_GIT_FETCH_WITH_CLI=true
 
-RUN curl -fsSL https://deno.land/x/install/install.sh | sh
-RUN curl -fsSL https://dprint.dev/install.sh | sh
+USER vscode
 
-RUN export PATH=/usr/local/cargo/bin:$PATH
-RUN export DPRINT_INSTALL="/home/vscode/.dprint"
-RUN export PATH="$DPRINT_INSTALL/bin:$PATH"
+RUN /bin/bash -c 'curl -fsSL https://deno.land/x/install/install.sh | sh \
+  && curl -fsSL https://dprint.dev/install.sh | sh \
+  && curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.39.1/install.sh | bash \
+  && export NVM_DIR=$HOME/.nvm \
+  && . $NVM_DIR/nvm.sh \
+  && nvm install --lts \
+  && npm -g install cspell@latest'
+
+ENV DENO_INSTALL=/home/vscode/.deno
+ENV PATH=$DENO_INSTALL/bin:$PATH
+ENV DPRINT_INSTALL=/home/vscode/.dprint
+ENV PATH=$DPRINT_INSTALL/bin:$PATH
--- a/words.txt
+++ b/words.txt
@@ -168,3 +168,4 @@ Lxkf
 Tpeyg
 getrandom
 Curto
+devcontainers


Dockerfile
```diff
@@ -2,13 +2,20 @@ FROM mcr.microsoft.com/vscode/devcontainers/rust:0-1
 
 RUN rustup target add wasm32-unknown-unknown
```
kratico, line 3: I've just noticed that rust is not used to build wasm dependencies.
Perhaps we could use a more basic image.

Dockerfile
```diff
@@ -2,13 +2,20 @@ FROM mcr.microsoft.com/vscode/devcontainers/rust:0-1
...
-RUN curl -fsSL https://deno.land/x/install/install.sh | sh
-RUN curl -fsSL https://dprint.dev/install.sh | sh
+USER vscode
 
-RUN export PATH=/usr/local/cargo/bin:$PATH
-RUN export DPRINT_INSTALL="/home/vscode/.dprint"
-RUN export PATH="$DPRINT_INSTALL/bin:$PATH"
+RUN /bin/bash -c 'curl -fsSL https://deno.land/x/install/install.sh | sh \
+  && curl -fsSL https://dprint.dev/install.sh | sh \
+  && curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.39.1/install.sh | bash \
```
kratico, line 12: this is only used for `cspell`.

An alternative installation method, could be to use node from NodeSource

```
curl -fsSL https://deb.nodesource.com/setup_18.x | sudo -E bash -
sudo apt-get install -y nodejs
```

Thoughts?

.devcontainer/devcontainer.json
```diff
@@ -9,7 +9,28 @@
       "**/.git/subtree-cache/**": true,
       "target/**": true
     },
-    "dprint.path": "/home/vscode/.dprint/bin/dprint"
+    "[jsonc]": {
```
kratico, line 12: This file seems to be a combination of `.vscode/settings.json` and `.vscode/extensions.json`.

It might be a good idea to add a deno task to generate it.

thoughts?

Dockerfile
harrysolovay: The wasm/Rust-related setup can be removed (thanks to #194 and #203). [The basic deno image](https://github.com/denoland/deno_docker) is a better base.

Dockerfile
harrysolovay: I suppose we still need to set up Node in order for Prettier and other devcontained extensions to work properly. [I stackoverflowed this](https://stackoverflow.com/questions/67959838/most-efficient-docker-image-with-both-node-deno-installed) over a year ago. The answer is subpar. @tjjfvi any recommendations?

.devcontainer/devcontainer.json
harrysolovay: Sounds like another thing we'd have to maintain. The configuration duplication is a bit frustrating... but let's hard-code it for now. Seems we can update the extension lists / get rid of Rust-related extensions such as `serayuzgur.crates`.

Dockerfile
tjjfvi: Not familiar with docker, but the setup seems fine to me 🤷

kratico: I decided to use `https://deb.nodesource.com/setup_18.x` so we don't need to create a non-root user to run `nvm`.

Note: this would be an alternative to setup nvm as a root or non-root user https://github.com/microsoft/vscode-dev-containers/blob/v0.195.0/containers/javascript-node/.devcontainer/base.Dockerfile#L10

.gitpod.Dockerfile
```diff
@@ -0,0 +1,18 @@
...
+  && curl -fsSL https://deb.nodesource.com/setup_18.x | bash - \
+  && apt-get install -y nodejs \
+  && npm -g install cspell@latest \
+  && apt-get autoremove -y \
+  && apt-get clean -y \
+  && rm -rf /var/lib/apt/lists/*
+
+# Gitpod creates a gitpod user to run without privileges
+# following https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#user
+ENV DENO_DIR=/home/gitpod/.deno
```
kratico, line 18: I duplicated the `Dockerfile` as workaround for Gitpod.
Gitpod runs deno (or any script) as non-root user `gitpod`.
Setting `DENO_DIR` allows the `gitpod` user to store deno cache files in a folder where `gitpod` has permissions.

.devcontainer/devcontainer.json
```diff
@@ -9,20 +9,33 @@
...
+    "deno.enable": true,
+    "deno.lint": true,
+    "editor.defaultFormatter": "dprint.dprint",
+    "editor.formatOnSave": true,
+    "editor.tabSize": 2,
+    "prettier.printWidth": 100
   },
   "extensions": [
     "EditorConfig.EditorConfig",
     "bungcip.better-toml",
```
harrysolovay, line 32: We can get rid of this as well, as I believe we no longer have any toml in this repo?

.gitpod.yml
```diff
@@ -1,14 +1,14 @@
 image:
-  file: Dockerfile
+  file: .gitpod.Dockerfile
+tasks:
+  - init: deno task star
 vscode:
   extensions:
     - EditorConfig.EditorConfig
     - bungcip.better-toml
```
harrysolovay, line 8: Same here

harrysolovay:
Running in a vscode dev container, I get the following:

Screen Shot 2022-09-07 at 3 02 04 PM

Meanwhile, not sure why this is showing (should be auto-installed, if I'm not mistaken):

Screen Shot 2022-09-07 at 3 02 15 PM

kratico:
> Running in a vscode dev container, I get the following:
> 
> Screen Shot 2022-09-07 at 3 02 04 PM
> 
> Meanwhile, not sure why this is showing (should be auto-installed, if I'm not mistaken):
> 
> Screen Shot 2022-09-07 at 3 02 15 PM

@harrysolovay Did you rebuild the container?

I'm not able to reproduce
After rebuilding the image `dprint` is globally installed
```
root@04b98afd969d:/workspaces/capi# which dprint
/usr/local/bin/dprint
```

.devcontainer/devcontainer.json
kratico: There is a `book.toml` file.
Shall I remove it anyway?

Dockerfile
```diff
@@ -1,14 +1,21 @@
...
+  && apt-get clean -y \
+  && rm -rf /var/lib/apt/lists/*
 
-RUN curl -fsSL https://deno.land/x/install/install.sh | sh
-RUN curl -fsSL https://dprint.dev/install.sh | sh
 
-RUN export PATH=/usr/local/cargo/bin:$PATH
-RUN export DPRINT_INSTALL="/home/vscode/.dprint"
-RUN export PATH="$DPRINT_INSTALL/bin:$PATH"
+FROM vscode as gitpod
```
kratico, line 17: Gitpod does not support multistage builds but VSCode devcontainer extension supports multi stage builds.

With this setup
- gitpod, uses the last stage to configure the container image
- devcontainer, uses the `vscode` stage configured in `devcontainer.json`

harrysolovay:
Do we want to remove [this line as well](https://github.com/paritytech/capi/blob/2c33c893110aca28a4d0b196b8f8e24a2cdaaef1/Readme.md?plain=1#L184) and add a badge such as [this](https://github.com/Effect-TS/core/blob/fb895d0c78db942d65b9b98af2298ad2d2dd89b6/README.md?plain=1#L1)?

kratico:
@harrysolovay this is badge ready now 🎉 

[Gitpod Open](https://gitpod.io/#https://github.com/paritytech/capi/tree/chore/dev-container)

[Open in Remote - Containers](https://vscode.dev/redirect?url=vscode://ms-vscode-remote.remote-containers/cloneInVolume?url=https://github.com/paritytech/capi)

harrysolovay (approved):

PR #211, opened 2022-09-07
feat: add extrinsic integration test

kratico:
**Description**

Add a simple extrinsic integration test that depends on the `polkadot` binary.

The `polkadot` binary is copied from the polkadot docker image.

**How Has This Been Tested?**

- validate that the test workflow action has run the `effect/extrinsic.test.ts` test
- check the test workflow logs to validate that the `polkadot` node was started during the test

image

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -9,6 +9,11 @@ jobs:
       - uses: denoland/setup-deno@004814556e37c54a2f6e31384c9e18e983317366 # v1.1.0
         with:
           deno-version: v1.x
+      - name: Download polkadot
+        run: |
+          docker run -d --entrypoint /bin/sh --rm -it --name tmp parity/polkadot:latest
+          docker cp tmp:/usr/bin/polkadot /usr/local/bin
+          docker stop tmp
       - run: deno task lint
       - run: deno task star
       - run: deno task test
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -11,9 +11,8 @@ jobs:
           deno-version: v1.x
       - name: Download polkadot
         run: |
-          docker run -d --entrypoint /bin/sh --rm -it --name tmp parity/polkadot:latest
-          docker cp tmp:/usr/bin/polkadot /usr/local/bin
-          docker stop tmp
+          curl -L -o /usr/local/bin/polkadot https://github.com/paritytech/polkadot/releases/download/v0.9.28/polkadot
+          chmod +x /usr/local/bin/polkadot
       - run: deno task lint
       - run: deno task star
       - run: deno task test
--- a/effect/extrinsic.test.ts
+++ b/effect/extrinsic.test.ts
@@ -1,0 +1,72 @@
+import { assertEquals, assertObjectMatch } from "../deps/std/testing/asserts.ts";
+import * as C from "../mod.ts";
+import * as t from "../test-util/mod.ts";
+import * as U from "../util/mod.ts";
+
+Deno.test({
+  name: "transfer",
+  fn: async (ctx) => {
+    const config = await t.config({ altRuntime: "westend" });
+
+    await ctx.step("extrinsic events", async () => {
+      const extrinsicEvents: string[] = [];
+
+      const root = C.sendAndWatchExtrinsic({
+        config,
+        sender: {
+          type: "Id",
+          value: t.alice.publicKey,
+        },
+        palletName: "Balances",
+        methodName: "transfer",
+        args: {
+          value: 12345n,
+          dest: {
+            type: "Id",
+            value: t.bob.publicKey,
+          },
+        },
+        sign(message) {
+          return {
+            type: "Sr25519",
+            value: t.alice.sign(message),
+          };
+        },
+        createWatchHandler(stop) {
+          return (event) => {
+            if (typeof event.params.result === "string") {
+              extrinsicEvents.push(event.params.result);
+            } else {
+              if (event.params.result.inBlock) {
+                extrinsicEvents.push("inBlock");
+              } else if (event.params.result.finalized) {
+                extrinsicEvents.push("finalized");
+                stop();
+              } else {
+                stop();
+              }
+            }
+          };
+        },
+      });
+
+      U.throwIfError(await root.run());
+
+      assertEquals(extrinsicEvents, ["ready", "inBlock", "finalized"]);
+    });
+
+    await ctx.step({
+      name: "account balance",
+      ignore: true,
+      fn: async () => {
+        const root = C.readEntry(config, "System", "Account", [t.bob.publicKey]);
+
+        const state = U.throwIfError(await root.run());
+
+        assertObjectMatch(state, { value: { data: { free: 12345n } } });
+      },
+    });
+
+    config.close();
+  },
+});
--- a/effect/extrinsic.test.ts
+++ b/effect/extrinsic.test.ts
@@ -1,56 +1,31 @@
+import type { KeyringPair } from "https://deno.land/x/polkadot@0.0.8/keyring/types.ts";
 import { assertEquals, assertObjectMatch } from "../deps/std/testing/asserts.ts";
 import * as C from "../mod.ts";
 import * as t from "../test-util/mod.ts";
 import * as U from "../util/mod.ts";
+import { SendAndWatchExtrinsicProps } from "./mod.ts";
 
 Deno.test({
-  name: "transfer",
+  name: "Balances.transfer",
   fn: async (ctx) => {
     const config = await t.config({ altRuntime: "westend" });
 
     await ctx.step("extrinsic events", async () => {
-      const extrinsicEvents: string[] = [];
-
-      const root = C.sendAndWatchExtrinsic({
-        config,
-        sender: {
-          type: "Id",
-          value: t.alice.publicKey,
-        },
-        palletName: "Balances",
-        methodName: "transfer",
-        args: {
-          value: 12345n,
-          dest: {
-            type: "Id",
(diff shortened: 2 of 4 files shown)


effect/extrinsic.test.ts
```diff
@@ -0,0 +1,72 @@
...
+      });
+
+      U.throwIfError(await root.run());
+
+      assertEquals(extrinsicEvents, ["ready", "inBlock", "finalized"]);
+    });
+
+    await ctx.step({
+      name: "account balance",
+      ignore: true,
```
kratico, line 60: I've disabled this step because I'm getting the following error
```
  account balance ... FAILED (5ms)
    error: InvalidStateError: readyState not OPEN
              ws.send(JSON.stringify(egressMessage));
                 ^
        at WebSocket.send (deno:ext/websocket/01_websocket.js:288:15)
```

.github/workflows/test.yml
```diff
@@ -9,6 +9,11 @@ jobs:
       - uses: denoland/setup-deno@004814556e37c54a2f6e31384c9e18e983317366 # v1.1.0
         with:
           deno-version: v1.x
+      - name: Download polkadot
+        run: |
+          docker run -d --entrypoint /bin/sh --rm -it --name tmp parity/polkadot:latest
+          docker cp tmp:/usr/bin/polkadot /usr/local/bin
```
kratico, line 15: The `polkadot` [docker image](https://github.com/paritytech/polkadot/blob/master/scripts/ci/dockerfiles/polkadot/polkadot_builder.Dockerfile#L10) is ubuntu based so it seems to be compatible with the GH workflow runner.

Perhaps there is a better alternative to get the `polkadot` binary 🤔

harrysolovay: @pepoviola any recommendation? Should we include a runner-compatible Polkadot bin in the Capi repo? Any thoughts would be greatly appreciated.

effect/extrinsic.test.ts
harrysolovay: Very interesting –– seems I have some reproducing to do :)

test-util/config.ts
```diff
@@ -52,18 +52,21 @@ export async function config(props?: NodeProps): Promise<Config> {
         ...props?.altRuntime ? [`--force-${props.altRuntime}`] : [],
       ],
       stderr: "piped",
-      stdout: "piped",
     });
+
     // For some reason, logs come in through `stderr`
     console.log(blue(`Piping node logs:`));
+
+    const encoder = new TextDecoder();
```
harrysolovay, line 60: The initialization cost of `TextDecoder`s is quite low. Imo not worth the overhead of a const definition.

.github/workflows/test.yml
pepoviola: Hey @harrysolovay / @kratico, getting the binary from the `latest` image is a nice way to get the binary. Other way is to download the binary from the `polkadot` repo, as we do in the setup command in zombienet ( https://github.com/paritytech/zombienet/blob/main/src/cli.ts#L120-L141).

Looking forward, we have the task to create a `zombienet action` to publish to the marketplace, then you will be able to just use the action and receive a `json` with the network info (relaychain validators and parachains).
Thanks!

harrysolovay: Thank you for the input @pepoviola –– will look forward to that action! :)

effect/extrinsic.test.ts
```diff
@@ -0,0 +1,72 @@
+import { assertEquals, assertObjectMatch } from "../deps/std/testing/asserts.ts";
```
kratico, line 1: @harrysolovay Do you have a suggestion for this file location/name?

It seems to be more an integration test than a unit test.
Perhaps it would be better to create a new folder `./e2e-test/..` in the project root.

Thoughts?

test-util/config.ts
```diff
@@ -31,16 +31,14 @@ export class Config
...
   if (props?.port) {
     if (!isPortAvailable(props.port)) {
       fail(`Port ${props.port} is unavailable`);
     }
     port = props.port;
   } else {
-    while (!isPortAvailable(port)) {
-      port++;
-    }
+    port = getRandomPort();
```
kratico, line 41: @harrysolovay added a random port so it's less likely to have the race condition in parallel tests

test-util/config.ts
```diff
@@ -64,7 +61,14 @@ export async function config(props?: NodeProps): Promise<Config> {
         break;
       }
     }
-    return new Config(port, process.close.bind(process), props?.altRuntime);
+    return new Config(
+      port,
+      () => {
+        process.kill("SIGKILL");
+        process.close();
```
kratico, line 68: `.close` releases Deno’s memory associated with the process but it doesn't `.kill` the process

effect/sys/run.ts
```diff
@@ -65,6 +65,7 @@ export const { run } = (new class Runtime implements RunContext {
         if (val.exit) {
           const applied = () => val.exit!(resolved);
           cleanup.set(val, applied);
+          this.#cache.delete(k);
```
kratico, line 68: @harrysolovay perhaps this is fixed in your refactor.
This removes the `RpcClient` atom from the cache because it has an `exit` fn.

Is the following statement correct?
_Atoms that have been cleaned up should no longer be cached_

harrysolovay: All of this will change considerably in the next few days. Will look forward to syncing about this with you & @jainkrati.

effect/extrinsic.test.ts
```diff
@@ -0,0 +1,150 @@
+import type { KeyringPair } from "https://deno.land/x/polkadot@0.0.8/keyring/types.ts";
+import { assertEquals, assertObjectMatch } from "../deps/std/testing/asserts.ts";
+import * as C from "../mod.ts";
+import * as t from "../test-util/mod.ts";
+import * as U from "../util/mod.ts";
+import { SendAndWatchExtrinsicProps } from "./mod.ts";
```
harrysolovay, line 6: You can delete this line and then access `SendAndWatchExtrinsicProps` from the namespace below:

```diff
async function collectExtrinsicEvents(
  { config, palletName, methodName, args }: Pick<
-   SendAndWatchExtrinsicProps,
+   C.SendAndWatchExtrinsicProps,
```

effect/extrinsic.test.ts
```diff
@@ -0,0 +1,150 @@
...
+      );
+
+      assertEquals(extrinsicEvents, ["ready", "inBlock", "finalized"]);
+    });
+
+    config.close();
+  },
+});
+
+async function collectExtrinsicEvents(
```
harrysolovay, line 105: This is an awesome utility. Mentioning in #215.

test-util/config.ts
```diff
@@ -31,16 +30,14 @@ export class Config
 
 export async function config(props?: NodeProps): Promise<Config> {
   if ("_browserShim" in Deno) return new Config(9944, () => {});
-  let port = 9944;
+  let port;
```
harrysolovay, line 33:
```suggestion
  let port: number;
```

test-util/config.ts
```diff
@@ -89,3 +86,47 @@ function isPortAvailable(port: number): boolean {
     throw error;
   }
 }
+
+function getRandomPort(min = 49152, max = 65534): number {
+  let randomPort;
```
harrysolovay, line 91:
```suggestion
  let randomPort: number;
```

test-util/config.ts
```diff
@@ -89,3 +86,47 @@ function isPortAvailable(port: number): boolean {
...
+
+  do {
+    randomPort = Math.floor(Math.random() * (max - min + 1) + min);
+  } while (!isPortAvailable(randomPort));
+
+  return randomPort;
+}
+
+interface WaitForPortOptions {
```
harrysolovay, line 100: These options don't seem to be supplied to the `waitForPort` call. Any reason to keep this declaration and the optional param on 107?

harrysolovay (changes requested):
Looking pretty good! Two thoughts:

- Might be good to add some comments + related issues for anything that is temporary (such as the semi-thread-safe means of getting an open port)
- The `isPortAvailable` also calls `close` without actually killing the process (thank you for pointing out this issue). Can you please sig-kill it as well?

test-util/config.ts
kratico: You are right.
These options can be re-added when needed, so I'll remove them.

kratico:
> * Might be good to add some comments + related issues for anything that is temporary (such as the semi-thread-safe means of getting an open port)

Will do

> * The `isPortAvailable` also calls `close` without actually killing the process (thank you for pointing out this issue). Can you please sig-kill it as well?

The `.close` in `isPortAvailable` is for closing a `Deno.listener` not a `Deno.Process`.
Searched for related [issues](https://github.com/denoland/deno/issues/2358) and it seems that `.close` is enough to free up listener resources.

effect/extrinsic.test.ts
```diff
@@ -0,0 +1,150 @@
...
+    createWatchHandler(stop) {
+      return (event) => {
+        if (typeof event.params.result === "string") {
+          extrinsicEvents.push(event.params.result);
+        } else {
+          if (event.params.result.inBlock) {
+            extrinsicEvents.push("inBlock");
+          } else if (event.params.result.finalized) {
+            extrinsicEvents.push("finalized");
+            stop();
```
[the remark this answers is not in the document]
kratico: I double checked it and the `stop()` needs to be called explicitly or `await root.run()` will not resolve/reject.

Is there an extrinsic event that signals that there will be no more events?

harrysolovay (approved):
LGTM –– please add any lessons / areas for improvement to #215

PR #225, opened 2022-09-15
Create secretscheck.yml

### LINKED ISSUE #36: Task that scans for secrets / errors out during pre-commit

jainkrati:
Resolves #36 

In order to [check for accidentally leaked secrets/api keys/passwords](https://github.com/paritytech/capi/issues/36), I checked out some existing solutions which we could integrate in our repo:
- Github Action [trufflehog](https://github.com/trufflesecurity/trufflehog)
- CLI Tool [git-secrets](https://github.com/awslabs/git-secrets)
- CLI Tool [repo-security-scanner](https://github.com/techjacker/repo-security-scanner)
- CLI Tool [gitrob](https://github.com/michenriksen/gitrob)
- Github's [secret-scanning](https://docs.github.com/en/code-security/secret-scanning/about-secret-scanning)

Considering the number of stars, frequency of update and ease of integration, I have integrated trufflehog for now. However, we can discuss more and I can update the PR accordingly.

### DIFF
--- a/.github/workflows/secretscheck.yml
+++ b/.github/workflows/secretscheck.yml
@@ -1,0 +1,14 @@
+name: Secrets check
+on:
+  pull_request:
+      - created
+      - edited
+jobs:
+  echo_issue_comment:
+    runs-on: ubuntu-latest
+    name: secrets check
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: TruffleHog OSS
+        uses: trufflesecurity/trufflehog@v3.10.3
--- a/.github/workflows/secretscheck.yml
+++ b/.github/workflows/secretscheck.yml
@@ -1,6 +1,7 @@
 name: Secrets check
 on:
   pull_request:
+    types:
       - created
       - edited
 jobs:
--- a/words.txt
+++ b/words.txt
@@ -171,3 +171,4 @@ Curto
 devcontainers
 noninteractive
 autoremove
+trufflesecurity


.github/workflows/secretscheck.yml
```diff
@@ -0,0 +1,14 @@
+name: Secrets check
+on:
+  pull_request:
+      - created
```
tjjfvi, line 4:
```suggestion
    types:
      - created
```

harrysolovay (approved):
The Trufflehog action looks tremendous. Does it protect us from instances such as the following?

`forgot-to-gitignore.ts`

```ts
export const ALICE_PUBLIC = "d43593c715fdd31c61141abd04a99fd6822c8558854ccde39a5684e7a56da27d";
export const ALICE_SECRET = "e5be9a5092b81bca64be81d212e7f2f9eba183bb7a90954f7b76361f6edb5c0a";
```

My hope is that––when PRed––this would result in a failing check, and a swift force push to the fork/branch could eliminate the secret from its history.

> Sidenote: it'd be nice to have a git hook that does the same check locally pre-commit.
> Sidenote: it'd be nice to scan the entire repo history as well.

In cases where we do want to commit/expose the secret (Alice is a commonly-known pair, after all), we could somehow allow it / explicitly pass validation.

It's unclear whether Trufflehog addresses this precaution (perhaps it does). Also, it's unclear whether this precaution can be appropriately addressed in the first place; after all, how does a program distinguish between a keypair secret vs. a scale-encoded hex, vs. some other string?.

Regardless of the answers to these questions, I think this PR is generally positive! Worse comes to worse, we can adjust the description of / back-log #36.

jainkrati:
Updates :
1. We add Trufflehog GA and after few weeks evaluate if Trufflehog is serving our purpose or do we need to replace it with a better one. Created an event for 4th November for the same.
2. To exclude files for scanning (in order to push in some test secrets on purpose), we can specify --exclude-paths=EXCLUDE-PATHS as part of extra_args:  https://github.com/trufflesecurity/trufflehog/blob/main/README.md#usage
3. I will manually rebase commit history to remove all commits which introduced any kinds of secrets. Going forward Github action will take care of preventing secrets in code.

Issue #47, opened 2022-06-02
approach to retries


harrysolovay:
When an RPC server disconnects unexpectedly, do we attempt to reconnect, re-send any egress messages, misc? How about for transactions? If the user has encoded the transaction with a nonce and window of validity, yes. What if they have not? Should users need to explicitly opt into retry behavior? Should they specify their own back-off algorithm? Should they specify their own fallback logic beyond simply re-sending? A lot of questions to answer regarding approach to retries.

kratico:
Message retry seems to depend on the message type.

Perhaps messages can be categorized in idempotent vs non-idempotent messages (like REST) depending on the operation that they trigger.

For example,
- in a subscription to `chain_subscribeNewHead`, a subscription retry should be fine
- in a `author_submitAndWatchExtrinsic`, an extrinsic retry should not be fine but a retry to monitor the extrinsic events should be fine

This leads me to think that the message producer needs some context to decide if a `call`/`subscription` can be retried.

Then, for message delivery strategies, there could be messages that are
- fire and forget
- fire and acknowledge success/failure
- fire, retry x times and acknowledge success/failure
- fire, timeout and ...

harrysolovay:
I completely agree with everything you wrote. Especially...

> Perhaps messages can be categorized in idempotent vs non-idempotent messages

... and...

> the message producer needs some context to decide if a `call`/`subscription` can be retried.

I think it's up to us to ensure that the necessary context is accessible to the developer, who can use it alongside their own monitoring/resubmission logic. Mocking out some example scenarios would be a good idea... assuming we think message-specific retries is an early priority (not so sure that it is––what're your thoughts?).

However... for fun, let's play this out a bit more. Your 4 bullet points highlight some key considerations, such as `how do we identify whether a non-idempotent message has been received?`, `what stage it is in?`, `do we need to await confirmation that the message did not go through before resubmission?`.

As I write out these questions, the hard-coding of use-case specific handling seems undesirable––even for std effects. Even if we do it "right," it will be "wrong" for some consumers. Probably best to steer clear of most message-related retries for now.

kratico:
> As I write out these questions, the hard-coding of use-case specific handling seems undesirable––even for std effects. Even if we do it "right," it will be "wrong" for some consumers. Probably best to steer clear of most message-related retries for now.

Yes, it seems that message retry should not be responsibility of the RPC client.
For now, it would be better to throw an error (as we currently do) and let the developer do the resubmission depending on the use-case.
[...]

PR #235, opened 2022-09-19
chore: RPC refactor

kratico:
**Description**

Add a `WsContainer` class to manage `WebSocket` reconnects.

The `WsContainer.send`, before sending a message to the `WebSocket, performs the following checks
- if there is no active connection, it attempts to create a new connection
- if the `WebSocket` connection attempt is in progress, it waits for the connection to be in the `WebSocket.OPEN` state
- if there is an active connection, use the active connection
- if the connection attempts are exhausted, it returns `FailedToOpenConnectionError`

Also, based on feedback, the `ClientHooks` were removed from the `RPCClient` abstract class. 

There are some open issues regarding `WebSocket` (or provider) disconnection, for example
- if the `WebSocket` is disconnected, What should be the behavior for inflight call/subscriptions?
- ~if the `WebSocket` is not connected, What should be the behavior for an attempt to send a message?~

Related #47 

**How Has This Been Tested?**

There are 2 unit tests that validate the WebSocket reconnection when
- the client internal WebSocket instance is closed
- the server WebSocket is closed


rpc/providers/proxy.test.ts
```diff
@@ -5,8 +5,6 @@ import { proxyClient } from "./proxy.ts";
 
 Deno.test({
   name: "Proxy RPC Client",
-  sanitizeResources: false,
-  sanitizeOps: false,
```
kratico, line 9: This seems to be the issue with WebSocket and Deno resources
```
Deno.test({
  name: "this test leaks",
  ignore: true,
  fn() {
    const ws = new WebSocket(polkadot.discoveryValue);

    ws.close();
  },
});

Deno.test({
  name: "this test doesn't leak",
  async fn() {
    const ws = new WebSocket(polkadot.discoveryValue);

    const isClosed = deferred();
    ws.onclose = () => isClosed.resolve();

    ws.close();

    await isClosed;
  },
});
```

rpc/providers/proxy.ts
```diff
@@ -3,44 +3,38 @@ import { deadline, deferred } from "../../deps/std/async.ts";
...
-    ws.addEventListener("open", onOpen);
-  } else {
-    pending.resolve(client);
+  const wsIsOpen = deferred();
+  const ws = new WebSocketProxy({
+    webSocketFactory: () => new WebSocket(config.discoveryValue),
+  });
+
+  ws.onopen = () => wsIsOpen.resolve();
+  ws.onerror = () => wsIsOpen.reject();
```
kratico, line 20: This works only for the first connection attempt.

kratico: This seems to be needed to send messages only when there is an open connection.
If that's the only use case, then, it might be better to handle message sending with a non-open connection in the RPC `Client` base class.
The reason is that the `Client` has enough context to decide what to do when the provider can not send the message.
The client could filter messages and decide for each case something like
- buffer a non-sent messages
- error for messages that can not be re-send
- discard messages

rpc/providers/web-socket-proxy.test.ts
```diff
@@ -0,0 +1,114 @@
...
+      },
+    });
+  },
+});
+
+/**
+ * Starts a WebSocket server that closes the WebSocket connection on the first message
+ */
+async function startWebSocketServer(listener: Deno.Listener) {
+  async function handleConn(conn: Deno.Conn) {
```
harrysolovay, line 94: Why not inline this in the `for await` loop?

rpc/providers/web-socket-proxy.test.ts
```diff
@@ -0,0 +1,114 @@
...
+    const httpConn = Deno.serveHttp(conn);
+    for await (const e of httpConn) {
```
harrysolovay, line 96:
```diff
- const httpConn = Deno.serveHttp(conn);
- for await (const e of httpConn) {
+ for await (const e of Deno.serveHttp(conn)) {
```

rpc/providers/web-socket-proxy.test.ts
```diff
@@ -0,0 +1,114 @@
...
+      e.respondWith(handle(e.request));
+    }
+  }
+
+  function handle(req: Request) {
```
harrysolovay, line 101: Same thing here––could be an anonymous arrow fn passed to `handle`?

rpc/providers/web-socket-proxy.ts
```diff
@@ -0,0 +1,120 @@
...
+  maxAttempts?: number;
+  delay?: number;
+};
+
+type WebSocketClientOptions = {
+  webSocketFactory: () => WebSocket;
+  reconnect?: ReconnectOptions;
+};
+
+export class WebSocketProxy implements Pick<WebSocket, "send" | "close"> {
```
harrysolovay, line 12: We don't need to utilize `WebSocketProxy` in slots expecting something conforming to the Ws interface. Why the `implements` clause?

rpc/providers/web-socket-proxy.ts
```diff
@@ -0,0 +1,120 @@
...
+      };
+    }
+  }
+
+  #connect() {
+    this.#isClosed = false;
+    this.#isReconnecting = false;
+    this.#webSocket = this.options.webSocketFactory();
+
+    this.#webSocket.onopen = this.#onWSOpen;
```
harrysolovay, line 71: In situations like this, we prefer `#onWsOpen`, as the `WS` abbreviation is considered a term unto itself (hence we camelize). Same thing for `onWSMessage`->`onWsMessage` and others.

test-util/config.ts
```diff
@@ -92,7 +92,7 @@ function isPortAvailable(port: number): boolean {
   }
 }
 
-function getRandomPort(min = 49152, max = 65534): number {
+export function getRandomPort(min = 49152, max = 65534): number {
```
harrysolovay, line 95: Given this is utilized elsewhere, want to extract it out of `test-util/config.ts` and into `test-util/port.ts`?

rpc/providers/web-socket-proxy.ts
harrysolovay: I don't think the name `WebSocketProxy` is fitting, given that we also use the word proxy to refer to our conduit to the on-chain world. Might I recommend `WsContainer`? Any other ideas?

harrysolovay (changes requested):
This is great progress in the direction of the recovery story we want for devs––great work @kratico.

You bring up an especially important point: decoupling certain pieces of state (such as inflight message IDs and `Deferred`s) from the underlying WebSocket instance. Let's run with this!

Had a few comments.

One non-line-specific comment: let’s prefer the Rust/Deno convention for file names (snake). `web-socket-proxy.ts` -> `ws_container.ts`

rpc/providers/proxy.ts
```diff
@@ -3,44 +3,38 @@ import { deadline, deferred } from "../../deps/std/async.ts";
...
-    ws.addEventListener("error", onOpenError);
-    ws.addEventListener("open", onOpen);
-  } else {
-    pending.resolve(client);
+  const wsIsOpen = deferred();
+  const ws = new WebSocketProxy({
+    webSocketFactory: () => new WebSocket(config.discoveryValue),
+  });
+
+  ws.onopen = () => wsIsOpen.resolve();
```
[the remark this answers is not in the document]
kratico: I wanted to keep the API very slim but I guess it would be more friendly with add/remove listeners.

Initially, I thought that this should re-dispatch the inner WebSocket events and clone those events as follows
```
const et = new EventTarget();

const ws = new WebSocket("wss://rpc.polkadot.io");

ws.addEventListener("open", (e) => {
  // this throws a DOMException InvalidStateError
  et.dispatchEvent(e);
  // cloning the event doesn't throw
  et.dispatchEvent(new e.constructor(e.type, e));
});

et.addEventListener("open", (e: Event) => {
  console.log(e);
});

```

However, this container could have its own events that don't need to match exactly to the WebSocket events.

I'll change this container to extend from `EventTarget`

rpc/providers/proxy.ts
kratico: The RPC `Client` calls the `Provider.send` assuming that there will be no errors when a message is sent.
I believe that we need to change that assumption and change this signature
```
  /**
   * The provider-specific send implementation
   *
   * @param egressMessage the message you wish to send to the RPC server
   */
  send: (egressMessage: msg.InitMessage<Config_>) => void;
```
to
```
  send: (egressMessage: msg.InitMessage<Config_>) => void | SendError;
```

Then, the `Client` could do
```
  send = (egressMessage: msg.InitMessage<Config_>): void | SendError => {
    this.hooks?.send?.(egressMessage);
    try {
      this.provider.send(egressMessage);
    } catch (e) {
      return new SendError(e);
    }
  };
```

rpc/providers/proxy.ts
```diff
@@ -3,44 +3,38 @@ import { deadline, deferred } from "../../deps/std/async.ts";
...
-  return pending;
+
+  return new ProxyClient(ws, hooks);
 }
 
 export class ProxyClient<Config_ extends Config>
   extends B.Client<Config_, MessageEvent, Event, FailedToDisconnectError>
 {
-  constructor(ws: WebSocket, hooks?: ProxyClientHooks<Config_>) {
+  constructor(ws: WebSocketProxy, hooks?: ProxyClientHooks<Config_>) {
```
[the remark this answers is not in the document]
kratico: Right now, hooks are not used.
And there is a mapping between a message request and a successful/failed response.

However, when the WebSocket connection is closed, the inflight requests are not completed with an error (or I'm not able to spot this in the current Client implementation).

Perhaps, this could be handed by the `Client`, Thoughts?

rpc/providers/web-socket-proxy.test.ts
```diff
@@ -0,0 +1,114 @@
...
+
+        await isOpen;
+
+        const isReopen = deferred();
+        webSocketProxy.onopen = () => isReopen.resolve();
+
+        const isClosed = deferred();
+        webSocketProxy.onclose = () => isClosed.resolve();
+
+        firstConnection!.close();
```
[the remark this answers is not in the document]
kratico: It might be better to change this to a `spy`.
Any thoughts on 
```
import sinon from "https://cdn.skypack.dev/sinon";
```

rpc/providers/proxy.ts
harrysolovay: Agreed. Is it worthwhile to parameterize the type of the send error independently?

rpc/providers/proxy.ts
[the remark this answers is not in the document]
harrysolovay: I may be missing something––I don't see why we need an `EventTarget`. Can you please elaborate from a slightly higher level?

rpc/providers/proxy.ts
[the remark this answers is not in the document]
harrysolovay: +1 to removing hooks. The hook-in approach isn't nearly as ergonomic as separating inflight message / dispatch state into a higher-level container that holds onto the `WsContainer` (or whatever we end up calling it). So yes: I think your approach makes a lot of sense: `Client` should hold onto the given state; perhaps through a `ClientState` / something reusable across providers. Then, retrying is simply about detaching from the provider (in this case `WsContainer`) and ensuring the new callbacks are hooked up to the existing provider.

rpc/providers/proxy.ts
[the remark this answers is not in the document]
kratico: It would be to reuse the add/remove event listener functionality.

I noticed that the `B.Client` implements its own event listening
https://github.com/paritytech/capi/blob/11106d7b46159657e5dbf3765688ddbce360cbad/rpc/Base.ts#L54-L60

Perhaps, there is a reason to not extend from `EventTarget` or `std/node/events.ts` `EventEmitter`.
Thoughts?

rpc/providers/web-socket-proxy.test.ts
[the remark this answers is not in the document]
harrysolovay: Sounds reasonable :) which functionality do you want to stub?

rpc/providers/web-socket-proxy.ts
kratico: WsContainer sounds good

rpc/providers/web-socket-proxy.test.ts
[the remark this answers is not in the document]
kratico: I would use a spy rather than a stub.
So, I would be able to spy on the WebSocketFactory return value and invoke `.close`.
Also spying on the factory would let me assert that it was invoked 2 times due to a disconnection.

rpc/providers/proxy.ts
```diff
@@ -1,46 +1,28 @@
...
-    pending.resolve(client);
-  }
-  return pending;
+  const ws = new WsContainer({
+    webSocketFactory: () => new WebSocket(config.discoveryValue),
+  });
+
+  return new Promise((resolve) => {
+    ws.once("open", () => resolve(new ProxyClient(ws)));
+    ws.once("error", () => resolve(new FailedToOpenConnectionError()));
```
kratico, line 18: I'm still not very convinced with this.
One or the other will stay attached.
Perhaps something like this would be better https://deno.land/std@0.156.0/node/events.ts?s=once

or 
```
ws.once(["open", "error"], (event) => ...);
```

rpc/Base.ts
```diff
@@ -64,18 +60,18 @@ export abstract class Client<
   onMessage: OnMessage<RawIngressMessage> = (message) => {
     const parsed = this.provider.parseIngressMessage(message);
     if (parsed instanceof Error) {
-      this.hooks?.error?.(parsed);
+      // FIXME: define behavior for provider.parseIngressMessage errors
```
kratico, line 63: cc @harrysolovay

rpc/Base.ts
```diff
@@ -108,7 +104,11 @@ export abstract class Client<
         }
       };
     });
-    this.send(init);
+    const sendResult = this.send(init);
+    if (sendResult instanceof Error) {
+      // FIXME: ingressMessagePending should be resolved to a non-RPC Error
+      // and stopListening() should be invoked
```
kratico, line 110: cc @harrysolovay

rpc/providers/proxy.ts
```diff
@@ -1,46 +1,30 @@
...
-      ws.removeEventListener("open", onOpen);
-    };
-    ws.addEventListener("error", onOpenError);
-    ws.addEventListener("open", onOpen);
-  } else {
-    pending.resolve(client);
+  const ws = new WsContainer({
+    webSocketFactory: () => new WebSocket(config.discoveryValue),
+  });
+  const event = await ws.once(["open", "error"]);
```
kratico, line 13: Got some inspiration from https://deno.land/std@0.156.0/node/events.ts?s=once

rpc/providers/proxy.ts
kratico: I got the suggestion and added `SendError` to `Client` type parameters

effect/atoms/RpcCall.ts
```diff
@@ -23,7 +23,9 @@ export function rpcCall<
         methodName,
         params as Parameters<(Methods & rpc.ProviderMethods)[T_<MethodName>]>,
       );
-      if (result.error) {
+      if (result instanceof Error) {
+        // FIXME: What would be the error for InternalError (non RPC response errors)?
```
kratico, line 27: cc @harrysolovay

effect/atoms/RpcSubscription.ts
```diff
@@ -27,13 +28,16 @@ export function rpcSubscription<
...
+        createErrorListener,
         cleanup
           ? (x) => {
             return this.run(cleanup(x.result));
           }
           : undefined,
       );
-      if (result?.error) {
+      if (result instanceof Error) {
+        // FIXME: What would be the error for InternalError (non RPC response errors)?
```
kratico, line 39: cc @harrysolovay

rpc/Base.ts
```diff
@@ -91,24 +95,39 @@ export abstract class Client<
...
+      .then((sendResult) => {
+        if (sendResult instanceof Error) {
           stopListening();
-          ingressMessagePending.resolve(res as IngressMessage);
+          ingressMessagePending.resolve(sendResult);
         }
-      };
-    });
-    this.send(init);
+      });
```
kratico, line 130: the `async/await` was causing an error in the extrinsic integration tests.
For those tests, there seem to be some synchronization issue where the first subscription message is lost.

rpc/Base.ts
```diff
@@ -123,40 +142,59 @@ export abstract class Client<
...
-      );
-      return (res) => {
-        if (res.params?.subscription && res.params.subscription === initRes.result) {
-          listenerCb(res as msg.NotifMessage<Config_, MethodName>);
-        } else {
-          // TODO: associate errors with subscriptions & exit
-        }
-      };
-    });
+    this.listen(
```
harrysolovay, line 157: This will change depending on whatever final interface we settle on –– will postpone reviewing this portion

rpc/providers/proxy.ts
```diff
@@ -54,31 +32,43 @@ export class ProxyClient<Config_ extends Config>
...
+  #initContainer() {
+    let index = 0;
+    const factory = ({ discoveryValue }: Config_) => {
+      if (typeof discoveryValue === "string") {
+        return new WebSocket(discoveryValue);
+      }
+      index %= discoveryValue.length;
+      return new WebSocket(discoveryValue[index++]!);
+    };
+    this.#container = new WsContainer({ client: this, factory });
```
harrysolovay, line 64: Did we decided to invert this?

harrysolovay (commented):
It's getting there. Let's temporarily pause development on this PR and model the north star interface, type-only. Let's do this in [TS playground](https://www.typescriptlang.org/play).

rpc/Base.ts
```diff
@@ -10,34 +10,30 @@ export type OnMessage<Message> = (message: Message) => void;
 export abstract class Client<
   Config_ extends Config,
   RawIngressMessage,
-  InternalError,
+  InternalError extends Error,
   CloseError extends Error,
 > {
   #nextId = 0;
   #listenerCbs = new Map<U.WatchHandler<msg.IngressMessage<Config_>>, true>();
+  #errorListenerCbs = new Map<U.WatchHandler<InternalError>, true>();
```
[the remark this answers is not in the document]
kratico: A `WeakMap` is not iterable so it doesn't seem to be an alternative.

_tasks/download_frame_metadata.ts
```diff
@@ -13,6 +13,9 @@ await Promise.all(
       assert(!(client instanceof Error));
       try {
         const metadata = await client.call("state_getMetadata", []);
+        if (metadata instanceof Error) {
```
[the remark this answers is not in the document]
kratico: great suggestion!

However, What should happen if the user does not handle the error and an error is thrown (for example a WebSocket error)?
If I understood the above suggestion, an unhandled error will not be unioned with the `.call` result type.

Would that runtime error need to be part of the `.call` result type?

rpc/Base.ts
```diff
@@ -91,24 +95,39 @@ export abstract class Client<
...
+          stopListening();
+          ingressMessagePending.resolve(error);
+        },
+    );
+    this.send(init)
+      .then((sendResult) => {
+        if (sendResult instanceof Error) {
           stopListening();
-          ingressMessagePending.resolve(res as IngressMessage);
+          ingressMessagePending.resolve(sendResult);
```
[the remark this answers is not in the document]
kratico: I found the issue around this
Check https://github.com/paritytech/capi/pull/276/files#r986010430

harrysolovay:
> It's getting there. Let's temporarily pause development on this PR and model the north star interface, type-only. Let's do this in [TS playground](https://www.typescriptlang.org/play).

In its simplest form: [here's a though](https://www.typescriptlang.org/play?#code/C4TwDgpgBAwgThAhsCB1ZBjAFjARgHgFEA3CAO2AD4oBeKACgBMJhFsAuBgSluuIHsAlox41q6YNjxFSFSgG4AUKEhQJUgiXJVaDCLOCctFUXyGMligPQAqG4qg2oAFSzQwcfsWEQ4UQRS+AGZs0ADOWPwArgA2jFC40IhhUAC2AYKpiDFQyVBg-GFhgrgxEAB0UACSwFAR0XFQGPwUiAFQLTEgUMBuCYgIUBBF2oLZYeUONlaKAShwIRjQAAqe3sxw+A5QACKCYc2kcCAANNswLWQQGMCEcJ5+EAAeKGSMKXcPZ1BQAMrkjFW-DAYW+fwBn34jxeAI+9yhYOMtVsUAABkjUVAomRBC1-GQMDEoswUnNfGRskN4XAUtMwXsDpdrrdqUMYW84V9FNQAN7bZpkK43ThMfaHXwgTgM8XHUxQIHpMIQfACYRQAA+sCZN0hcAU2wA7pgsCKMAhkGhjXhOPAkCh1DhNAZKHLVRZtkq3iKPMCwpx-m8gSC5Qr9sq3RrwW9dfqfowxdrDNxePLPIrw+ZI9LEzGlABfRSF2z2RwuPqomAxQTaTEYRBkXK4MLAOBsWpeXxooHrXyYg2CXo9LABADmKSrAGtoAgBULgLiyCc0XWYjFMfX4qiwlEm2aShBUZNSzUmvW0m1Wu1mxaoEEoVBiANcVEUr1R6C6lFsLkUqlhmFEBHCAAFowkgDBBCCQQMCgLANzKGlcjeKAEBbasJimGYyQWUJYCrbQth+C5BWZXU2Ved4oF1MEA0BTwQRoiFWWeCjOQRbYkTOXl+WyGJOHwAAxXByNhKABN43A2AnfBiLnaio0YGNKHobYfh9EF-QBINQVU28JKkzghLBEM0zDGRtEjAAlFgojgMhnHAZUhMofUPR3A44H3fihJEjkxP0jBpNk0jqSXWilJUn41IGRBUk0wN6J0qKmnNe0rVwG1UstSRHXMuQwR+EJV0kwLDNwYyU1DJU8tqTVrOAWz7McwTcBc-NCxUaBxOKqTCK1EidWY9lKOo7ZwqG1iqNC7ldHoXwHhtHNWU1caHjlbEJzIfgDTIJQgA). This doesn't touch on the factories that produce providers/clients, but it does describe their raw shape/separation of concern. @kratico are we thinking about this similarly?

harrysolovay:
Closing in favor of #341

PR #240, opened 2022-09-22
feat: add RPC client fallback URLs

kratico:
**Description**

Fixes #

**How Has This Been Tested?**

### DIFF
--- a/rpc/providers/proxy.ts
+++ b/rpc/providers/proxy.ts
@@ -4,11 +4,22 @@ import * as B from "../Base.ts";
 import { FailedToSendMessageError, ParseRawIngressMessageError } from "../common.ts";
 import { WsContainer } from "./ws_container.ts";
 
-export async function proxyClient<Config_ extends Config<string>>(
+export async function proxyClient<Config_ extends Config<string | string[]>>(
   config: Config_,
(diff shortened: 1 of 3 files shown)


rpc/providers/proxy.ts
```diff
@@ -4,11 +4,22 @@ import * as B from "../Base.ts";
...
+        return new WebSocket(discoveryValue);
+      }
+      index %= discoveryValue.length;
+      return new WebSocket(discoveryValue[index++]!);
+    };
+  };
   const ws = new WsContainer({
-    webSocketFactory: () => new WebSocket(config.discoveryValue),
+    discoveryValue: config.discoveryValue,
+    webSocketFactory: createWebSocketFactory(),
```
kratico, line 22: This will work for reconnects but not for initial connections.

This makes me think that, it would be easier to model the reconnect as a stream of attempts to reconnect.
Once the attempts are exhausted, the following should happen
- if it's creating the `WsContainer`, it should fail with `FailedToOpenConnectionError`
- if there are messages waiting for a connection, it should fail with `FailedToSendMessageError`

The above could be simplified if a connection attempt is only triggered when there is a message to be sent.

kratico: cc @harrysolovay

PR #256, opened 2022-10-02
chore: generate simpler lookups from ss58 registry

harrysolovay:
Should be useful for codegen purposes.

See lookups [here](https://github.com/paritytech/capi/blob/4400b66390eeb2bbcd85f826e415166adbb06d99/known/lookups.ts).


tjjfvi:
I don't understand why this needs to be codegenned at all. Couldn't `lookups.ts` just be
```ts
import ss58Registry from "../deps/ss58_registry.ts";

export interface ChainInfo {
  prefix: number;
  network: string;
  displayName: string;
  symbols: string[];
  decimals: number[];
  standardAccount: string | null;
  website: string | null;
}

export const lookup: Record<string, ChainInfo> = Object.fromEntries(
  ss58Registry.map((obj) => [obj.network, obj]),
);
```

harrysolovay:
Would be a tiny bit slower. Couldn't help but micro-optimize... in retrospect, the complexity of the build step might make this less worthwhile. Thoughts?

tjjfvi:
Very tiny, and it's only ever run once. If anything, I would say optimize it by including the *specific* object in the codegen output (which would be as simple as `JSON.stringify(lookup[network])`)

harrysolovay:
Closing in light of #260

PR #278, opened 2022-10-04
[WIP]Multisig examples

jainkrati:
Resolves #

- [ ] The title conforms to [the conventional commit standard](https://www.conventionalcommits.org/en/v1.0.0/)

### DIFF
--- a/examples/multisig_creation.ts
+++ b/examples/multisig_creation.ts
@@ -1,0 +1,18 @@
+import * as C from "../mod.ts";
+
+declare const PRIMARY: Uint8Array;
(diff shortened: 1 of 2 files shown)


tjjfvi:
I think we should have a separate place for "future" examples, i.e. examples that should eventually work but do not currently, so that all the files under `examples` can continue to be used as manual tests of sorts, at least until we have a better solution.

harrysolovay:
I agree with @tjjfvi –– also, please open PRs in draft if they're not ready for a (potentially-approving) reviewer.

PR #281, opened 2022-10-04
chore: remove PR title format checkbox

harrysolovay:
We have a GH check for this now


tjjfvi (approved):
Tests are failing, but that seems to be a thing sometimes now

harrysolovay:
> Tests are failing, but that seems to be a thing sometimes now

I believe this is an RPC issue. Soon to be addressed (#235).

Issue #215, opened 2022-09-12
Expand Test Utils + Zombienet


harrysolovay:
Let's model the Zombienet configuration (here is [the grammar's new/WIP AST](https://github.com/paritytech/zombienet/blob/4f0501c0f56b4d2fcfa41466257471b55a8bd519/crates/parser/src/ast.rs)) and integrate it into `test-util`.

cc @pepoviola

---

Ongoing list of test feature ideas:

- running dev-chain-spawning tests in parallel
- `collectExtrinsicEvents` (introduced in #211) –– perhaps with a slightly different interface.
- nice console messaging from test ctx / spawned Polkadot processes

kratico:
For testing, it would be nice to specify a node configuration
For example, compiling the polkadot node with `cargo build --release --features fast-runtime`, creates a binary with shorter Launch/Voting/Enactment periods (see [polkadot](https://github.com/paritytech/polkadot/blob/2b0d83258089578b65712e143c794cf4922941de/runtime/polkadot/src/lib.rs#L590-L600) repo).

I'm not sure if it would be possible to get a new configuration with a runtime upgrade.

pepoviola:
> For testing, it would be nice to specify a node configuration to speed up some operations.
> 
> For example, compiling the polkadot node with `cargo build --release --features fast-runtime`, creates a binary with shorter Launch/Voting/Enactment periods (see [polkadot](https://github.com/paritytech/polkadot/blob/2b0d83258089578b65712e143c794cf4922941de/runtime/polkadot/src/lib.rs#L590-L600) repo).
> 
> I'm not sure if it would be possible to get a new configuration with a runtime upgrade.

Using `fast-runtime` in our CI is part of this [pr](https://github.com/paritytech/polkadot/pull/5978/files#diff-037ea159eb0a7cb0ac23b851e66bee30fb838ee8d0d99fa331a1ba65283d37f7R220). That will produce an image (polkadot-debug) that we can use in zombienet with the fast-runtime.

kratico:
Another thing to consider is launching polkadot nodes in parallel for testing.

Doing #211, the following was discovered
- `polkadot --dev --ws-port [port]` may not start to listen in the selected port if some other process is listening in that port and it will chose a different port.
  A workaround could be to make start a named polkadot node and make an rpc request to get the node name to validate that the node was started successfully in the specified port.
- killing a `Deno.Process` that started a polkadot non-detached docker container does not stop the container
  ```
[...]

PR #291, opened 2022-10-12
feat: add multisig example

kratico:
Add a multi-signature transfer example.

The example does the following
- derive a multi-signature address for Alice, Bob and Charlie with 2 as a threshold
- Alice transfers some balance to the multi-signature address
- Alice creates a proposal to transfer to Dave from the multi-signature address
- Bob approves the proposal to transfer to Dave from the multi-signature address
- log Dave's new account balance to the console


examples/multisig_transfer.ts
```diff
@@ -0,0 +1,121 @@
...
+        value: sender.sign(message),
+      };
+    },
+    createWatchHandler,
+  });
+
+const _getTimepoint = () => {
+  return C.readEntry(config, "Multisig", "Multisigs", [
+    ABC_MULTISIG_ADDRESS,
+  ]);
```
kratico, line 93: The timepoint is required for any approval after the proposal creation

image

frame_metadata/Key.ts
```diff
@@ -28,23 +28,39 @@ export function $storageKey(props: StorageKeyProps): $.Codec<unknown[]> {
...
-    ),
+  const $keys = [...Array(keyCodecs.length + 1).keys()].reduce(
+    (acc, i) => {
+      const codec = $.tuple(
+        ...keyCodecs.slice(0, i).map(($key, i) =>
+          H[(props.storageEntry as M.MapStorageEntryType).hashers[i]!].$hash($key)
+        ),
+      );
+      const keySize = keyCodecs.slice(0, i).reduce(
+        (acc, $key) => acc + ($key._metadata![1] as number),
```
kratico, line 39: This is the key param size
There may be a better API to get this info

frame_metadata/Key.ts
```diff
@@ -28,23 +28,39 @@ export function $storageKey(props: StorageKeyProps): $.Codec<unknown[]> {
...
-    ),
+  const $keys = [...Array(keyCodecs.length + 1).keys()].reduce(
+    (keys, i) => {
+      const codec = $.tuple(
+        ...keyCodecs.slice(0, i).map(($key, i) =>
+          H[(props.storageEntry as M.MapStorageEntryType).hashers[i]!].$hash($key)
+        ),
+      );
+      const keySize = keyCodecs.slice(0, i).reduce(
+        (size, $key) => size + ($key._metadata![1] as number),
```
kratico, line 39: This is the key param size
There may be a better API to get this info

examples/multisig_transfer.ts
kratico: Depends on https://github.com/paritytech/capi/issues/294

examples/multisig_transfer.ts
```diff
@@ -0,0 +1,150 @@
...
+if (!hostname || !portRaw) {
+  throw new Error("Must be running inside a test ctx");
+}
+
+import { createKeyMulti } from "../deps/polkadot/util-crypto.ts";
+import * as C from "../mod.ts";
+import * as T from "../test_util/mod.ts";
+import * as U from "../util/mod.ts";
+
+type KeyringPair = typeof T.alice;
```
harrysolovay, line 13: Please delete this line and instead specify the `Uint8Array` and `C.M.SignExtrinsic` types where applicable.

examples/multisig_transfer.ts
```diff
@@ -0,0 +1,150 @@
...
+    }))
+    .watch(function(status) {
+      console.log(status);
+      if (C.TransactionStatus.isTerminal(status)) {
+        this.stop();
+      }
+    });
+}
+
+interface CreateOrApproveProposalProps {
```
harrysolovay, line 86: Given that this interface isn't reused, should we inline this constraints of `createOrApproveMultisigProposal`? Ie.

```diff
function createOrApproveMultisigProposal(
- { sender, maybe_timepoint }: CreateOrApproveProposalProps,
+ publicKey: Uint8Array,
+ sign: C.M.SignExtrinsic,
+ maybeTimepoint?: bigint,
) {
```

examples/multisig_transfer.ts
```diff
@@ -0,0 +1,150 @@
...
+    })
+      .signed((message) => ({
+        type: "Sr25519",
+        value: sender.sign(message),
+      }));
+  };
+  const withoutActualWeight = createEx().extrinsic;
+  const weight = C.payment.queryInfo(T.polkadot)(withoutActualWeight)
+    .access("weight");
+  C.Z.call(weight, (d) => {
```
harrysolovay, line 140: Not sure if this serves the example (as much as debugging the example). Can we get rid?

rpc/client.ts
```diff
@@ -53,10 +53,11 @@ export class Client<
     } else if (e.id) {
       const pendingCall = this.pendingCalls[e.id];
       if (!pendingCall) {
-        console.log({ e });
+        console.log("!pendingCall for", { e });
         // TODO: pipe error to listeners and message the likely cause,
         //       a duplicate client.
-        throw new Error();
+        // throw new Error();
+        return;
```
harrysolovay, line 60: Is not throwing a good idea? No pending call would signal faulty behavior.

test_util/local.ts
```diff
@@ -53,6 +53,9 @@ export class LocalClientEffect extends Z.Effect<LocalClient, PolkadotBinNotFound
   }
 
   private async createClient(): Promise<LocalClient> {
+    // TODO: remove logging
+    console.log("creating client...");
```
harrysolovay, line 57: Added `nice console messaging from test ctx / spawned Polkadot processes` to our ongoing list in #215.

harrysolovay (changes requested):

examples/multisig_transfer.ts
kratico: right now, the `weight` is hardcoded, I'll comment it out until there is a better way to inject effects in the extrinsic params

rpc/client.ts
kratico: will revert

examples/multisig_transfer.ts
```diff
@@ -0,0 +1,163 @@
...
+
+// Create proposal
+U.throwIfError(
+  await createOrApproveMultisigProposal({
+    publicKey: T.alice.publicKey,
+    sign: C.Z.call(0, () =>
+      (message) => ({
+        type: "Sr25519",
+        value: T.alice.sign(message),
+      })),
```
kratico, line 35: @harrysolovay I found a weird issue.
If a plain function is used this doesn't work.
It seems that the result of the first `sign()` call is cached so on the 2nd extrinsic (when bob approves the proposal) the signature check fails

harrysolovay: For now, let's just use the test ctx and call each with `run` (wihout first binding to an env).

harrysolovay (approved):

PR #293, opened 2022-10-12
chore: fix failing RPC test

harrysolovay:
The RPC test was failing. This is likely due to the fact that it made use of a remote RPC node, which may rate limit / is sometimes unavailable. This PR changes the test to utilize a local node instead.


kratico (approved):
LGTM 🚀

Issue #226, opened 2022-09-16
All-in on Codegen DX


harrysolovay:
The following proposal describes a would-be critical shift in the Capi DX. Feedback would be greatly appreciated.

---

Currently, Capi utilizes `Config`s for its discovery values and typings. This is true of both the static and dynamic DXs.

```ts
class Polkadot extends Config.from<
  PolkadotCallMethods,
  PolkadotSubscriptionMethods,
  PolkadotErrorDetails,
  PolkadotMetadata
>() {
  constructor() {
    super("wss://rpc.polkadot.io", 0);
  }
};
const polkadot = new Polkadot();
```

The idea is that the types within a given `Config` will inform the rest of Capi usage. While the narrow typing is worthwhile, the current execution is overly-complex: `Config` is somewhere in between static and dynamic; it has one foot in both worlds, but fits into neither. Yes, we can generate `Config`s with FRAME-specific types, but no, this does not result in the optimal static DX.

```ts
import * as C from "capi";
import { config } from "./generated.ts";

const value = await chain(config).pallet("SomePallet").entry("SomeEntry").read();
```

> Note: we cannot attach method/storage-specific TSDocs unless microsoft/TypeScript#41165 is addressed.

The one notable advantage to this experience is that it is consistent; if the chain is not known ahead of time, one can apply `any` to the config's type params. However, I'd argue that the static use case (where target chain is known ahead of time) is far more common. If we optimize for this use case, the experience can be dramatically simplified.

```ts
import { SomePallet } from "./generated.ts";

const value = await SomePallet.SomeEntry.read();
```

Meanwhile, the dynamic use case can be satisfied via direct usage of std effects.

```ts
import * as C from "capi";

const value = await C.readEntry(
  C.config("wss://some.chain.io", 0),
  "SomePallet",
  "SomeEntry",
).read();
```

> Note: config creation could be much simpler without the need for a named signature containing RPC, error and FRAME types.

In addition to offering simpler DX, a key benefit of this approach is the ease of adjusting to runtime upgrades. Simply regenerate the bindings and fix type errors.

```sh
deno run -A -r https://deno.land/x/capi/codegen.ts \
  --out="polkadot.ts" \
  --src="wss://rpc.polkadot.io"
```

kratico:
This would be super nice.

Apollo has something similar https://www.the-guild.dev/graphql/codegen

kratico:
This is another DX example for inspiration https://trpc.io/

PR #296, opened 2022-10-14
feat: reworking test util

harrysolovay:
Reworks our approach to spawning test chains. Before, we were spawning in the same place that we were using. This excessive spawn/kill slowed down test execution. For our testing purposes, isolated test networks don't necessarily matter. If/when this becomes a requirement, we can manage this within `test_ctx` as well.

To run a network-dependent command, prefix the command with usage of the `test_ctx` CLI.

```
deno task run test_ctx.ts <your command here>
```

The test_ctx.ts starts a server, which is responsible for opening polkadot CLI processes on-demand. It also runs your command, passing in the server's hostname and port as env vars. Configs within `test_util/configs` use these env vars to request their respective runtimes' dev net ports. The subprocess then uses the config as normal. Upon the subprocess's exit, the test ctx process kills all polkadot process instances.

Sounds like a mouthful, I know. So why this approach?

The main reason is to support our codegen vision #226. Developers will be able to test their code with ease.

`main.ts`

```ts
import { polkadotPattern } from "https://deno.land/x/some-polkadot-pattern-lib/mod.ts";

const root = polkadotPattern(/* some usage here */);
```

Normally, the developer would run this with something such as `deno run -A main.ts`

To make use of the test env and configs, they need not touch the actual code:

```
deno run -A https://deno.land/x/capi/test_ctx.ts \
  deno run -A main.ts --import-map=./test_config_remappings.json
```

---

Related changes:

- Moved the codegen CLI entrypoint to the root. Felt necessary given that we now expose a root-level `test_ctx.ts` CLI as well
- Test util import alias `t` -> `T` (was conflicting with `t` context exposed from Deno.test)


test_util/config.ts
```diff
@@ -1,131 +1,73 @@
...
+          requiresSetupError();
+        }
+        conn.write(new Uint8Array([TestConfigRuntime.CODES[runtimeName]]));
+        const port = await (async () => {
+          for await (const x of conn.readable) {
+            return new DataView(x.buffer).getUint16(0);
+          }
+          return undefined!;
+        })();
+        return `ws://127.0.0.1:${port}`;
```
kratico, line 27: Shall the TCP connection be closed at this time?

kratico:
```
conn.close()
```

test_ctx.ts
```diff
@@ -0,0 +1,115 @@
...
+
+const cmdProcess = Deno.run({
+  cmd: Deno.args,
+  env: {
+    TEST_CTX_HOSTNAME: listener.addr.hostname,
+    TEST_CTX_PORT: listener.addr.port.toString(),
+  },
+});
+await cmdProcess.status();
+cleanup();
```
kratico, line 27:
```suggestion
const status = await cmdProcess.status();
cleanup();

Deno.exit(status.code);
```

kratico (dismissed):
LGTM 🚀

kratico (approved):

kratico (approved):

kratico (approved):

PR #300, opened 2022-10-17
feat: add example GH action

### LINKED ISSUE #299: Add GH action to run examples in parallel
### Suggestion

Run examples on pull requests and run them in parallel

### Motivation

Examples are chain interactions that can be treated as pseudo integration tests.

Therefore, running them in CI improves the quality.

kratico:
Resolves #299 

- [x] There is an associated issue (**required**)
- [x] The change is described in detail
- [x] There are new or updated tests validating the change (if applicable)

## Description

Run examples in parallel using reusable workflows.
The caller workflow `spawn_examples.yml` generates the list of example paths and calls the `example.yml` reusable workflow.
The `example.yml` workflow uses a marix strategy to create a job for each example file.

Also, this PR adds a `setup-polkadot` composite action that caches and downloads the polkadot binary.
I'd like to get some feedback on the composite action before integrating it in other workflows

### DIFF
--- a/.github/workflows/example.yml
+++ b/.github/workflows/example.yml
@@ -1,0 +1,21 @@
+name: Run example
+on:
+  workflow_dispatch:
+    inputs:
+      example_path:
+        type: text
+        required: true
+jobs:
+  run_example:
+    name: Run example
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: denoland/setup-deno@004814556e37c54a2f6e31384c9e18e983317366 # v1.1.0
+        with:
+          deno-version: v1.x
+      - name: Download polkadot
+        run: |
+          curl -L -o /usr/local/bin/polkadot https://github.com/paritytech/polkadot/releases/download/v0.9.28/polkadot
+          chmod +x /usr/local/bin/polkadot
+      - run: deno task run test_ctx.ts deno task run ${{ github.event.inputs.example_path }}
--- a/.github/workflows/example.yml
+++ b/.github/workflows/example.yml
@@ -10,7 +10,6 @@ on:
       example_paths:
         type: string
         required: true
-  push:
 jobs:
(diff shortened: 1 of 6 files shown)


.github/workflows/example.yml
```diff
@@ -0,0 +1,33 @@
...
+      - uses: actions/checkout@v3
+      - uses: denoland/setup-deno@004814556e37c54a2f6e31384c9e18e983317366 # v1.1.0
+        with:
+          deno-version: v1.x
+      - name: Cache Deno dependencies
+        uses: actions/cache@v3
+        with:
+          path: |
+            ~/.cache/deno
+          key: ${{ runner.os }}-build-${{ hashFiles('lock.json', 'deps/**/*.ts') }}
```
kratico, line 30: added `deps/*` to the hash because `deps/*` it's not always in sync with the `lock.json` file.

.github/workflows/example.yml
```diff
@@ -0,0 +1,33 @@
...
+      - name: Setup Polkadot
+        uses: ./.github/actions/setup-polkadot
+      - run: deno task run test_ctx.ts deno task run ${{ matrix.example_path }}
```
kratico, line 33: What do you think about adding a deno task `run:test_ctx` or similiar?

harrysolovay: I suppose we could, although it doesn't do much for us.

```
deno task run:test_ctx
```

vs.

```
deno task run test_ctx.ts
```

kratico: Something like
```
deno task run:test_ctx ${{ matrix.example_path }}
```
Instead of
```
deno task run test_ctx.ts deno task run ${{ matrix.example_path }}
```

harrysolovay: Resolved by #304

harrysolovay (approved):

PR #305, opened 2022-10-21
feat: add example GH action (with zones)

### LINKED ISSUE #299: Add GH action to run examples in parallel

kratico:
Resolves #299 

- [x] There is an associated issue (**required**)
- [x] The change is described in detail
- [x] There are new or updated tests validating the change (if applicable)

## Description

Run examples in parallel using reusable workflows.
The caller workflow `spawn_examples.yml` generates the list of example paths and calls the `example.yml` reusable workflow.
The `example.yml` workflow uses a marix strategy to create a job for each example file.

Also, this PR adds a `setup-polkadot` composite action that caches and downloads the polkadot binary.
I'd like to get some feedback on the composite action before integrating it in other workflows


kratico:
covered in #300

PR #306, opened 2022-10-21
chore: publish using `npm_publish_automation`

tjjfvi:


.github/workflows/release.yml
```diff
@@ -22,8 +22,17 @@ jobs:
...
+      - uses: actions/upload-artifact@v3
+        with:
+          name: package
+          path: "./target/npm/*.tgz"
+      - uses: octokit/request-action@v2.x
         if: startsWith(github.ref, 'refs/tags/')
+        with:
+          route: POST /repos/paritytech/npm_publish_automation/actions/workflows/publish.yml/dispatches
+          ref: main
+          inputs: '${{ format(''{{ "repo": "{0}", "run_id": "{1}" }}'', github.repository, github.run_id) }}'
```
kratico, line 36: It may be easier to use the `gh` cli

```
echo '{...}' | gh workflow run -R paritytech/npm_publish_automation ...
```

harrysolovay (approved):

.github/workflows/release.yml
tjjfvi: Perhaps, but this is tested and working, so I'm inclined to leave it as is

PR #311, opened 2022-10-21
feat: add .ignore support for spawn_example workflow

kratico:
Switch `find` with `fd-find` to support `.ignore` files inside `./examples`

https://github.com/sharkdp/fd#installation


harrysolovay (approved):
LGTM!

PR #315, opened 2022-10-22
chore: block and entry effect cleanup

harrysolovay:
Attempts to simplifies the APIs for reading and watching blocks and storage


harrysolovay:
Was a nice experiment, but I don't know if this is necessary for now.

effect/Block.ts
```diff
@@ -0,0 +1,88 @@
...
+import * as U from "../util/mod.ts";
+import { $extrinsic } from "./core/$extrinsic.ts";
+import { deriveCodec } from "./core/deriveCodec.ts";
+import { Metadata } from "./Metadata.ts";
+import { RpcCall } from "./RpcCall.ts";
+import { RpcSubscription } from "./RpcSubscription.ts";
+import { run } from "./run.ts";
+
+export namespace Block {
+  export function read<Rest extends [blockHash?: Z.$<U.HexHash | undefined>]>(
```
tk-o, line 14: Quick question: why did you use a namespace here to group some set of static functions?

Alternatively, a class with static members could be used instead (and seems a more straight-forward choice, at least for someone who's not familiar with TS namespaces).

Thank you in advance for sharing an answer 🙃

harrysolovay: While the overhead of a defining a class would be negligible, it still feels *extra*; we don't instantiate anything. We just want the intermediate name. Namespaces aren't going anywhere anytime soon, so it shouldn't be of concern. Still, you're right to prompt the question of whether this is an outdated practice.

PR #317, opened 2022-10-23
chore: update update deno dependencies workflow

tjjfvi:
Should hopefully fix the failing "basic tests" thing


tjjfvi:
It's worth noting that the failing "basic tests" thing affects `main` at times (namely, when `udd` runs but doesn't open a PR):
image

harrysolovay (approved):

Issue #245, opened 2022-09-27
RPC Client call and subscribe errors


kratico:
**Is your feature request related to a problem? Please describe.**

The current RPC Client captures errors encoded in the RPC responses.
With #235 the RPC Client could capture and return the following errors
- `.call`/`subscribe` error when it's not possible to make a WebSocket connection
- `.call`/`subscribe` error when the message was sent to the WebSocket but the connection get's closed before the `onmessage` event is triggered

**Describe the solution you'd like**

When the above errors are returned, complete the `.call`/`.subscribe` with an error and stop listening for `onmessage`.

Inflight `.call`/`.subscribe` calls can be closed by having the `WsContainer` sending an error message to the RPC Client `.onMessage(ConnectionErrorMessage)` when a `WebSocket` connection is closed. 

This will require to update `.onMessage` type signature to `OnMessage<RawIngressMessage | ConnectionErrorMessage>`

Most of the refactor will be in these lines

https://github.com/paritytech/capi/blob/c63e6252e92fefbf1e3093ce0b829445eb5acbb0/rpc/Base.ts#L97-L112

https://github.com/paritytech/capi/blob/c63e6252e92fefbf1e3093ce0b829445eb5acbb0/rpc/Base.ts#L128-L150

kratico:
Closing as the new client implementation cover these errors

PR #339, opened 2022-10-28
Feat/rpc client

kratico:
Resolves #

### DIFF
--- a/rpc_new/client.ts
+++ b/rpc_new/client.ts
@@ -10,8 +10,28 @@ import {
 
 export class Client<DiscoveryValue, SendE, InternalE, CloseE> {
   providerRef;
-  listeners = new Set<U.Listener<msg.IngressMessage | InternalE>>();
+  // Is this used?
+  // listeners = new Set<U.Listener<msg.IngressMessage | InternalE>>();
   pendingCalls: Record<string, Deferred<unknown>> = {};
+  pendingSubscriptions: Record<
+    string,
+    U.Listener<
+      | msg.NotificationMessage
+      | msg.ErrorMessage
+      | ProviderSendError<SendE>
+      | ProviderHandlerError<InternalE>
+    >
+  > = {};
+  activeSubscriptions: Record<
+    string,
+    U.Listener<
+      | msg.NotificationMessage
+      | msg.ErrorMessage
+      | ProviderSendError<SendE>
+      | ProviderHandlerError<InternalE>
+    >
+  > = {};
+  activeSubscriptionByMessageId: Record<string, string> = {};
 
   constructor(
     readonly provider: Provider<DiscoveryValue, InternalE, SendE, CloseE>,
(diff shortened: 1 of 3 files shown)


rpc_new/provider/proxy.ts
```diff
@@ -70,38 +69,34 @@ function ensureWsOpen(ws: WebSocket): Promise<void | Event> {
     return Promise.resolve();
   }
   return new Promise<void | Event>((resolve) => {
-    const openHandler = () => {
-      removeOpenHandlers();
+    const controller = new AbortController();
+    const { signal } = controller;
```
kratico, line 73: with the `AbortController` we don't need a listener ref to remove the listener

rpc_new/client.ts
```diff
@@ -45,6 +80,23 @@ export class Client<DiscoveryValue, SendE, InternalE, CloseE> {
...
+    const stop = () => {
+      delete this.pendingSubscriptions[message.id];
+      const activeSubscriptionId = this.activeSubscriptionByMessageId[message.id];
+      if (activeSubscriptionId) {
+        delete this.activeSubscriptions[activeSubscriptionId];
+      }
+    };
+    this.pendingSubscriptions[message.id] = createListenerCb(stop);
+    this.call(message);
+  };
```
kratico, line 98: I'll rename this to `subscribe` and remove the method from line 83

kratico, line 98: I'll rename this to `subscribe` and remove the method from line 83

rpc_new/client.ts
```diff
@@ -26,13 +46,28 @@ export class Client<DiscoveryValue, SendE, InternalE, CloseE> {
...
       delete this.pendingCalls[e.id];
+      if (this.pendingSubscriptions[e.id]) {
+        if (e.error) {
+          this.pendingSubscriptions[e.id]!(e);
+        } else {
+          this.activeSubscriptions[e.result] = this.pendingSubscriptions[e.id]!;
+          this.activeSubscriptionByMessageId[e.id] = e.result;
+        }
+        delete this.pendingSubscriptions[e.id];
+      }
```
kratico, line 68: there are many `if`s here but for subscriptions we need to filter the initialization message so we can setup the listener for the actual subscription messages

rpc_new/client.ts
```diff
@@ -26,13 +46,28 @@ export class Client<DiscoveryValue, SendE, InternalE, CloseE> {
         const pendingCall = this.pendingCalls[id]!;
         pendingCall.resolve(e);
         delete this.pendingCalls[id];
+        this.pendingSubscriptions[id]?.(e);
+        delete this.pendingSubscriptions[id];
+      }
+      for (const id in this.activeSubscriptions) {
+        this.activeSubscriptions[id]?.(e);
+        delete this.activeSubscriptions[id];
```
kratico, line 54: as long as the provider communicate errors through the listeners, this will cover https://github.com/paritytech/capi/issues/245

PR #341, opened 2022-10-28
feat: RPC client re-write

kratico:
Resolves #


rpc_new/client.ts
```diff
@@ -0,0 +1,152 @@
...
+        if (e.error) {
+          this.pendingSubscriptions[e.id]!(e);
+        } else {
+          this.activeSubscriptions[e.result] = this.pendingSubscriptions[e.id]!;
+          this.activeSubscriptionByMessageId[e.id] = e.result;
+        }
+        delete this.pendingSubscriptions[e.id];
+      }
+    } else if (e.params) {
+      this.activeSubscriptions[e.params.subscription]?.(e);
```
kratico, line 56: @harrysolovay I just reverted this to `?.` because after calling `.stop()` from inside a subscription, the inner WebSocket could still emit an event for the stopped subscription.

words.txt
```diff
@@ -186,4 +186,12 @@ tostring
...
 statemigration
 xxhash
+statemine
+westmint
 chainspec
+chainspecs
+contramap
+merkle
+dispatchable
+xxhash
```
tjjfvi, line 196: xxxxhash

rpc_new/util.ts
```diff
@@ -0,0 +1,12 @@
+// TODO: expose these utils from `util/mod.ts`
```
tjjfvi, line 1: Any reason not to do this now?

rpc_new/provider/base.ts
```diff
@@ -0,0 +1,71 @@
...
+export type ProviderListener<SendErrorData, HandlerErrorData> = U.Listener<
+  | msg.IngressMessage
+  | ProviderSendError<SendErrorData>
+  | ProviderHandlerError<HandlerErrorData>
+>;
+
+export interface ProviderRef<CloseErrorData> {
+  nextId(): string;
+  send(message: msg.EgressMessage): void;
+  release(): Promise<void | ProviderCloseError<CloseErrorData>>;
```
tjjfvi, line 23:
```suggestion
  release(): Promise<undefined | ProviderCloseError<CloseErrorData>>;
```

rpc_new/provider/base.ts
```diff
@@ -0,0 +1,71 @@
...
+}
+
+declare const providerListener_: unique symbol;
+type providerListener_ = typeof providerListener_;
+
+export abstract class ProviderConnection<Inner, SendErrorData, HandlerErrorData> {
+  declare [providerListener_]: ProviderListener<SendErrorData, HandlerErrorData>;
```
tjjfvi, line 30: This makes the class invariant on these type parameters

rpc_new/provider/base.ts
```diff
@@ -0,0 +1,71 @@
...
+    if (this.listeners.has(listener)) {
+      return;
+    }
+    this.listeners.set(
+      listener,
+      listener.bind({
+        stop: () => {
+          this.listeners.delete(listener);
+        },
+      }),
```
tjjfvi, line 51: Kind of odd

rpc_new/provider/base.ts
```diff
@@ -0,0 +1,71 @@
...
+export type Provider<DiscoveryValue, SendErrorData, HandlerErrorData, CloseErrorData> = (
+  discoveryValue: DiscoveryValue,
+  listener: ProviderListener<SendErrorData, HandlerErrorData>,
+) => ProviderRef<CloseErrorData>;
+
+export type ProviderListener<SendErrorData, HandlerErrorData> = U.Listener<
+  | msg.IngressMessage
+  | ProviderSendError<SendErrorData>
+  | ProviderHandlerError<HandlerErrorData>
+>;
```
tjjfvi, line 18:
```suggestion
export type ProviderListener<SendErrorData, HandlerErrorData> = U.Listener<
  | msg.IngressMessage
  | ProviderSendError<SendErrorData>
  | ProviderHandlerError<HandlerErrorData>,
  { stop: () => void }
>;
```

rpc_new/provider/base.ts
```diff
@@ -0,0 +1,71 @@
...
+export interface ProviderRef<CloseErrorData> {
+  nextId(): string;
+  send(message: msg.EgressMessage): void;
+  release(): Promise<void | ProviderCloseError<CloseErrorData>>;
+}
+
+declare const providerListener_: unique symbol;
+type providerListener_ = typeof providerListener_;
+
+export abstract class ProviderConnection<Inner, SendErrorData, HandlerErrorData> {
```
tjjfvi, line 29: Why is this abstract?

rpc_new/known.ts
```diff
@@ -0,0 +1,45 @@
+import * as known from "../known/rpc/mod.ts";
+import * as U from "../util/mod.ts";
+import { call, subscription } from "./effects.ts";
+
+// TODO: generate the following?
```
tjjfvi, line 5: Once this is merged I'll rework this with my full set of known types

rpc_new/util.ts
```diff
@@ -0,0 +1,12 @@
...
+
+export type Listener<Event, This = any> = (this: This, event: Event) => void;
+
+export function contramapListener<From, Into>(
```
tjjfvi, line 5: Should probably pass `this` through

rpc_new/client.ts
```diff
@@ -0,0 +1,169 @@
...
+        pendingCall.resolve(e);
+        delete this.pendingCalls[id];
+        this.pendingSubscriptions[id]!(e);
+        delete this.pendingSubscriptions[id];
+      }
+      for (const id in this.activeSubscriptions) {
+        this.activeSubscriptions[id]!(e);
+        delete this.activeSubscriptions[id];
+        this.subscriptionStates.delete(id);
+      }
```
tjjfvi, line 52: Seems pretty nuclear. Do errors never correspond to particular calls / subscriptions?

rpc_new/provider/proxy.ts
```diff
@@ -0,0 +1,92 @@
...
+      return Promise.resolve();
+    },
+  };
+};
+
+function connection(
+  url: string,
+  listener: ProviderListener<Event, Event>,
+): ProxyProviderConnection {
+  let conn = connections.get(url);
```
tjjfvi, line 42: Should use `getOrInit` util

rpc_new/provider/proxy.ts
```diff
@@ -0,0 +1,92 @@
...
+      conn!.forEachListener(new ProviderHandlerError(e));
+    }, controller);
+    conn = new ProxyProviderConnection(ws, () => controller.abort());
+    connections.set(url, conn);
+  }
+  conn.addListener(listener);
+  return conn;
+}
+
+function ensureWsOpen(ws: WebSocket): Promise<void | Event> {
```
tjjfvi, line 59: Lots of `void | Foo` here – should be `undefined | Foo`

rpc_new/provider/smoldot.ts
```diff
@@ -0,0 +1,92 @@
...
+      } catch (_e) { /* TODO */ }
+    });
+    connections.set(chainSpec, conn);
+    const loop = async () => {
+      try {
+        conn!.forEachListener(msg.parse(await inner.nextJsonRpcResponse()));
+      } catch (e) {
+        conn!.forEachListener(new ProviderHandlerError(e));
+        await loop();
+      }
```
tjjfvi, line 82: Looks like right now each `loop` will only ever output one response?

```suggestion
      while (true) {
        try {
          conn!.forEachListener(msg.parse(await inner.nextJsonRpcResponse()));
        } catch (e) {
          conn!.forEachListener(new ProviderHandlerError(e));
        }
      }
```

rpc_new/provider/smoldot.ts
```diff
@@ -0,0 +1,92 @@
...
+    };
+    // TODO: should we be greedy & kick off a few additional?
+    //       We don't want the inner buffer to overflow.
+    loop();
+    loop();
+    loop();
```
tjjfvi, line 88: Why three?

rpc_new/provider/smoldot.ts
```diff
@@ -0,0 +1,92 @@
...
+async function connection(
+  chainSpec: string,
+  listener: ProviderListener<SmoldotSendErrorData, SmoldotHandlerErrorData>,
+): Promise<SmoldotProviderConnection> {
+  if (!client) {
+    client = start(
+      <ClientOptions> {
+        forbidTcp: true,
+        forbidNonLocalWs: true,
+      },
```
tjjfvi, line 63:
```suggestion
      {
        forbidTcp: true,
        forbidNonLocalWs: true,
      } as ClientOptions,
```

rpc_new/provider/smoldot.ts
```diff
@@ -0,0 +1,92 @@
...
+    nextId,
+    send: (message) => {
+      (async () => {
+        const conn = await connection(chainSpec, listener);
+        try {
+          return conn.inner.sendJsonRpc(JSON.stringify(message));
+        } catch (e) {
+          return new ProviderSendError(e);
+        }
+      })();
```
tjjfvi, line 36: These errors are getting sent to the void

rpc_new/provider/smoldot.ts
```diff
@@ -0,0 +1,92 @@
...
+
+const nextId = nextIdFactory();
+
+export const smoldotProvider: Provider<
+  string,
+  SmoldotSendErrorData,
+  SmoldotHandlerErrorData,
+  SmoldotCloseErrorData
+> = (chainSpec, listener) => {
+  return {
```
tjjfvi, line 26:
```suggestion
  const connProm = connection(chainSpec, listener);
  return {
```

rpc_new/provider/smoldot.ts
```diff
@@ -0,0 +1,92 @@
...
+    nextId,
+    send: (message) => {
+      (async () => {
+        const conn = await connection(chainSpec, listener);
```
tjjfvi, line 30:
```suggestion
        const conn = await connProm;
```

rpc_new/provider/smoldot.ts
```diff
@@ -0,0 +1,92 @@
...
+        try {
+          return conn.inner.sendJsonRpc(JSON.stringify(message));
+        } catch (e) {
+          return new ProviderSendError(e);
+        }
+      })();
+    },
+    release: async () => {
+      const { cleanUp, listeners, inner } = await connection(chainSpec, listener);
```
tjjfvi, line 39:
```suggestion
      const { cleanUp, listeners, inner } = await connProm;
```

tjjfvi (changes requested):

rpc_new/util.ts
harrysolovay: It'll break effects that use an existing `Listener`/`CreateListener`. Will do this in a follow-up PR.

rpc_new/provider/base.ts
harrysolovay: What problems do you foresee this causing? It's just there for convenience. Can get rid if you think it'll be problematic.

rpc_new/provider/base.ts
harrysolovay: Good way or bad? The alternative is the existing interface `(stop: () => void) => (e: E) => void`.

rpc_new/provider/base.ts
tjjfvi: It's not *necessarily* a problem, but it is counter-semantic.

rpc_new/provider/base.ts
harrysolovay: Causes type errors elsewhere

rpc_new/provider/base.ts
harrysolovay: It's the lowest-common denominator between providers: a container for cbs, a connection & some cleanup. Any concerns?

rpc_new/util.ts
harrysolovay: True... come to think of it, I'm not sure if we'll end up using this util. `Listener` is much simpler now (the mapping requires little code) and we'll likely want to customize the context depending on the effect (aka. `apply({ /* customizing this */ }, [...])`). Still, will pass it through for now.

rpc_new/client.ts
harrysolovay: Not send/handler/close errors. If we encounter those, there's an internal error and it's likely best that we throw away the provider instance and restart––we don't currently tackle this recovery, but we certainly will. Thoughts @kratico?

rpc_new/provider/smoldot.ts
harrysolovay: No particular reason. Needs tweaking. This impl is far from complete.

rpc_new/provider/smoldot.ts
harrysolovay: How does one break out of this? Let's tackle this provider more in depth in a subsequent PR. This is haphazard atm.

rpc_new/provider/smoldot.ts
harrysolovay: Whoops. Thought I had a TODO note there.

rpc_new/provider/smoldot.ts
harrysolovay: Initialize the connection before a send is initiated? I suppose the effect system will take care of orchestration... still, not sure if this is preferable. Thoughts @kratico?

rpc_new/provider/base.ts
tjjfvi: Since there are no `abstract` methods or properties, it seems like it could just be a regular class. Then you wouldn't need to have empty subclasses of it.

rpc_new/provider/smoldot.ts
tjjfvi: Generally I'd expect creating a client to initialize a connection. But the main reason I did that is to avoid calling `connection` repeatedly (though it returns the same thing, there's no need for the extra calls)

rpc_new/client.ts
kratico: right now, if there is a network error we don't do any recovery.

Errors do not always correspond to particular calls, you could have 2 active subscriptions and get the WebSocket closed.

once this PR is merged we can re-visit connection/disconnection recovery.

rpc_new/provider/smoldot.ts
kratico: for now, I would leave it inside the `.send`
for recovery, it might be easier to re-initialize the connection on a new `.send` invocation

PR #351, opened 2022-11-02
chore: error cleanup

harrysolovay:
- Return instead of throw from Ss58 transcoders
- Create new error ctors without a bespoke util


kratico (approved):
LGTM

PR #366, opened 2022-11-08
chore: misc effect cleanup

harrysolovay:
- Got rid of duplicate effect
- Got rid of unnecessary intermediate effects that would call specific codec encoding on specific values... now we just use `encode`/`decode` from the `effects/core/scale.ts`
- Simplify some effect logic


kratico (approved):
LGTM

PR #368, opened 2022-11-09
feat: more codegen

### LINKED ISSUE #226: All-in on Codegen DX

tjjfvi:
Resolves #226 

Heavily WIP


codegen/serve.ts
```diff
@@ -0,0 +1,396 @@
...
+    return U.getOrInit(this.filesMemo, metadata, () => {
+      return codegen({
+        metadata,
+        clientDecl: chainUrl.startsWith("dev:")
+          ? `
+import { LocalClientEffect } from ${JSON.stringify(`/@${this.version}/test_util/local.ts`)}
+export const client = new LocalClientEffect(${JSON.stringify(chainUrl.slice(4))})
+          `
+          : `
+export const client = C.rpc.rpcClient(C.rpc.proxyProvider, ${JSON.stringify(chainUrl)})
```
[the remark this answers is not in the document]
tjjfvi: > as a path to a TS file with a default-exported provider impl

This won't work, as remote modules cannot import local modules. We can consider this further later.

.github/workflows/test.yml
```diff
@@ -12,5 +12,6 @@ jobs:
       - name: Setup polkadot
         uses: ./.github/actions/setup-polkadot
       - run: deno task lint
+      - run: deno task codegen
```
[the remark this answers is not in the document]
tjjfvi: Yes, it's needed to type-check the `examples/`

_tasks/codegen.ts
```diff
@@ -0,0 +1,24 @@
+import { FsCache } from "../codegen/cache.ts"
+import { CodegenServer } from "../codegen/serve.ts"
+import * as fs from "../deps/std/fs.ts"
+import * as path from "../deps/std/path.ts"
+
+const cacheDir = "target/codegen"
+await fs.emptyDir(path.join(cacheDir, "generated"))
+const cache = new FsCache(cacheDir)
+const port = 5646
```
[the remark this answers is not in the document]
tjjfvi: I think we should decide on a port to use for `capi.dev` local development and go with that. `5646` is just a random example; I'm not attached to it.

tjjfvi:
https://capi.dev/@ref:more-codegen/proxy/wss:rpc.polkadot.io/@v0.9.32/pallets/Session.ts

examples/mod.ts
```diff
@@ -0,0 +1,25 @@
+// This script runs all examples in sequence. We should ultimately delete this script...
+// ... but it's currently proving useful for local debugging.
+
+// moderate
```
harrysolovay, line 4:
```suggestion
// moderate --exclude multisig_transfer.ts
```

codegen/serve.ts
```diff
@@ -0,0 +1,396 @@
+// deno-lint-ignore-file require-await
+
+import { tsFormatter } from "../deps/dprint.ts"
```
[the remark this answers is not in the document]
tjjfvi: #423

_tasks/codegen.ts
```diff
@@ -0,0 +1,21 @@
+import { LocalCapiCodegenServer } from "../codegen/server/local.ts"
+import * as fs from "../deps/std/fs.ts"
+
+await fs.emptyDir("target/codegen/generated")
+const port = 5646
+const server = new LocalCapiCodegenServer()
```
harrysolovay, line 6: Should `LocalCodegenServer` (and its ancestors) accept the signal via props (as does `serve` from `http/server.ts`)?

codegen/genMetadata.ts
```diff
@@ -25,80 +31,92 @@ export function genMetadata(metadata: M.Metadata, decls: Decl[]) {
...
-    }
+      if (pallet.calls) {
+        const ty = pallet.calls as M.Ty & M.UnionTyDef
+        const isStringUnion = ty.members.every((x) => !x.fields.length)
+        for (const call of ty.members) {
+          const type = normalizeCase(call.name)
+          const typeName = typeVisitor.visit(ty)! + "." + type
+          const [params, data]: [string, string] = call.fields.length
+            ? call.fields[0]!.name
+              ? [`value: Omit<${typeName}, "type">`, `{ ...value, type: ${S.string(type)} }`]
```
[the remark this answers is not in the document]
tjjfvi: I think both solutions have merit; I'm inclined to leave it as-is for now, and we can reconsider later on.

codegen/server/cache.ts
```diff
@@ -0,0 +1,82 @@
...
+  async _list(prefix: string): Promise<string[]> {
+    const result = []
+    for await (const entry of Deno.readDir(path.join(this.location, prefix))) {
+      result.push(entry.name)
+    }
+    return result
+  }
+}
+
+export class InMemoryCache extends Cache {
```
harrysolovay, line 74: What are your plans for this?

codegen/server/capi_repo.ts
```diff
@@ -0,0 +1,96 @@
+import { PermanentMemo, TimedMemo } from "../../util/memo.ts"
+import { shaAbbrevLength } from "./git_utils.ts"
+import { CodegenServer } from "./server.ts"
+
+const tagsTtl = 60_000 // 1 minute
```
harrysolovay, line 5: Nit: any reason to deviate from constant naming convention of JS?

codegen/server/local.ts
```diff
@@ -0,0 +1,58 @@
+import { PermanentMemo } from "../../util/memo.ts"
+import { Cache, FsCache } from "./cache.ts"
+import { CapiCodegenServer } from "./capi_repo.ts"
+import { getModuleIndex } from "./git_utils.ts"
+
+export class LocalCapiCodegenServer extends CapiCodegenServer {
+  version
+  cache: Cache = new FsCache("target/codegen", this.abortController.signal)
```
harrysolovay, line 8: Why the explicit `Cache` typing? Removing doesn't appear to adversely impact star's typecheck.

harrysolovay (dismissed):

_tasks/codegen.ts
tjjfvi: I originally wrote it that way, but found it annoying to create a new abort controller every time.

codegen/server/cache.ts
tjjfvi: Hmm, thought I used it somewhere. Might still be useful for testing, so I'm going to leave it.

codegen/genMetadata.ts
```diff
@@ -25,80 +32,103 @@ export function genMetadata(metadata: M.Metadata, decls: Decl[]) {
...
-            ["{ return", data, "}"],
-          ],
-        })
+                : "$.tuple()",
+              getRawCodecPath(entry.value),
+            ]})`,
+        )
+      }
+      if (pallet.calls) {
+        const ty = pallet.calls as M.Ty & M.UnionTyDef
```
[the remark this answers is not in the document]
tjjfvi: `satisfies` wouldn't make sense in this context, as this is an unsafe downcast.

codegen/server/capi_repo.ts
tjjfvi: Life's too short for typing in SCREAMING_SNAKE_CASE.

Just personal preference; do you want me to change it?

codegen/server/local.ts
tjjfvi: At one point in time I had a test where I set it to be a different cache (`InMemoryCache`, pretty sure). I don't anymore, but future tests might make use of that, so I'm inclined to leave it.

tjjfvi: Actually, `codecs.test.ts` should be using `InMemoryCache` right now. I'll fix that.

codegen/server/cache.ts
tjjfvi: Actually, `codecs.test.ts` should be using `InMemoryCache` right now. I'll fix that.

codegen/genMetadata.ts
```diff
@@ -25,80 +32,103 @@ export function genMetadata(metadata: M.Metadata, decls: Decl[]) {
...
+`,
+      ]
+      for (const entry of pallet.storage?.entries ?? []) {
+        items.push(
+          makeDocComment(entry.docs)
+            + `export const ${entry.name} = new C.fluent.Storage(${[
+              "client",
+              S.string(entry.type),
+              S.string(entry.modifier),
+              S.string(pallet.name),
```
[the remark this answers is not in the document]
tjjfvi: Not sure I understand. Do you mean generate the pallet & entry name hashes ahead of time? Could certainly be worthwhile, but I think that's best revisited later.

harrysolovay (approved):

PR #371, opened 2022-11-09
Smoldot integration

wirednkod:
Smoldot Light client (and the integration with it) should not be treated as a provider - rather as a full node. To explain, comparing the light client, to any other provider that may use a `url` to connect to, smoldot requires time to start up before start responding to rpc requests. Sync is needed for specific requests but some rpc calls can receive a response even prior to full sync of the light client.

The approach here is mainly a "migration" (if I may say) from the substrate-connect code and logic. Tests in `smoldot.test.ts` file are 3 simpleones that covers the main cases of `AddChain`, `AddWellknownChain` and `Add a Parachain`.

In `assets` directory exist 4 chainspecs of the WellKnown Chains. These could/should be updated - meaning update the `lightSyncState` field of the existing chain spec with the value of spec that's been downloaded with CI script (following the logic of Substarte-connect), every X days (normally every one day). This process, updates the 'light client checkpoint' that exists inside the chainspec, giving the capability to Smoldot, to sync from that point and not not from an older one (the older the checkpoint - the more time smoldot needs to wrap sync).

`TODO`: 
- Discuss what else we may need to export/use from Smoldot;
- Ask about use cases that may not be too familiar to me concerning CAPI and smoldot;
- Clarify (with bold strong words) that testing is just experimental and "setTimeout" is not a permanent solution;


wirednkod:
Closing this as smoldot integration is done

PR #372, opened 2022-11-09
chore: curry extrinsic effect factory

harrysolovay:
A previous PR introduced currying of the (`client`) first param of certain effect factories. The thinking was that this would simplify codegen. One could effectively swap out a single statement in a generated file to inform the discovery value and RPC error signatures. Aka.

`generated.ts`

```ts
// codegen-replacement-start
declare const client: Effect<rpc.Client<any, any, any, any>, Error>;
// codegen-replacement-end

export const entryRead = C.entryRead(client);
```

Then, the user could import `generated.ts` and interact with the effect factory without needing to think about the client's provider type and discovery value.

This PR does the same for the extrinsic effect factory.


kratico (approved):
LGTM

PR #374, opened 2022-11-09
feat: improve smoldot integration

### LINKED ISSUE #376: Smoldot Integration
Improve smoldot integration based on @wirednkod 

Related to #371 

Tasks
- [ ] relay chain integration
- [ ] parachain integration
- [ ] chain spec fetching

kratico:
Resolves #376

## Description

Improve Smoldot integration for relay chains by
- capturing Smoldot errors
- mapping specific `.send` requests to specific `.send` calls

Smoldot parachain integration will be tacked in an upcoming PR

### DIFF
--- a/rpc/client.ts
+++ b/rpc/client.ts
@@ -77,7 +77,12 @@ export class Client<
   call: ClientCall<SendErrorData, HandlerErrorData> = (message) => {
     const waiter = deferred<ClientCallEvent<SendErrorData, HandlerErrorData>>();
     this.pendingCalls[message.id] = waiter;
-    this.providerRef.send(message);
+    try {
+      this.providerRef.send(message);
+    } catch (error) {
+      delete this.pendingCalls[message.id];
+      waiter.resolve(error as ProviderSendError<SendErrorData>);
+    }
     return waiter;
   };
 
@@ -103,7 +108,13 @@ export class Client<
         );
       },
     }) as ClientSubscribeListener<SendErrorData, HandlerErrorData>;
-    this.call(message);
+    this.call(message)
+      .then((maybeError) => {
+        if (maybeError instanceof Error) {
+          this.pendingSubscriptions[message.id]!(maybeError);
+          stop();
+        }
+      });
     return waiter;
   };
 
--- a/rpc/client.ts
+++ b/rpc/client.ts
@@ -37,7 +37,18 @@ export class Client<
   }
 
   #listener: ProviderListener<SendErrorData, HandlerErrorData> = (e) => {
-    if (e instanceof Error) {
+    if (e instanceof ProviderSendError && e.egressMessage) {
+      const egressMessageId = e.egressMessage.id;
+      const pendingCall = this.pendingCalls[egressMessageId];
+      if (!pendingCall) {
+        console.log({ e });
+        // TODO: pipe error to listeners and message the likely cause,
+        //       a duplicate client.
+        throw new Error();
+      }
+      pendingCall.resolve(e);
+      delete this.pendingCalls[egressMessageId];
+    } else if (e instanceof Error) {
       for (const id in this.pendingCalls) {
         const pendingCall = this.pendingCalls[id]!;
         pendingCall.resolve(e);
@@ -77,12 +88,7 @@ export class Client<
   call: ClientCall<SendErrorData, HandlerErrorData> = (message) => {
     const waiter = deferred<ClientCallEvent<SendErrorData, HandlerErrorData>>();
     this.pendingCalls[message.id] = waiter;
-    try {
-      this.providerRef.send(message);
-    } catch (error) {
-      delete this.pendingCalls[message.id];
-      waiter.resolve(error as ProviderSendError<SendErrorData>);
-    }
+    this.providerRef.send(message);
     return waiter;
   };
 
--- a/rpc/client.ts
+++ b/rpc/client.ts
@@ -37,7 +37,7 @@ export class Client<
   }
 
   #listener: ProviderListener<SendErrorData, HandlerErrorData> = (e) => {
-    if (e instanceof ProviderSendError && e.egressMessage) {
+    if (e instanceof ProviderSendError) {
       const egressMessageId = e.egressMessage.id;
       const pendingCall = this.pendingCalls[egressMessageId];
       if (!pendingCall) {
@@ -103,7 +103,7 @@ export class Client<
       delete this.activeSubscriptionByMessageId[message.id];
       waiter.resolve(activeSubscriptionId);
     };
-    this.pendingSubscriptions[message.id] = listener.bind({
+    const listenerBound = listener.bind({
       message,
       stop,
       state: <T>(ctor: new() => T) => {
@@ -114,10 +114,11 @@ export class Client<
         );
       },
     }) as ClientSubscribeListener<SendErrorData, HandlerErrorData>;
+    this.pendingSubscriptions[message.id] = listenerBound;
     this.call(message)
       .then((maybeError) => {
         if (maybeError instanceof Error) {
-          this.pendingSubscriptions[message.id]!(maybeError);
+          listenerBound(maybeError);
           stop();
         }
       });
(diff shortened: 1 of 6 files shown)


rpc/provider/smoldot.test.ts
```diff
@@ -1,30 +1,66 @@
...
 import { smoldotProvider } from "./smoldot.ts";
 
 const POLKADOT_CHAIN_SPEC_URL =
   "https://raw.githubusercontent.com/paritytech/substrate-connect/main/packages/connect/src/connector/specs/polkadot.json";
 
 Deno.test({
   name: "Smoldot Provider",
-  ignore: true,
+  sanitizeOps: false,
+  sanitizeResources: false,
```
kratico, line 13: Smoldot seems to not close a few timers started in 
```
at start_timer (https://deno.land/x/smoldot@light-js-deno-v0.7.6/instance/bindings-smoldot-light.js:107:17)
```

rpc/client.ts
```diff
@@ -37,7 +37,18 @@ export class Client<
...
-    if (e instanceof Error) {
+    if (e instanceof ProviderSendError && e.egressMessage) {
+      const egressMessageId = e.egressMessage.id;
+      const pendingCall = this.pendingCalls[egressMessageId];
+      if (!pendingCall) {
+        console.log({ e });
+        // TODO: pipe error to listeners and message the likely cause,
+        //       a duplicate client.
+        throw new Error();
+      }
```
kratico, line 48: I think this check can be removed.
It happens when the provider is being discarded and an inflight/pending response is emitted.
In that scenario, I assume that we don't care about the response so we can ignore it.

Thoughts @harrysolovay?

rpc/provider/errors.ts
```diff
@@ -1,6 +1,11 @@
+import { msg } from "../mod.ts";
+
 export class ProviderSendError<Data> extends Error {
   override readonly name = "ProviderSendError";
-  constructor(override readonly cause: Data) {
+  constructor(
+    override readonly cause: Data,
+    readonly egressMessage?: msg.EgressMessage,
```
harrysolovay, line 7: Should this param be optional? In what cases will this be undefined?

rpc/provider/smoldot.ts
```diff
@@ -27,12 +43,21 @@ export const smoldotProvider: Provider<
     nextId,
     send: (message) => {
       (async () => {
-        const conn = await connection(chainSpec, listener);
+        let conn: SmoldotProviderConnection;
+        try {
+          conn = await connection(chainSpec, listener);
+        } catch (error) {
+          listener(new ProviderHandlerError(error as SmoldotHandlerErrorData));
```
harrysolovay, line 50: Awesome solution!

rpc/client.ts
harrysolovay: Which lines specifically?

rpc/client.ts
```diff
@@ -103,7 +114,13 @@ export class Client<
         );
       },
     }) as ClientSubscribeListener<SendErrorData, HandlerErrorData>;
-    this.call(message);
+    this.call(message)
+      .then((maybeError) => {
+        if (maybeError instanceof Error) {
+          this.pendingSubscriptions[message.id]!(maybeError);
```
harrysolovay, line 120: Perhaps alias the `listener.bind...` block on 106 as `listenerBound` before assigning to `this.pendingSubscriptions[message.id]` (`message.id` could also be aliased within this method as `id`)?

rpc/provider/smoldot.ts
```diff
@@ -1,13 +1,29 @@
...
+  | CrashError
+  | JsonRpcDisabledError
+  | MalformedJsonRpcError
+  | QueueFullError;
+type SmoldotHandlerErrorData =
+  | AlreadyDestroyedError
+  | CrashError
+  | JsonRpcDisabledError
+  | AddChainError;
+type SmoldotCloseErrorData = AlreadyDestroyedError | CrashError;
```
harrysolovay, line 26: Awesome

rpc/provider/smoldot.ts
```diff
@@ -27,12 +43,21 @@ export const smoldotProvider: Provider<
...
+        }
         try {
           return conn.inner.sendJsonRpc(JSON.stringify(message));
-        } catch (e) {
-          // TODO: funnel this to relevant handler(s)
-          return new ProviderSendError(e);
+        } catch (error) {
+          if (error instanceof MalformedJsonRpcError || error instanceof QueueFullError) {
+            listener(new ProviderSendError(error, message));
+            return;
```
harrysolovay, line 58: Why return here? If you get a not-all-branches-return err, a more fitting place for the `return` may be after line 61.

harrysolovay (changes requested):

rpc/provider/smoldot.ts
```diff
@@ -67,27 +92,24 @@ async function connection(
   }
   let conn = connections.get(chainSpec);
   if (!conn) {
-    // TODO: try catch this and send through handler within a `ProviderHandlerError`
     const inner = await client.addChain({ chainSpec });
```
wirednkod, line 95: for parachain support, all you need to do is to pass in the potentialRelayChains all the chains inner that you collect during the addChain. Meaning something like:
```suggestion
    const innerMap: Chain[] = [];
    connections.forEach((value: SmoldotProviderConnection) => {
      innerMap.push(value.inner);
    });
    const inner = await client.addChain({ chainSpec, potentialRelayChains: innerMap });
```

rpc/client.ts
kratico:
```
if (!pendingCall) {
        console.log({ e });
        // TODO: pipe error to listeners and message the likely cause,
        //       a duplicate client.
        throw new Error();
      }
```

rpc/provider/smoldot.ts
kratico: thx @wirednkod 
The parachain support will be done in a separate PR

kratico: @harrysolovay we need to decide 2 things

- 1st when connecting to a parachain, I believe that we need to pass the parachain spec and the potential relay chain spec
- 2nd smoldot connection clean up, given that it takes time to sync the light client, connecting/disconnecting to the same chain will make it slow, so perhaps we can keep the synced state without closing the connections

Thouths?

wirednkod: > thx @wirednkod The parachain support will be done in a separate PR

#378

rpc/provider/smoldot.ts
```diff
@@ -67,27 +91,38 @@ async function connection(
...
+    if (typeof chainSpec === "string") {
+      inner = await client.addChain({ chainSpec });
+    } else {
+      const [parachainSpec, relayChainSpec] = chainSpec;
+      const relayChainConnection = await client.addChain({ chainSpec: relayChainSpec });
+      inner = await client.addChain({
+        chainSpec: parachainSpec,
+        potentialRelayChains: [relayChainConnection],
+      });
+    }
```
kratico, line 104: @harrysolovay @wirednkod WDYT?

rpc/client.ts
harrysolovay: Gotcha. Removing sounds good.

rpc/provider/smoldot.ts
```diff
@@ -1,24 +1,43 @@
...
 let client: undefined | Client;
-const connections = new Map<string, SmoldotProviderConnection>();
+const connections = new Map<ChainSpecOptions, SmoldotProviderConnection>();
 class SmoldotProviderConnection
   extends ProviderConnection<Chain, SmoldotSendErrorData, SmoldotHandlerErrorData>
 {}
 
 const nextId = nextIdFactory();
 
+type ChainSpecOptions = string | [parachainSpec: string, relayChainSpec: string];
```
kratico, line 37: reverse and make parachain spec optional

harrysolovay (approved):

rpc/provider/smoldot.ts
wirednkod: Giving my 2 cents here, I think this will strict the user always to start relay chain and parachain at the same spot of the application... IMO these steps should be separated (add relay chain, add parachain) with the same function

kratico: thanks for the input @wirednkod 
For now, the relay/parachain start will be on demand on the first `.send` invocation.

PR #378, opened 2022-11-10
feat: improve smoldot integration (parachain)

wirednkod:
Base branch: #374 

Tests are failing due to an issue with either the spec or the collators of `westmint` parachain.
Browsers only allow web pages to establish secure WebSocket connections, in other words only to nodes that present a TLS certificate (WebRTC will resolve this) and the fact that `westmint` does not have one make the tests failing .

Already opened a case to [devops repo](https://github.com/paritytech/devops/issues/2093)


rpc/provider/smoldot.ts
```diff
@@ -92,7 +92,14 @@ async function connection(
...
   if (!conn) {
-    const inner = await client.addChain({ chainSpec });
+    const innerMap: Chain[] = [];
+    connections.forEach((value: SmoldotProviderConnection) => {
+      innerMap.push(value.inner);
+    });
+    const inner = await client.addChain({
+      chainSpec,
+      potentialRelayChains: innerMap,
+    });
```
kratico, line 102: Thanks for the suggestion @wirednkod 
This `innerMap` will be re-created when there is no existing connection for the given chainSpec, so it seems that it will always be empty.

Anyway, based on your previous tips, I added
https://github.com/paritytech/capi/pull/374/files#diff-312ccecf597532308900880045fdbe77f2bf18934717d547439a320860f3dcc8R94-R104

wirednkod:
Closing this as smoldot integration is done

PR #399, opened 2022-11-14
feat: add smoldot known clients

kratico:
Add Smoldot known clients based on substrate connect chain specs https://github.com/paritytech/substrate-connect/tree/main/packages/connect/src/connector/specs

## How to test?

```
import * as C from "./mod.ts"
const w = await C.knownSmoldotClients.westend.run()
console.log(
  await w.call({
    jsonrpc: "2.0",
    id: w.providerRef.nextId(),
    method: "system_health",
    params: [],
  }),
)

```


effects/rpc_known_smoldot_clients.ts
```diff
@@ -0,0 +1,32 @@
+import * as Z from "../deps/zones.ts"
+import * as rpc from "../rpc/mod.ts"
+import { rpcClient } from "./rpc.ts"
+
+function smoldotClient(relayChainSpecUrl: string, parachainSpecUrl?: string) {
+  return rpcClient(
+    rpc.smoldotProvider,
+    Z.lift([]).next(async () => {
```
kratico, line 8: @harrysolovay used `Z.lift().next()` because `Z.call(async () => ...)` doesn't resolve to the correct type.

Should I create a PR to update `Z.call<R>` to `Z.call<Awaited<R>`

harrysolovay:
@kratico thank you for this chainspec URL list + corresponding clients –– as we've since discussed offline: we'll close this PR, as we'll ultimately want to provide utilities for end developers to enable end users to inject a chainspec (regularly updated by the end user) into a client. While we may––during development––benefit from a list of chainspecs, we'll want to remove most "known" things (proxy WS URLs included).

PR #402, opened 2022-11-14
feat: compute max_weight for multisig example

kratico:
Use `extrinsic.feeEstimation` to compute multisig `max_weight`

### DIFF
--- a/deno.jsonc
+++ b/deno.jsonc
@@ -38,7 +38,6 @@
     "test:update": "deno task test -- -- --update",
     "mdbook:watch": "mdbook watch -o",
     "bench": "deno bench -A --no-check=remote --unstable",
-    "multisig": "deno task run test_util/ctx.ts -- deno task run examples/multisig_transfer.ts",
     "polkagen": "deno task run codegen.ts -d=polkadot -o=target/polkagen --import=../../mod.ts"
   }
 }
--- a/examples/multisig_transfer.ts
+++ b/examples/multisig_transfer.ts
@@ -63,7 +63,6 @@ U.throwIfError(await proposal.run())
 U.throwIfError(await approval.run())
 console.log(U.throwIfError(await daveBalance.run()))
 
-// FIXME: weight calculation (`payment.queryInfo(extrinsic, atBlockHash)`)
 function createOrApproveMultisigProposal<
   Rest extends [
     MaybeTimepoint?: C.Z.$<{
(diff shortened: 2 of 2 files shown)


examples/multisig_transfer.ts
```diff
@@ -93,11 +93,39 @@ function createOrApproveMultisigProposal<
...
     }),
+  })
+    .feeEstimate
+    .access("weight")
+    .next((weight) => {
+      return {
+        ref_time: BigInt(weight.ref_time),
+        proof_size: BigInt(weight.proof_size),
+      }
+    })
```
kratico, line 109: @tjjfvi Would you know if this is codec related?

`ref_time` and `proof_size` are `u64` but they are not decoded as biging

See https://github.com/paritytech/substrate/blob/59da38b3a07b4ddf380ff9b01a5874fd883138a9/primitives/weights/src/weight_v2.rs#L29-L35

examples/multisig_transfer.ts
```diff
@@ -93,11 +93,39 @@ function createOrApproveMultisigProposal<
...
+        value: {
+          type: "transfer_keep_alive",
+          dest: C.compat.multiAddressFromKeypair(T.dave),
+          value: 1_230_000_000_000n,
+        },
+      },
+      other_signatories: signatories.filter((value) => value !== pair.publicKey),
+      store_call: false,
+      max_weight: maxWeight,
+      maybe_timepoint: maybeTimepoint as Rest[0],
```
kratico, line 127: @harrysolovay the extrinsic data is almost the same but the spread operator doesn't work well with `C.Z.rec`

Would you know a better alternative?

I could factor out the `args` within `Z.rec`, WDYT?

kratico: Or, given that it's an example, code duplication might be fine

tjjfvi: I think you want to calculate the weight for the `Balances.transfer_keep_alive` call, not the `Multisig.as_multi`, which would solve this duplication.

harrysolovay: Yea, I believe you'll be able to get the appropriate weight using the encoded inner extrinsic.

On a more general note though...

- A bit of redundancy in examples can definitely help with reader comprehension
- In this particular case, I suppose we could extract the common fields –– another way we'll make this example nicer: @tjjfvi is working on #368, which will enable us to generate/use narrow call data types (and their factory fns)

examples/multisig_transfer.ts
kratico: you are right

It might be a good DX to have the codec convert `number` to `bigint` when it's needed, WDYT?

deno.jsonc
```diff
@@ -38,7 +38,6 @@
     "test:update": "deno task test -- -- --update",
     "mdbook:watch": "mdbook watch -o",
     "bench": "deno bench -A --no-check=remote --unstable",
-    "multisig": "deno task run test_util/ctx.ts -- deno task run examples/multisig_transfer.ts",
```
tjjfvi, line 41: Why was this removed?

examples/multisig_transfer.ts
tjjfvi: Having any implicit coercion legitimizes implicit coercion in other cases, which could quickly get out of hand (c.f. `@polkadot/types`, which is a rat's nest of implicit coercion).

Additionally, all codecs from `scale-ts` currently roundtrip exactly for every valid input (that is,
```ts
!$.is($foo, value) || deepEqual(value, $foo.decode($foo.encode(value)))
```
is always true), and I'd like to preserve this behavior.

tjjfvi: I think it would probably make the most sense for the `.feeEstimate` effect to do this BigInt coercion.

tjjfvi (approved):

examples/multisig_transfer.ts
tjjfvi: Yeah, I'll resolve the duplication in the codegen branch using the call factories once this is merged – no need to address it now.

deno.jsonc
kratico: When I committed the multisig example, I forgot to remove this task

examples/multisig_transfer.ts
kratico: if we do it in the effect, as soon as the `weight` shape changes, the effect will need to be refactored (like it happened with recent polkadot updates).

that's why I believe that implicit coercion from `number` to `bigint` will scale better for this case.

PR #408, opened 2022-11-16
feat: add more RPC client tests

kratico:
Resolves #


rpc/client.test.ts
```diff
@@ -1,8 +1,28 @@
...
 import * as U from "../util/mod.ts"
 import * as known from "./known/mod.ts"
 import * as msg from "./messages.ts"
+import { Provider } from "./mod.ts"
+import { ProviderHandlerError, ProviderSendError } from "./provider/errors.ts"
+
+const createMockClient = () => {
+  let listener: C.rpc.ProviderListener<Error, Error>
+  const providerMockFactory: Provider = (_discoveryValue, clientListener) => {
+    listener = clientListener
```
kratico, line 13: this is very coupled to how the `rpc.Client` calls the provider factory.
I'm open to more elegant suggestions to extract the client listener so we can emit testing messages/errors

rpc/client.ts
```diff
@@ -102,10 +102,15 @@ export class Client<
         )
       },
     }) as ClientSubscribeListener<SendErrorData, HandlerErrorData>
-    this.pendingSubscriptions[message.id] = listenerBound
+    this.pendingSubscriptions[message.id] = (maybeError) => {
+      listenerBound(maybeError)
+      if (maybeError instanceof Error) {
+        stop()
+      }
+    }
```
kratico, line 110: this is required to stop subscriptions on errors

rpc/client.ts
```diff
@@ -102,10 +102,15 @@ export class Client<
...
     this.call(message)
       .then((maybeError) => {
-        if (maybeError instanceof Error) {
+        if (maybeError instanceof Error || maybeError.error) {
```
kratico, line 113: bug fix, if the subscription creation is not successful it should stop the subscription

rpc/client.ts
```diff
@@ -47,7 +47,7 @@ export class Client<
         const pendingCall = this.pendingCalls[id]!
         pendingCall.resolve(e)
         delete this.pendingCalls[id]
-        this.pendingSubscriptions[id]!(e)
+        this.pendingSubscriptions[id]?.(e)
```
kratico, line 50: bug fix, on error there may not be pendingSubscriptions for the given message id

rpc/provider/proxy.ts
```diff
@@ -28,7 +34,13 @@ export const proxyProvider: Provider<string, Event, Event, Event> = (url, listen
       })()
     },
     release: () => {
-      const { cleanUp, listeners, inner } = connection(url, listener)
+      let conn
+      try {
+        conn = connection(url, listener)
```
kratico, line 39: if the connection was not created in `.send`, this call will attempt to create it again.
And if it fails to create the connection it will throw.

rpc/provider/smoldot.test.ts
```diff
@@ -139,5 +143,46 @@ Deno.test({
...
+          )
+        )
+          .text()
+        const stopped = deferred()
+        const provider = smoldotProvider({ relayChainSpec }, (message) => {
+          assertInstanceOf(message, Error)
+          stopped.resolve()
+        })
+        // @ts-ignore make JSON.stringify to throw
+        provider.send(1n)
```
harrysolovay, line 181: Same comment as above

rpc/provider/smoldot.ts
```diff
@@ -66,7 +66,13 @@ export const smoldotProvider: Provider<
       })()
     },
     release: async () => {
-      const { cleanUp, listeners, inner } = await connection(props, listener)
+      let conn
+      try {
+        conn = await connection(props, listener)
+      } catch (_error) {
+        return
```
harrysolovay, line 73: Although an error would likely mean that the connection is dead / "released", this––for some reason––seems not ideal. What're your thoughts on this?

harrysolovay (changes requested):
Added some thoughts –– want to pair on this next week?

rpc/client.ts
tjjfvi: This is a *very* important fix, as these errors were not getting caught, and thus crashed the process, including on the codegen server.

rpc/provider/proxy.test.ts
```diff
@@ -1,25 +1,117 @@
...
+      name: "send non-JSON message",
+      async fn() {
+        const server = createWebSocketServer()
+        const stopped = deferred()
+        const provider = proxyProvider(server.url, (message) => {
+          A.assertInstanceOf(message, Error)
+          stopped.resolve()
+        })
+        // @ts-ignore make JSON.stringify to throw
+        provider.send(1n)
```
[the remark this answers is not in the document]
tjjfvi: I doubt `null!` would (as JSON has no problem with `null`), but `1n as never` would work instead.

rpc/provider/smoldot.ts
kratico: I believe that it would be better to just do
```
conn = connections.get(props)
```

Then, we need to decide what release should do if there is no connection.
One option could be that if there isn't anything to release, then do nothing.
Thoughts?

harrysolovay (approved):

PR #413, opened 2022-11-19
feat: rpc subscription refactor + returning results

harrysolovay:
This PR refactors RPC subscriptions / the Capi-wide listener interface such that one can return (and infer types from) results.

Previously...

```ts
const root = C.chain.unsubscribeNewHeads(T.polkadot)(
  C.chain.subscribeNewHeads(T.polkadot)([], function(header) {
    if (someCondition)
      return this.stop()
  }),
)
```

With these changes...

```ts
const root = C.chain.subscribeNewHeads(T.polkadot)([], (ctx) =>
  (header) => {
    if (someCondition)
      return ctx.stop()
  })
```

The calling of a corresponding unsubscribe call method is now taken care of at the RPC level. This means we no longer need to return the subscription ID from RPC subscription effects, which allows us to return other values.

```ts
const header = C.chain.subscribeNewHeads(T.polkadot)([], (ctx) =>
  (header) => {
    if (header.number > THRESHOLD)
      return ctx.stop(header)
  })
```

`header` can be utilized as any other effect. This enables myriad new possibilities for modeling effects that observe and react to on-chain events.


util/mapping.ts
```diff
@@ -0,0 +1 @@
+export {}
```
tjjfvi, line 1: What's this for?

examples/raw_rpc_client_subscription.ts
```diff
@@ -1,36 +1,27 @@
...
+    }
+  },
 )
 
-console.log(
-  // cspell:disable-next-line
-  `${result ? "S" : "Uns"}uccessfully unsubscribed from subscription ${subscriptionId}`,
-)
+// cspell:disable-next-line
+console.log(`${result ? "S" : "Uns"}uccessfully unsubscribed`)
```
tjjfvi, line 25: Hmm, mightn't `result` be an error, and thus truthy?

tjjfvi (approved):

examples/raw_rpc_client_subscription.ts
harrysolovay: If there is an error, it will call the listener with the error but ultimately, the promise will resolve to undefined... this is definitely worth revisiting. Hoping @kratico will take a look :)

rpc/client.test.ts
```diff
@@ -97,24 +89,26 @@ Deno.test({
...
         emitEvent(new ProviderSendError(null!, message))
-        A.assertEquals(await response, undefined)
+        A.assertEquals(await pending, undefined)
         assertClientCleanup(client)
       },
     })
 
     await t.step({
       name: "subscribe error after subscribing",
+      ignore: true,
```
kratico, line 103: This test has a pending promise because the mock doesn't handle the unsubscription call.

https://github.com/paritytech/capi/blob/db9d0a52100ecb6755627165bd85a555bbe30937/rpc/client.ts#L100-L106

rpc/client.ts
```diff
@@ -73,57 +70,82 @@ export class Client<
...
-    this.call(message)
-      .then((maybeError) => {
+      this.pendingSubscriptions[id] = (event) => {
+        const result = listener(event)
+        if (result instanceof Promise) {
+          result.then(maybeStop)
+        } else {
+          maybeStop(result)
+        }
+      }
```
kratico, line 133: If `event` is an `Error` and `listener` doesn't returns it, the subscription won't stop.

Perhaps `maybeStop` could be updated to 


```
      const maybeStop = (maybeEnd: unknown, event: unknown) => {
        if (maybeEnd instanceof U.End) {
          stop(maybeEnd.value)
        } else if (maybeEnd instanceof Error) {
          stop(maybeEnd)
        } else if (event instanceof Error) {
          stop(event)
        }
      }
```

kratico (dismissed):

examples/raw_rpc_client_subscription.ts
kratico: This is related to https://github.com/paritytech/capi/pull/413/files#r1029743739

if there is an error, and the listener does not return it, then the error will be lost.
we need to decide if the listener should handle the error and stop the subscription, or if the client should check for an error, pass it to the listener and instead of returning the listener result, return the error

WDYT?

kratico (approved):
LGTM

PR #430, opened 2022-11-23
fix: codegen rpc.Client setup

kratico:
Fixes the `C.rpc.Client` setup

### DIFF
--- a/codegen/server/server.ts
+++ b/codegen/server/server.ts
@@ -152,7 +152,7 @@ export const client = new LocalClientEffect(${JSON.stringify(chainUrl.slice(4))}
           `
           : `
 import * as C from ${JSON.stringify(`/@${version}/mod.ts`)}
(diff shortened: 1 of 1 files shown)


codegen/server/server.ts
```diff
@@ -152,7 +152,7 @@ export const client = new LocalClientEffect(${JSON.stringify(chainUrl.slice(4))}
           `
           : `
 import * as C from ${JSON.stringify(`/@${version}/mod.ts`)}
-export const client = C.rpc.rpcClient(C.rpc.proxyProvider, ${JSON.stringify(chainUrl)})
+export const client = new C.rpc.Client(C.rpc.proxyProvider, ${JSON.stringify(chainUrl)})
```
tjjfvi, line 155: I think we still want it to be an effect, but the effect isn't in the `rpc` namespace:
```suggestion
export const client = C.rpcClient(C.rpc.proxyProvider, ${JSON.stringify(chainUrl)})
```

tjjfvi (approved):
Thanks for catching this!

PR #448, opened 2022-11-30
chore: rename `const` to `constant`

harrysolovay:


tjjfvi:
Unless there's a need for this now, I'd be inclined to roll this into the other effect system changes.

harrysolovay:
Sounds reasonable.

PR #449, opened 2022-12-01
feat: use AbortController for client.discard

kratico:
Resolves #


tjjfvi:
Supplanted by #591

PR #454, opened 2022-12-03
feat: rune

### LINKED ISSUE #320: Nicer Signatures For Effects With Voidable Params
Take the following example:

```ts
const root = new C.EntryRead(T.polkadot, "System", "Account", [T.alice.publicKey]);
```

Currently, `root` has a signature of `C.EntryRead<"System", "Account", [Uint8Array], []>`. The final param reflects the param `Rest extends [blockHash?: Z.$<U.HexHash | undefined>]`, which enables users to forgo specifying the block hash entirely. In the case that a block hash is supplied, it populates `0` of the tuple. We of course do not care about the tuple, just the `0` element. A better signature would be `C.EntryRead<"System", "Account", [Uint8Array], undefined>`. This same dilemma presents itself in other effects too.

### LINKED ISSUE #388: fluent APIs (?)
One such example is that of extrinsic event extraction. Something like this would be useful:

```ts
const events = await C.extrinsic({ ... })
  .signed(signFn)
  .sent
  .finalized
  .events;
```

> Taken from #272

### LINKED ISSUE #451: re-thinking the result type
The following boilerplate is starting to feel hefty.

```ts
if (result instanceof Error) {
  // ...
} else {
  // ...
}
```

While typing the result as a union avoids the need for a dedicated `Result` type, it complicates usage of the result. Perhaps the following would be better?

```ts
type Result<T, E extends Error> = {
  ok: T;
  error?: never;
} | {
  ok?: never;
  error: E;
}
```

This would allow for much simpler use of the `ok` value.

```ts
if (result.ok) {
  // ...
}
```

This would also enable the removal of intermediate nesting in cases such as `entryRead`, as we'd no longer need to worry about an `any`-typed `ok` resulting in the loss of error type narrowing.

Thoughts?

tjjfvi:
Resolves #451 
Resolves #388
Resolves #320

Goodbye Zones, hello Rune – an all-new ~~effect~~ *rune* system coming soon to a capi version near you.

This PR is very unstable – avoid reviewing thoroughly lest you suffer from severe WIP-lash.


harrysolovay (approved):

harrysolovay:
Just archived [Zones](https://github.com/paritytech/zones). Goodbye old friend.

PR #457, opened 2022-12-05
docs: update links

NukeManDan:
I see prettier is used in the  .vscode setting elsewhere, use it in md too.

[Metadata.test.ts.snap](https://github.com/paritytech/capi/pull/457/files#diff-e9670ee5a561952b762271c45942a6d95ea4e4ccdc6f0b2b4e1964b3aa683a33) -- Found more links issues, but I think this is a generated file, so actually the runtime has the wrong links embedded in them? :grimacing: 
If so, can put in redirects on docs for these

### DIFF
--- a/docs/Types.md
+++ b/docs/Types.md
@@ -1,6 +1,6 @@
 # Types
 
-The types of the on-chain world are declared in the given chain's Rust source code. While many types may remain consistent across chains, many may differ. On one chain, `AccountData` may be defined with fields describing fungible assets; on another (hypothetical) chain, perhaps `AccountData` references non-fungible assets, reputation, linked accounts or something else entirely. Although FRAME certainly helps to standardize chain properties, those properties can be customized to the extent that we cannot make assumptions regarding shapes of data across chains. Additionally, types can change upon runtime upgrades; your assumptions about the shape of a type may become invalid; to interact with these highly-dynamic on-chain environments––and to do so from a JavaScript environment––poses inherent difficulty. We JS developers must (A) think in terms of Rust data types and (B) keep a lookout for breaking changes to chain runtimes. This document does not provide a silver-bullet solution to this complexity. But it should provide the background necessary for you to address types for your specific use cases.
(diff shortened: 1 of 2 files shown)


docs/Types.md
```diff
@@ -17,21 +17,21 @@ Every FRAME chain exposes metadata about its types and capabilities. This metada
 As always, our first step is to bring Capi into scope.
 
 ```ts
-import * as C from "https://deno.land/x/capi/mod.ts"
+import * as C from "https://deno.land/x/capi/mod.ts";
 ```
 
-Now let's fetch the metadata.
+As an example, let's connect to Polkadot and fetch the chain's metadata for the present highest block.
```
NukeManDan, line 23: Is this true? Of is this somehow statically defined in capi?

NukeManDan:
Perhaps I am failing CI as you instead wanted dprint to format md files?
https://github.com/dprint/dprint-vscode/issues/39 trying now to get this working on my end :crossed_fingers:

harrysolovay:
Correct: we use dprint for MD formatting. Want to remove the change to the vscode settings and then re-run dprint to clear out undesired formatting changes?

docs/Types.md
harrysolovay: I'm not sure exactly what you mean. Can you please re-word this question?

docs/Types.md
```diff
@@ -280,14 +277,12 @@ Let's look at the same example from before: reading some `AccountData`.
...
-const result = await C
-  .entryRead(C.polkadot)("System", "Account", [key])
-  .run()
+const result = await C.entryRead(C.polkadot)("System", "Account", [key]).run()
 ```
 
 In this storage read example, `result` is typed as the successfully-retrieved value (container) unioned with all possible errors.
 
-There are several ways to "unwrap" the inner `value`. The recommended path is to first check for and handle all possible errors, which may encapsulate error specific data (as do SCALE validation errors).
+There are several ways to "unwrap" the inner `value`. The recommended path is to first check for and handle all possible errors, which may encapsulate error specific data (as do [SCALE](https://docs.substrate.io/reference/scale-codec/) validation errors).
```
tjjfvi, line 285: The referenced validation errors are specific to `scale-ts`:

```suggestion
There are several ways to "unwrap" the inner `value`. The recommended path is to first check for and handle all possible errors, which may encapsulate error specific data (as do [SCALE](https://github.com/paritytech/scale-ts) validation errors).
```

tjjfvi:
> Found more links issues, but I think this is a generated file, so actually the runtime has the wrong links embedded in them? 😬

That's generated off of a fixed version of the metadata, which was last refetched about a month ago, so it may very well be fixed in the latest runtime.

tjjfvi (approved):

docs/Types.md
NukeManDan: Am I correct in assuming that the example here using Polkadot metadat is gathered at ruhtime from the network? So this actually is an RPC call that connects to some default node?

tjjfvi: > Am I correct in assuming that the example here using Polkadot metadata is gathered at runtime from the network? So this actually is an RPC call that connects to some default node?

Correct

PR #461, opened 2022-12-12
feat: add ink flipper smart contract example

kratico:
## Description

Add a smart contract deployment and call example


kratico:
covered in https://github.com/paritytech/capi/pull/462

PR #474, opened 2022-12-21
feat: fluent Storage

kratico:
Resolves #


harrysolovay:
Unfortunately, this will require somewhat of a re-write given the changes to the effect system and (soon) changes of #64.

PR #480, opened 2022-12-21
chore: reorganize frame metadata

### LINKED ISSUE #472: line between type info & frame metadata
A small piece of housekeeping: the directory `frame_metadata` should not house `scale_info.ts`, `TyVisitor` and `Contract.ts`.

`Ty` (modeled after that of [`paritytech/scale-info`](https://github.com/paritytech/scale-info)) is not `frame_metadata`-specific. On a related note: building an Ink contract results in generation of bespoke metadata, which includes serde-serialized `Ty`s. That metadata isn't related to FRAME.

harrysolovay:
Fixes #472 

The general aim of this PR is to separate and simplify reflection/metadata-related utilities


reflection/mod.ts
```diff
@@ -0,0 +1,4 @@
+export * from "./Codec.ts"
```
tjjfvi, line 1: I think this should be `scale_info/`

reflection/redirects/mod.ts
```diff
@@ -0,0 +1 @@
+export * from "./Era.ts"
```
tjjfvi, line 1: I think this should be `overrides/`

frame_metadata/_downloaded/mod.ts
```diff
@@ -0,0 +1,29 @@
+// This file was generated by `_tasks/download_frame_metadata.ts`
```
tjjfvi, line 1: No reason for this file to be generated; it's only ever consumed as a full object, so `knownClients` can just be mapped over at runtime.

harrysolovay: We previously used a `setup` fn from `test-common.ts` to avoid reloading metadata between tests, so either way, we have an extra file. Felt fitting to place that file alongside the SCALE files with which it is concerned (and to make sure that––no matter how the download task URL list is defined, the mod file is generated to reflect all downloaded files). Any concerns?

tjjfvi: Fair enough.

kratico (approved):
LGTM 🚀

tjjfvi (approved):

PR #485, opened 2022-12-23
feat: add fluent.Multisig

kratico:
Add `fluent.Multisig` with
- `ratify`, to create or approve proposals for execution
- `vote`, to vote proposals without proposal execution
- `cancel`, to cancel an existing proposal
- `isProposed`, to check if if the given call hash has been proposed
- `proposals`, to list outstanding proposals

It also adds the following utils 
- `u8a.isEqual` to compare `Uint8Array` 
- `callHash` effect to compute the Blake256 hash for the given scale encoded call


effects/extrinsic.ts
```diff
@@ -161,3 +166,22 @@ function $extrinsic<
...
+export function callHash<
+  Client extends Z.$<rpc.Client> = Z.$<rpc.Client>,
+  Call extends Z.$<unknown> = Z.$<unknown>,
+>(client: Client) {
+  return (call: Call) => {
+    return Z.ls($call(client), call).next(([codec, call]) => {
+      return Blake2_256.hash(codec.encode(call))
+    })
+  }
+}
```
kratico, line 187: @harrysolovay I'm not sure about the best location for these `$call` and `callHash`

harrysolovay (approved):

Issue #471, opened 2022-12-20
redefining responsibilities of `serve.ts`


harrysolovay:
Currently, running `serve.ts` starts a server from which one can get request / import on-the-fly codegened TypeScript files (also implements import spec completions prescribed by the Deno LS intellisense API). Capi also houses three approaches to running dev nets:

1. run `test_util/ctx.ts` can be run––this allows multiple processes to share / simultaneously interact with the same dev net(s)
2. spawn the test chain as a subprocess
3. Via zombienet ([this approach](https://github.com/paritytech/capi/blob/f1e447d76dfe1ed24c6f4b486193f5ebcfa9fb48/test_util/zombienet.ts) still needs to be integrated into the other test utils)

**I propose that we blur the line between the codegen and testing servers**. The standard path for development would still be to run the `serve.ts` of the desired Capi version. This is comparable to `react-scripts` in [`create-react-app`](https://github.com/facebook/create-react-app); the user runs a single command to spin up an optimal development environment. In the case of `create-react-app`, one spins up a server that handles hot-module reloading, babel transpilation, etc. In our case, we spin up a server that handles launching test networks, generating chain-specific TypeScript files from a number of discovery value types (both remote and local), import completion and––of course––serving the desired version of Capi (should be the same as the version of the `serve.ts` in use (?)).

Running `serve.ts` locally means that the codegen server can interact with the user's file system, to for instance scan the file tree for chainspecs, zombienet configs, smart contract wasm+metadata, etc. `capi.dev` would still exist, but it would allow only a subset of discovery values and lack test chain management.

If we agree on this direction, a follow up question is: what kinds of discovery values / import specifiers do we want to support?

In the case of chain specs? Maybe...

```ts
"http://localhost:5646/chainspec/../path/to/chainspec.json/mod.ts"
"http://localhost:5646/chainspec/https://raw.githubusercontent.com/paritytech/substrate-connect/main/packages/connect/src/connector/specs/polkadot.json/mod.ts"
```

Do we want to allow reference to scale bytes? This would be more in line with Subxt's codegen.

```ts
"http://localhost:5646/scale_bytes/../path/to/metadata.scale/mod.ts"
```

How about allowing `serve.ts` to launch zombienets?

```ts
"http://localhost:5646/zombienet/../path/to/zombienet.toml#some_parachain/mod.ts"
```

How about an arbitrary bin?

```ts
"http://localhost:5646/bin/../path/to/substrate_chain_bin/mod.ts"
```

[...]

PR #486, opened 2022-12-23
chore: misc server cleanup

harrysolovay:
- Completions in JSON file (we get to use `$schema` for checking, and `serveFile` from std).
- Break apart cache-related code
- Ensure deps defined in `deps`, not inline (for consistency / simplicity moving forward)
- Rename `Cache` to `CacheBase` to avoid potential confusion with built-in Cache API


deps/escad_babel.ts
```diff
@@ -0,0 +1 @@
+export * from "https://escad.dev/deps/babel.ts"
```
tjjfvi, line 1: Just call this `deps/babel.ts` IMO

server/server.ts
```diff
@@ -162,46 +161,14 @@ export const client = C.rpcClient(C.rpc.proxyProvider, ${JSON.stringify(chainUrl
   async chainIndex(chainUrl: string, version: string, chainVersion: string) {
     return await this.cache.get(
       `generated/@${version}/${chainUrl}/@${chainVersion}/_index`,
-      $index,
+      $.array($.str),
```
tjjfvi, line 164: This won't work because the cache memoizes off of the identity of the codec

server/server.ts
```diff
@@ -71,7 +70,7 @@ export abstract class CodegenServer {
   async root(request: Request): Promise<Response> {
     const fullPath = new URL(request.url).pathname
     if (fullPath === "/.well-known/deno-import-intellisense.json") {
-      return this.autocompleteSchema()
+      return await serveFile(request, new URL("completions.json", import.meta.url).pathname)
```
tjjfvi, line 73: This won't work when the server is loaded from a remote URL; should be e.g.
```ts
return await fetch(new URL("completions.json", import.meta.url))
```

deps/escad_babel.ts
tjjfvi: This should also get removed once #471 unblocks #385

tjjfvi (approved):

tjjfvi (approved):

tjjfvi (approved):

PR #487, opened 2022-12-24
chore: codegen + test ctx server refactor

### LINKED ISSUE #471: redefining responsibilities of `serve.ts`

harrysolovay:
Resolves #471


deno.jsonc
```diff
@@ -38,7 +38,6 @@
     "dnt": "deno run -A _tasks/dnt.ts",
     "test": "deno task ctx -- deno test -A -r -L=info --ignore=target --parallel",
     "test:update": "deno task test -- --update",
-    "moderate": "deno run -A https://deno.land/x/moderate@0.0.5/mod.ts && dprint fmt",
```
tjjfvi: ?

harrysolovay: Didn't we decide to get rid of the `mod.ts` in examples?

tjjfvi: Yes, but we use `moderate` elsewhere

.github/workflows/test.yml
```diff
@@ -13,10 +13,32 @@ jobs:
...
+        uses: ./.github/actions/setup-binary
+        with:
+          binary-name: polkadot
+          binary-version: v0.9.36
+          binary-github: https://github.com/paritytech/polkadot
+      - name: Setup Cumulus
+        uses: ./.github/actions/setup-binary
+        with:
+          binary-name: polkadot-parachain
+          binary-version: v0.9.33
```
ryanleecode, line 33:
```suggestion
          binary-version: v0.9.330
```

rpc/client.test.ts
```diff
@@ -10,12 +10,13 @@ import {
...
   ProviderListener,
   ProviderSendError,
+  proxyProvider,
 } from "./mod.ts"
 
 Deno.test({
   name: "RPC Client",
   async fn(t) {
-    const client = await T.polkadot.client
+    const client = new Client(proxyProvider, discoveryValue)
```
tjjfvi, line 19: Can't this import the raw client?

server/Env.ts
```diff
@@ -0,0 +1,24 @@
...
+            `${generatorId} Provider with \`providerId\` of \`${providerId}\` already initialized`,
+          )
+        }
+        generatorProviders[providerId] = provider
+      } else {
+        this.providers[generatorId] = { [providerId]: provider }
+      }
+      return provider
+    })
+  }
```
tjjfvi, line 23:
```suggestion
  constructor(readonly signal: AbortSignal, providerFactories: ProviderFactory[]) {
    for (const factory of providerFactories) {
      const provider = factory(this)
      const { generatorId, providerId } = provider
      const generatorProviders = this.providers[generatorId] ??= {}
      if (generatorProviders[providerId]) {
        throw new Error(
          `${generatorId} Provider with \`providerId\` of \`${providerId}\` already initialized`,
        )
      }
      generatorProviders[providerId] = provider
    }
  }
```

server/PathInfo.test.ts
```diff
@@ -0,0 +1,81 @@
...
+  assertExists(r)
+  assertObjectMatch(r, {
+    vCapi,
+    generatorId,
+    providerId,
+    target,
+  })
+})
+
+// @t6... you were right about naming constants... let's break convention where fitting.
```
tjjfvi, line 75: 👍 
```suggestion
```

server/PathInfo.ts
```diff
@@ -0,0 +1,48 @@
...
+    }
+  }
+  return isPathInfo(tmp) ? tmp : undefined
+}
+
+export function isPathInfo(inQuestion: object): inQuestion is PathInfo {
+  return !!("generatorId" in inQuestion && typeof inQuestion.generatorId === "string"
+    && "providerId" in inQuestion && typeof inQuestion.providerId === "string"
+    && "target" in inQuestion && typeof inQuestion.target === "string")
+}
```
tjjfvi, line 48:
```suggestion
const rPathInfo =
  /^(@(?<vCapi>.+?)\/)?(?<generatorId>.+?)\/(?<providerId>.+?)\/(?<target>.+?)(\/@(?<vRuntime>.+?)(\/(?<filePath>.+))?)?$/

export function parsePathInfo(src: string): PathInfo | undefined {
  return rPathInfo.exec(src)?.groups as PathInfo | undefined
}
```

server/local/factories.ts
```diff
@@ -0,0 +1,57 @@
...
+  if (path.startsWith("file://")) {
+    return await fetch(path)
+  }
+  return new Response(null, {
+    status: Status.Found,
+    headers: { Location: path },
+  })
+}
+
+export function fourOFour(req: Request): U.PromiseOr<Response> {
```
tjjfvi, line 35:
```suggestion
export function notFound(req: Request): U.PromiseOr<Response> {
```

server/local/factories.ts
```diff
@@ -0,0 +1,57 @@
...
+  if (acceptsHtml(req)) {
+    return page(FourOFourPage({}))
+  }
+  return new Response("404", { status: Status.NotFound })
+}
+
+export function fiveHundred(req: Request, message?: string): U.PromiseOr<Response> {
```
tjjfvi, line 42:
```suggestion
export function serverError(req: Request, message?: string): U.PromiseOr<Response> {
```

server/local/mod.ts
```diff
@@ -0,0 +1,54 @@
...
+    if (path === "") return f.page(LandingPage())
+    const pathInfo = parsePathInfo(path)
+    if (pathInfo) {
+      const { vCapi, vRuntime, providerId, generatorId, filePath } = pathInfo
+      const generatorProviders = env.providers[generatorId]
+      if (generatorProviders) {
+        const provider = generatorProviders[providerId]
+        if (provider) {
+          if (vCapi) {
+            return await f.fiveHundred(
```
tjjfvi, line 19:
```suggestion
            return f.fiveHundred(
```
(and similar along the rest)

deno.jsonc
```diff
@@ -1,6 +1,10 @@
 {
   "compilerOptions": {
     "checkJs": false,
+    "jsx": "react",
+    "jsxFactory": "h",
+    "jsxFragmentFactory": "Fragment",
+    "lib": ["deno.ns", "dom"],
```
tjjfvi, line 7: I don't think we should introduce a dependency on preact. The use is so minimal it's not worthwhile. Once we do get to making a more fancy server, it'll likely be out of the capi repo anyway.

server/local/factories.ts
```diff
@@ -0,0 +1,57 @@
...
+  if (acceptsHtml(req)) {
+    return page(FiveHundredPage({ message }))
+  }
+  return new Response("500", { status: Status.InternalServerError })
```
tjjfvi, line 46:
```suggestion
  return new Response(message || "500", { status: Status.InternalServerError })
```

server/local/factories.ts
```diff
@@ -0,0 +1,57 @@
...
+    status: Status.Found,
+    headers: { Location: path },
+  })
+}
+
+export function fourOFour(req: Request): U.PromiseOr<Response> {
+  if (acceptsHtml(req)) {
+    return page(FourOFourPage({}))
+  }
+  return new Response("404", { status: Status.NotFound })
```
tjjfvi, line 39: Even if it accepts HTML it should still get the status code. And the HTML isn't really adding anything.

```suggestion
  return new Response("404", { status: Status.NotFound })
```

server/local/pages/Code.tsx
```diff
@@ -0,0 +1,74 @@
...
+  text-decoration-color: currentColor;
+}
+`
+  return (
+    <>
+      <style>{STYLE}</style>
+      <body>
+        <h3>
+          <code>
+            <a style="color:inherit" href="${escapeHtml(path)}">${path}</a>
```
tjjfvi, line 67: This isn't right. And the JSX isn't really adding anything over just a template string.

server/local/pages/Code.tsx
```diff
@@ -0,0 +1,74 @@
...
+          </code>
+        </h3>
+        <pre class="shiki"><code dangerouslySetInnerHTML={{__html: codeContent}} /></pre>
```
tjjfvi, line 70: Here, the JSX is actually making it less clean.

server/local/pages/Landing.tsx
```diff
@@ -0,0 +1,5 @@
+import { h } from "../../../deps/preact.ts"
+
+export function LandingPage() {
+  return <div>Welcome to the Capi Console</div>
```
tjjfvi, line 4: If there's no styling, I'd rather see a plain text response than HTML.

server/mod.ts
```diff
@@ -1,3 +1,3 @@
-export * from "./cache/mod.ts"
-export * from "./local.ts"
-export * from "./server.ts"
+export * from "./Env.ts"
```
tjjfvi, line 1: I couldn't comment on the favicon file, so: thoughts on the favicon being changed to the logo I made?

util/string.ts
```diff
@@ -0,0 +1,9 @@
+export type Splitter = (char: string, src: string) => undefined | [string, string]
+
+export const [splitFirst, splitLast] = (["indexOf", "lastIndexOf"] as const).map((m) => {
+  return (char, src) => {
+    const i = src[m](char)
+    if (i === -1) return undefined
+    return [src.slice(0, i), src.slice(i + 1)]
+  }
+}) as [Splitter, Splitter]
```
tjjfvi, line 9: Here's a case where I think it's better WET than DRY. But it can be tightened some with a ternary.
```suggestion
export function splitFirst(char: string, src: string) {
  const i = src.indexOf(char)
  return i === -1 ? undefined : [src.slice(0, i), src.slice(i, 1)]
}

export function splitLast(char: string, src: string) {
  const i = src.lastIndexOf(char)
  return i === -1 ? undefined : [src.slice(0, i), src.slice(i, 1)]
}
```

words.txt
```diff
@@ -210,3 +210,12 @@ lparachain
 unioned
 rustc
 procps
+twind
```
tjjfvi, line 213: This file should be sorted, like cspell's quick fix does automatically.

rpc/client.test.ts
harrysolovay: It could, but this test is meant to cover client instantiation, discarding, etc. –– hence, best to do it manually with the discovery value

server/local/pages/Code.tsx
harrysolovay: Agreed, but it's a start

server/local/pages/Landing.tsx
harrysolovay: There will be––just a start

server/mod.ts
harrysolovay: YES –– wanna swap it out with yours?

server/PathInfo.ts
harrysolovay: Good point: better for maintainability

providers/frame/Zombienet.ts
```diff
@@ -42,9 +42,10 @@ export class ZombienetProvider extends FrameProxyProvider {
       const network = await this.zombienet(configPath)
       const node = network.nodesByName[nodeName]
       if (!node) {
+        console.log({ pathInfo, configPath, nodeName, target, targetParts })
```
tjjfvi, line 45: Was this intentionally left here?

deno.jsonc
```diff
@@ -1,9 +1,6 @@
 {
   "compilerOptions": {
     "checkJs": false,
-    "jsx": "react",
-    "jsxFactory": "h",
-    "jsxFragmentFactory": "Fragment",
     "lib": ["deno.ns", "dom"],
```
tjjfvi, line 4: This dom shouldn't be necessary

deps/preact.ts
```diff
@@ -0,0 +1 @@
+export * from "https://esm.sh/preact@10.5.15"
```
tjjfvi, line 1: This should be deleted as well

server/local/mod.ts
```diff
@@ -1,13 +1,12 @@
...
 import { Env, parsePathInfo } from "../mod.ts"
 import * as f from "./factories.ts"
-import { LandingPage } from "./pages/mod.ts"
 
 export function handler(env: Env): Handler {
   return async (req) => {
     const url = new URL(req.url)
     const path = url.pathname.slice(1)
-    if (path === "") return f.page(LandingPage())
+    if (path === "") return f.notFound()
```
tjjfvi, line 9:
```suggestion
    if (path === "") return new Response("capi dev server active")
```

harrysolovay:
Disabled the `test:examples` step in the test workflow. See #532.

harrysolovay:
When removing the `deno.jsonc`'s `lib` field entirely:
 
without_either

Strange given that we didn't used specify `lib`

When getting rid of just "dom":

Screen Shot 2023-01-28 at 3 10 14 PM

tjjfvi (approved):

ryanleecode:
🥳

PR #492, opened 2023-01-03
feat: add connection retry to proxyProvider

kratico:
Add proxy connection retry.
The retry logic uses https://deno.land/std@0.170.0/async/mod.ts?s=retry

How it works?
- apply the retry logic to create an opened `WebSocket`
- if the `WebSocket` gets closed, attached listeners will be sent an error
- the next attempt to get an opened `WebSocket` will apply the retry logic again

There are 2 topics about retries
- connection retries
- call/subscription retry

This PR attempts to solve the 1st bullet by improving the provider.
To solve the call/subscription retry, some knowledge about the inflight message intents (and ids) is needed.
That means that call/subscription retry need to be solved at the client level or above.

Next step
- implement connection retry for `smoldot`

### DIFF
--- a/examples/derived.ts
+++ b/examples/derived.ts
@@ -1,9 +1,18 @@
 import * as C from "http://localhost:5646/@local/mod.ts"
 import * as U from "http://localhost:5646/@local/util/mod.ts"
 
-const ids = C.entryRead(C.polkadot)("Paras", "Parachains", [])
-  .access("value")
-  .as<number[]>()
+// TODO: uncomment these lines to run a single effect upon solving `count` in zones
+// or effects/rpc.ts#discardCheck
+// const ids = C.entryRead(C.polkadot)("Paras", "Parachains", [])
+//   .access("value")
+//   .as<number[]>()
+
+const ids = U.throwIfError(
+  await C.entryRead(C.polkadot)("Paras", "Parachains", [])
+    .access("value")
+    .as<number[]>()
+    .run(),
+)
 
 const root = C.Z.each(ids, (id) => {
   return C.entryRead(C.polkadot)("Paras", "Heads", [id])
--- a/rpc/provider/base.ts
+++ b/rpc/provider/base.ts
@@ -86,11 +86,11 @@ export class ListenersContainer<
   >()
 
   set(discoveryValue: DiscoveryValue, listener: ProviderListener<SendErrorData, HandlerErrorData>) {
-    let map = this.#listeners.get(discoveryValue)
-    if (!map) {
-      map = new Map()
-      this.#listeners.set(discoveryValue, map)
-    }
+    const map = U.getOrInit(this.#listeners, discoveryValue, () =>
+      new Map<
+        ProviderListener<SendErrorData, HandlerErrorData>,
+        ProviderListener<SendErrorData, HandlerErrorData>
+      >())
     if (map.has(listener)) return
     map.set(
       listener,
--- a/rpc/provider/base.ts
+++ b/rpc/provider/base.ts
@@ -71,3 +71,54 @@ export function nextIdFactory() {
   let i = 0
   return () => i++
 }
+
+export class ListenersContainer<
+  DiscoveryValue,
+  SendErrorData,
+  HandlerErrorData,
+> {
+  #listeners = new Map<
+    DiscoveryValue,
+    Map<
+      ProviderListener<SendErrorData, HandlerErrorData>,
+      ProviderListener<SendErrorData, HandlerErrorData>
+    >
+  >()
+
+  set(discoveryValue: DiscoveryValue, listener: ProviderListener<SendErrorData, HandlerErrorData>) {
+    let map = this.#listeners.get(discoveryValue)
+    if (!map) {
+      map = new Map()
+      this.#listeners.set(discoveryValue, map)
+    }
+    if (map.has(listener)) return
+    map.set(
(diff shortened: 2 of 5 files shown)


examples/derived.ts
```diff
@@ -1,9 +1,18 @@
...
+// const ids = C.entryRead(C.polkadot)("Paras", "Parachains", [])
+//   .access("value")
+//   .as<number[]>()
+
+const ids = U.throwIfError(
+  await C.entryRead(C.polkadot)("Paras", "Parachains", [])
+    .access("value")
+    .as<number[]>()
+    .run(),
+)
```
kratico, line 15: I re-wrote this because there is an issue with the client reference counting.
After
```
 const ids = C.entryRead(C.polkadot)("Paras", "Parachains", [])
   .access("value")
   .as<number[]>()
```
The client reference count reaches `1` and it's discarded in https://github.com/paritytech/capi/blob/7edf6a2c26fd55b49f68108cd989f1f73134061d/effects/rpc.ts#L93-L102

The above invokes the `proxyProvider.release` that removes the `WebSocket` listener.
As a result the next call `C.Z.each(...)` successfully creates a new `WebSocket` but there is not listener for the incoming messages.

rpc/provider/proxy.ts
```diff
@@ -1,95 +1,107 @@
...
 
-/** Global lookup of existing connections */
-const connections = new Map<string, ProxyProviderConnection>()
-type ProxyProviderConnection = ProviderConnection<WebSocket, Event, Event>
-
 const nextId = nextIdFactory()
 
+const listenersContainer = new ListenersContainer<string, Event, Event>()
+const activeWs = new Map<string, WebSocket>()
+const connectingWs = new Map<string, Promise<WebSocket>>()
```
kratico, line 10: These 2 maps are used for
- caching an already opened `WebSocket`
- deduplicating `WebSocket` creation while there is one that is connecting (not OPEN)

rpc/provider/proxy.ts
```diff
@@ -1,95 +1,107 @@
...
+const CUSTOM_WS_CLOSE_CODE = 4000
```
kratico, line 11: Status codes in the range 4000-4999 are reserved for private use (see https://www.rfc-editor.org/rfc/rfc6455.html#section-7.4.1)

This code is used to signal a graceful client `WebSocket` close.

rpc/provider/base.ts
```diff
@@ -71,3 +71,54 @@ export function nextIdFactory() {
   let i = 0
   return () => i++
 }
+
+export class ListenersContainer<
```
kratico, line 75: The intent of this class is to replace the `ProviderConnection` class and use it in the `smoldot` provider too.
The `ProviderConnection` class has the `inner` and `cleanUp` props, but they are not used by the class methods, so it seems that they don't belong to this class.
After removing these props from `ProviderConnection`, this class has the responsibility of containing listeners for a single `WebSocket` connection.
The `ListenersContainer` is an improvement on the above that can contain listeners for many `WebSocket` connections.

rpc/provider/proxy.ts
```diff
@@ -1,95 +1,107 @@
...
+
 export const proxyProvider: Provider<string, Event, Event, Event> = (url, listener) => {
```
kratico, line 13: I believe that we need to pass [retryOptions](https://deno.land/std@0.170.0/async/mod.ts?s=RetryOptions) here.
Perhaps, creating with a `proxyProviderFactory`
```
const proxyProviderFactory = (retryOptions) => (url, listener) => {...};
const proxyProvider = proxyProviderFactory(defaultRetryOptions);
```

Thoughts?

harrysolovay (changes requested):
With these changes, provider `send` calls are more likely to successfully go through, bc unexpected provider open failures are caught and their opens are retried. **This is great**. However, is this the right first step towards a comprehensive recovery story? Does it position us to address other parts of the recovery story?

What else might trigger a retry? Can that retry be triggered by the client handler?

As a thought experiment, what refactoring would be required to **not** boil in the retry logic to the provider itself.

```ts
const p = retryProvider(baseProvider, retryOptions)
const c = new Client(p, discoveryValue)
```

> Note: it is unclear to me whether retry logic should or should not be the default... but this is an important consideration given our thinking that we'll gradually add on recovery-related safeguards in the `handler` of `Client`. With this in mind, should things such as `openedWs` be accessible from `Client`? What other provider context might we want to access from elsewhere? And when?

rpc/provider/base.ts
```diff
@@ -71,3 +71,54 @@ export function nextIdFactory() {
...
+  DiscoveryValue,
+  SendErrorData,
+  HandlerErrorData,
+> {
+  #listeners = new Map<
+    DiscoveryValue,
```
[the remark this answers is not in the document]
kratico: `DiscoveryValue` is not a good name.
This is just a key type for `Map` so it doesn't always need to be the provider `discoveryValue`

For smoldot, it's a bit challenging to pick a key for parachains.
To connect to a parachain, we need the potential relay chain spec.
So, a better key could be the concatenation of the relay+parachain specs.
Currently the key is an object which can easily have a different object ref for the same values
```
{
  chainSpec: {
    relay: string
    para?: string
  }
}
```

harrysolovay:
For context: this is on hold until @kratico returns from the academy :)

PR #498, opened 2023-01-06
chore: update scale dependency to 0.10.0

### LINKED ISSUE #497: upgrade scale to v0.10.0
Breaking changes were introduced in paritytech/scale-ts#128; most notably:
```diff
 const foo = $.object(
-  ["bar", $.u8],
-  ["baz", $.str],
+  $.field("bar", $.u8),
+  $.field("baz", $.str),
 )
```

For now, for simplicity, I think we should use `$.field("foo", $.option($bar))` instead of `$.optionalField("foo", $bar)` in the codegen. But in the non-generated code, I see no reason not to use `$.optionalField`.

ryanleecode:
Resolves #497.

One minor downside to this change is that the types can no longer "catch" the optional fields. i.e. if the field is optional and we don't put it in the codec then it wont be caught by the compiler.

### DIFF
--- a/codegen/genCodecs.ts
+++ b/codegen/genCodecs.ts
@@ -27,24 +27,25 @@ import type * as types from "./types/mod.ts"
       return addCodecDecl(
         ty,
         `$.object(${
-          ty.fields.map((x) =>
-            S.array([
-              S.string(normalizeCase(x.name!)),
-              this.visit(x.ty),
-            ])
-          ).join(", ")
+          ty.fields.map((x) => `$.field(${S.string(normalizeCase(x.name!))}, ${this.visit(x.ty)})`)
+            .join(", ")
         })`,
       )
     },
     option(ty, some) {
       return addCodecDecl(ty, `$.option(${this.visit(some)})`)
     },
+
+    //  $.instance(ChainError, $.tuple(this.visit(err)), (x) => x.value),
+
     result(ty, ok, err) {
       return addCodecDecl(
         ty,
         `$.result(${
           this.visit(ok)
-        }, $.instance(C.ChainError<$.Native<typeof $${err.id}>>, ["value", ${this.visit(err)}]))`,
+        }, $.instance(C.ChainError<$.Native<typeof $${err.id}>>, $.tuple(${
+          this.visit(err)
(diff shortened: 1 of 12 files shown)


codegen/genCodecs.ts
```diff
@@ -27,24 +28,23 @@ import type * as types from "./types/mod.ts"
...
-          ).join(", ")
+          ty.fields.map((x) => `$.field(${S.string(normalizeCase(x.name!))}, ${this.visit(x.ty)})`)
+            .join(", ")
         })`,
       )
     },
     option(ty, some) {
       return addCodecDecl(ty, `$.option(${this.visit(some)})`)
     },
+
```
tjjfvi, line 39:
```suggestion
```

scale_info/Codec.ts
```diff
@@ -26,14 +26,17 @@ export function DeriveCodec(tys: Ty[]): DeriveCodec {
...
       )
     },
     option(_ty, some) {
       return $.option(this.visit(some))
     },
     result(_ty, ok, err) {
-      return $.result(this.visit(ok), $.instance(ChainError, ["value", this.visit(err)]))
+      return $.result(
+        this.visit(ok),
+        $.instance(ChainError, $.tuple(this.visit(err)), (x) => x.value),
```
tjjfvi, line 38:
```suggestion
        $.instance(ChainError, $.tuple(this.visit(err)), (x) => [x.value]),
```

codegen/genCodecs.ts
```diff
@@ -27,24 +27,23 @@ import type * as types from "./types/mod.ts"
...
     result(ty, ok, err) {
       return addCodecDecl(
         ty,
         `$.result(${
           this.visit(ok)
-        }, $.instance(C.ChainError<$.Native<typeof $${err.id}>>, ["value", ${this.visit(err)}]))`,
+        }, $.instance(C.ChainError<$.Native<typeof $${err.id}>>, $.tuple(${
+          this.visit(err)
+        }), (x) => x.value))`,
```
tjjfvi, line 46:
```suggestion
        }), (x) => [x.value]))`,
```

tjjfvi (approved):

PR #505, opened 2023-01-19
feat: ci test for browser

### LINKED ISSUE #328: ensure examples run examples in Deno+puppeteer in CI
We need a minimal way to test that simple examples run in the browser. This should take the form of a check that appropriately block merging upon fails.

Related

- #437

tamrai:
Resolves #328 and #385 .
- [x] - `deno task test:browser examples/derived.ts  # one example`
- [x] - `deno task test:browser examples/derived.ts examples/read_events.ts  # two examples, parallel`
- [x] - `deno task test:browser  # all examples, parallel`


rpc/known/mod.ts
```diff
@@ -7,7 +7,8 @@ export * from "./contracts.ts"
 export * from "./framesystem.ts"
 export * from "./grandpa.ts"
 export * from "./mmr.ts"
-export * from "./mod.ts"
+// BUG: https://github.com/paritytech/capi/issues/328#issuecomment-1382020099
+// export * from "./mod.ts"
```
tjjfvi, line 11: No need to keep this commented out; it can just be deleted.

test_browser/expected_results.ts
```diff
@@ -0,0 +1,8 @@
+type IExpectedResults = Record<string, string>
+
+const derived =
+  "[1000,1001,2000,2002,2004,2006,2007,2008,2011,2012,2013,2019,2021,2026,2027,2030,2031,2032,2034,2035,2037,2039,2040,2043,2046,2048,2051,2052,2053,2055,2056,2058,2086,2090,2091,2092]"
+
+const derived2 = "{\"value\":\"Derived 2 test results\"}"
```
tjjfvi, line 6: Though I theoretically like being able to communicate expected results back to the server, none of the examples currently have expected results, other than "shouldn't throw". If we were to flesh out the examples to be better tests, I think we'd likely want to inline assertions into the files, rather than putting expected results in another file – in part because there is often not just *one* expected result, as the output often includes non-deterministic things like block hashes.

I think all we need is to communicate pass/fail to the server, and ideally to intercept console logs to send to the server.

test_browser/puppeteer/puppeteer_deno.ts
```diff
@@ -0,0 +1,18 @@
+// BUG:
+// PUPPETEER_PRODUCT=chrome deno run -A --unstable puppeteer_deno.ts
+// Uncaught PermissionDenied: Permission denied (os error 13)
+// https://github.com/lucacasonato/deno-puppeteer/issues/72
+
+import puppeteer from "https://deno.land/x/puppeteer@16.2.0/mod.ts"
+// import puppeteer from "npm:puppeteer"
```
tjjfvi, line 7: I see there's a bug with the deno.land version of puppeteer; does the `npm:` specifier work?

It's definitely a good first step to run it at all, but it would be ideal if it could work without needing node.

tjjfvi (commented):
I know it's draft, but I couldn't help myself :)

test_browser/expected_results.ts
harrysolovay: This is a great point––I also like the idea of us refactoring our examples to include assertions (in a different PR). No need for the browser CI tests to check actual/expected––just that there were no throws.

.gitignore
```diff
@@ -2,3 +2,4 @@
 .env
 benches.json
 target
+node_modules
```
harrysolovay, line 5: Can we use a subdir as the root? `target` even?

test_browser/test_browser.ts
```diff
@@ -0,0 +1,34 @@
...
+import { PORT, STATIC_DIR } from "./constants.ts"
+import EXPECTED_RESULTS from "./expected_results.ts"
+
+const handler = async (req: Request): Promise<Response> => {
+  const isPOST = req.method === "POST"
+  const pattern = new URLPattern({ pathname: "/:file_path" })
+  const filePath = pattern.exec(req.url)?.pathname.groups.file_path
+  const isFinished = filePath === "end_of_tests"
+
+  if (isFinished) Deno.exit(0)
```
harrysolovay, line 14: Not the ideal way to terminate the process. You can abort controller and supply its signal to `serve`'s props. Then call `abort` when necessary.

harrysolovay:
Another thing: make sure to remove [test_browser/puppeteer/package-lock.json](https://github.com/paritytech/capi/pull/505/files#diff-82fc09013fcf087503ba2f075a3bf2b5fc760f4fb67d8616b61af1db0f036b9c)

tamrai:
#### What API are we aiming to for now?

Ideas.

- `deno task run test:browser test_browser/examples/derived.ts` - single test
- `deno task run test:browser test_browser/examples/derived.ts test_browser/examples/derived2.ts` - multiple tests
- `deno task run test:browser` - all tests inside `test_browser/examples/*.ts`
- or just required tests hardcoded in `test_browser/constants.ts`

#### How the source example test file should look like?

Ideas.

- Input file: `test_browser/examples/derived.ts`

```js
import * as C from "capi.js"

const root = C.entryRead(C.polkadot)("Paras", "Parachains", []).access("value")

console.log(C.throwIfError(await root.run()))
```

- Output file: `test_browser/static/derived.html`

```js
<script type="module">
  import * as C from "./capi.js"

  ;(async function browserTest() {
    const root = C.entryRead(C.polkadot)("Paras", "Parachains", []).access("value")

    console.log(C.throwIfError(await root.run()))

    console.info('derived')
  })()
</script>
```

// @harrysolovay @tjjfvi

test_browser/puppeteer/puppeteer_deno.ts
harrysolovay: Can you please share the full error?

.gitignore
harrysolovay: How would this help us avoid a `node_modules` bloating the root dir?

test_browser/test_browser.ts
harrysolovay: Great! In the future, no need to add comments like `Thanks, applied!`. Just mark the comment as resolved.

.gitignore
tjjfvi: The `node_modules` in already nested inside `test_browser`; only the `.gitignore` for it is located in the root.

tjjfvi:
We want to run the existing examples in the browser, so there's no need for a `test_browser/examples` directory.

Ideally it should look like:
```
deno task run test:browser examples/transfer.ts # one example
deno task run test:browser examples/transfer.ts examples/derived.ts # two examples, parallel
deno task run test:browser # all examples, parallel
```

Each example file can be `deno bundle`d independently; no need to bundle capi and then tweak the imports.

Also, could you nest `test_browser` in `_tasks`?

tjjfvi:
Also, there's no need for the `browserTest` async IIFE, as all modern browsers support top-level await in modules.

tjjfvi:
And the output HTML files should reference the bundled file with
```
<script src="..." type="module"></script>
```
rather than inlining it, to avoid issues with HTML encoding

test_browser/puppeteer/puppeteer_deno.ts
tamrai: fixed -> https://github.com/lucacasonato/deno-puppeteer/issues/51

_tasks/test_browser/helpers.ts
```diff
@@ -0,0 +1,17 @@
+export const pathToJS = (path: string) => path.replace(".ts", ".js")
```
harrysolovay, line 1: Always treat common abbreviations as words. Aka., the casing of `JS` in the context of `pathToJS` should be `pathToJs`. Same thing with `pathToHTML`.

_tasks/test_browser/server.ts
```diff
@@ -0,0 +1,15 @@
+import { serveDir } from "https://deno.land/std@0.172.0/http/file_server.ts"
+import { serve } from "https://deno.land/std@0.172.0/http/server.ts"
+
+import { PORT, STATIC_DIR } from "./constants.ts"
+
+const ac = new AbortController()
+
+const handler = async (req: Request): Promise<Response> =>
```
harrysolovay, line 8: Not sure why you're splitting apart so many blocks. You can inline the handler in `serve`, given that it's not utilized elsewhere. Same thing with the `() => ac.abort()`.  Frankly, same thing with the constants file and others. There's not a ton of code here. Let's git it all in a single `_tasks/test_browser.ts`.

deno.jsonc
```diff
@@ -29,14 +29,14 @@
...
     "run": "deno run -A",
-    "run:browser": "deno task run test_util/ctx.ts -- deno task run _tasks/run_browser.ts",
     "debug": "deno task run --inspect-brk",
     "udd": "deno task star && deno task run https://deno.land/x/udd@0.5.0/main.ts target/star.ts",
     "dnt": "deno task run _tasks/dnt.ts",
     "star": "deno task run _tasks/star.ts && deno cache --check target/star.ts",
     "codegen": "deno task run cache.ts examples/mod.ts",
     "test": "deno task run test_util/ctx.ts -- deno test -A -L=info --ignore=target --parallel",
     "test:update": "deno task test -- -- --update",
+    "test:browser": "deno task run --unstable _tasks/test_browser/main.ts",
```
harrysolovay, line 39: Why is the `--unstable` flag necessary?

_tasks/test_browser/server.ts
```diff
@@ -0,0 +1,15 @@
...
+  await serveDir(req, { fsRoot: STATIC_DIR })
+
+export const startServer = () => {
+  serve(handler, { port: PORT, signal: ac.signal })
+}
+
+export const stopServer = () => ac.abort()
```
harrysolovay, line 15: One cannot reuse an abort controller instance, so exposing this––in this way––is moot. Like I say above, let's avoid unnecessary aliasing. If it's no reused, air on the side of less code.

harrysolovay (changes requested):
There's a lot of preemptive separation/utils/aliasing. Let's get this all into a single script in `_tasks/test_browser.ts`

_tasks/test_browser/helpers.ts
```diff
@@ -0,0 +1,17 @@
...
+export const pathToHTML = (path: string) => path.replace(".ts", ".html")
+export const pathToName = (path: string) => path.split("/").pop()?.split(".")[0]
```
tjjfvi, line 3:
```suggestion
export const pathToName = (path: string) => path.split("/").at(-1)?.split(".")[0]
```

deno.jsonc
tamrai: Seems like it works without an `--unstable` . Removed.
- https://github.com/lucacasonato/deno-puppeteer#usage

- https://alexandrempsantos.com/deno/puppeteer-with-deno/
> Due to deno-puppeteer using a few Deno APIs that are still under active development (and subject to changes), we will need to run our program with the --unstable flag.

tamrai:
#### How do we want to run multiple tests in parallel?
- run multiple deno processes
- run multiple async JS functions

tjjfvi:
Since these tests just spawn another process, no need for multiple deno processes; there will be multiple puppeteer processes either way. Parallel promises should suffice.

.github/workflows/test_browser.yml
```diff
@@ -0,0 +1,21 @@
...
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 16
+      - uses: denoland/setup-deno@9db7f66e8e16b5699a514448ce994936c63f0d54 # v1.1.0
+        with:
+          deno-version: v1.x
+      - run: PUPPETEER_PRODUCT=chrome deno run -A --unstable https://deno.land/x/puppeteer@16.2.0/install.ts
+      # - run: deno task test:browser # TODO: uncomment when all tests in example folder fixed
```
tamrai, line 20: note: https://github.com/paritytech/capi/issues/328#issuecomment-1382020099
> It is worth noting that some of the examples will currently fail to run in the browser, as they attempt to spin up a dev chain, which currently hits a raw TCP server – not something the browser supports. Fixing this is out of the scope of this issue, and will likely be fixed along with #471.

tjjfvi:
We're going to take the lessons learned from this and incorporate it into a more holistic example-testing strategy.

PR #535, opened 2023-01-31
chore: move import map contents into deno.jsonc

harrysolovay:
Now possible thanks to [Deno v1.30.0](https://github.com/denoland/deno/releases/tag/v1.30.0).


examples/smart_contract.ts
```diff
@@ -1,6 +1,6 @@
 import * as C from "capi/mod.ts"
 
-import { client } from "capi/frame/zombienet/examples/smart_contract/zombienet.toml/collator01/@v0.9.360/mod.ts"
```
ryanleecode, line 3: what is the reason for this change?

ryanleecode: capi --> localhost that is

harrysolovay: Changed the import map example scope mapping of `capi/` to `capi/mod.ts` bc it was causing the frustrating import rewrite suggestions. This meant that `capi/frame...` became invalid. That being said, closing this out.

harrysolovay:
Keeping the import map separate is more elegant. Closing after all.

PR #537, opened 2023-01-31
chore: remove branded type fluff

harrysolovay:


tjjfvi (commented):
I would go one step further and say remove `Hex` and `Hash` as well

harrysolovay:
Two thoughts:

1. if we remove those types, let's do so after merging Rune (otherwise we'll unnecessarily touch upon many soon-to-be deleted effect files)
2. `Hex` and `Hash` do offer some legibility gains... I suppose good tsdocs will ultimately cover this.

tjjfvi (approved):

PR #539, opened 2023-02-01
chore: misc ss58 cleanup

harrysolovay:
Previously, errors were returned from this utility. This is not a common pattern (even though we do so in Zones/soon Rune). Figured it should be split out of the raw ss58 utils.

### DIFF
--- a/util/ss58.test.ts
+++ b/util/ss58.test.ts
@@ -1,4 +1,5 @@
 import { assertEquals, assertInstanceOf } from "../deps/std/testing/asserts.ts"
+import { returnThrows } from "./error.ts"
 import * as ss58 from "./ss58.ts"
 import { alice } from "./test_pairs.ts"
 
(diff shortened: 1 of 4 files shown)


util/ss58.ts
```diff
@@ -1,45 +1,27 @@
...
   validNetworkPrefixes?: readonly number[],
-) => {
-  const encodeRawResult = encodeRaw(prefix, pubKey, validNetworkPrefixes)
-  if (encodeRawResult instanceof Error) return encodeRawResult
-  return base58.encode(encodeRawResult)
+): string {
+  return base58.encode(encodeRaw(prefix, pubKey, validNetworkPrefixes))
 }
+
 export const encodeRaw = (
```
tjjfvi, line 12: Lets make the rest of these `function` declarationss instead of arrow functions as well.

util/ss58.test.ts
```diff
@@ -33,26 +34,37 @@ for (
 
 Deno.test("ss58.encode invalid public key length", () => {
   assertInstanceOf(
```
tjjfvi, line 36: I think these would be cleaner as `assertThrows` instead.

util/ss58.test.ts
```diff
@@ -32,27 +32,36 @@ for (
 }
 
 Deno.test("ss58.encode invalid public key length", () => {
-  assertInstanceOf(
-    ss58.encode(0, alice.publicKey.slice(0, 30)),
-    ss58.InvalidPublicKeyLengthError,
+  assertThrows(
+    () => ss58.encode(0, alice.publicKey.slice(0, 30)),
+    ss58.InvalidPublicKeyLengthError.prototype.message,
```
tjjfvi, line 37: `assertThrows` can take an error subclass:```suggestion
    ss58.InvalidPublicKeyLengthError,
```

harrysolovay: OH AWESOME. Never saw that it had that overload. Nice!

tjjfvi (approved):

tjjfvi (approved):

PR #541, opened 2023-02-02
chore: support `@latest` and past-version-based codegen

harrysolovay:
- Enables imports to be specified with `@latest` as version
- frame providers naively search for specified prev version metadata (in order to use it to initialize the codegen instance)
- tidying the CLI code + write-to-disk feature
- Got rid of trufflehog workflow (check was failing... not super useful)


main.ts
```diff
@@ -24,31 +24,64 @@ if (help) {
...
-function runServe(port: number) {
-  if (src || out) throw new Error("Cannot simultaneously specify serve and write args")
+if (typeof serve_ === "string") {
+  const port = +(serve_ || 4646)
+  const href = `http://localhost:${port}`
+
+  let portErr: string | undefined
+  try {
+    if (await (await fetch(`${href}/capi_cwd`)).text() === Deno.cwd()) {
+      portErr = `Capi server already running on port ${port}`
```
tjjfvi, line 71: It shouldn't error in this case

main.ts
```diff
@@ -24,31 +24,64 @@ if (help) {
...
+if (src) {
+  const { providerId, generatorId, cacheKey, codegen } = await env.digest(src)
+  const codegenDest = path.join(dest, generatorId, providerId, cacheKey)
+  try {
+    await Deno.remove(codegenDest, { recursive: true })
+  } catch (_e) {}
+  const depPath = path.join(codegenDest, "capi.ts")
+  const capi_ = capi.startsWith("file://")
+    ? capi
+    : path.relative(path.join("..", depPath), path.join(Deno.cwd(), capi))
```
tjjfvi, line 49:
```suggestion
  const capi_ = /^\w+:\/\//.test(capi)
    ? capi
    : path.relative(path.join("..", depPath), path.join(Deno.cwd(), capi))
```

tjjfvi (approved):

PR #554, opened 2023-02-09
chore: zombienet manifest-loading logic + example cleanup

harrysolovay:
Now uses FS-watching to detect and load network manifest (instead of process logs). Zombienet example specifier cleanup as well.

Originally suggested by @kratico (thanks for recommending this approach).


providers/frame/zombienet.ts
```diff
@@ -58,6 +63,20 @@ export class ZombienetProvider extends FrameProxyProvider {
   zombienet(configPath: string): Promise<Network> {
     return this.zombienetMemo.run(configPath, async () => {
       const tmpDir = Deno.makeTempDirSync({ prefix: `capi_zombienet_` })
+      const watcher = Deno.watchFs([tmpDir])
+      const networkManifestPath = path.join(tmpDir, "zombie.json")
+      const configPending = deadline(
+        (async () => {
+          for await (const e of watcher) {
+            if (e.kind === "create" && e.paths.some((path) => path.endsWith(networkManifestPath))) {
```
tjjfvi, line 71: Why `.endsWith`?

providers/frame/zombienet.ts
```diff
@@ -58,6 +63,20 @@ export class ZombienetProvider extends FrameProxyProvider {
   zombienet(configPath: string): Promise<Network> {
     return this.zombienetMemo.run(configPath, async () => {
       const tmpDir = Deno.makeTempDirSync({ prefix: `capi_zombienet_` })
+      const watcher = Deno.watchFs([tmpDir])
```
tjjfvi, line 66: If it times out, it won't close the watcher.

providers/frame/zombienet.ts
```diff
@@ -82,27 +101,12 @@ export class ZombienetProvider extends FrameProxyProvider {
       })
       const out = new Buffer()
```
tjjfvi, line 102: Perhaps this would work instead:
```ts
const config = await configPending.catch(() => undefined)
if(config) return config
await copy(process.stderr, Deno.stderr)
throw new Error("Zombienet exited without launching network")
```

providers/frame/zombienet.ts
harrysolovay: For some reason, these paths are prefixed with `/private/`, whereas the `networkManifestPath` is not. Will test to make sure this is true in other envs.

providers/frame/zombienet.ts
harrysolovay: Used `copy` to pipe buffer into `Deno.stderr`, but the other changes result in errors hanging (imports never load).

examples/block_author.ts
```diff
@@ -1,8 +1,6 @@
-import { consensus } from "capi/patterns"
```
harrysolovay, line 1: Given that we may reference chain-specific capi server routes from this patterns dir, we don't want to expose everything from the root (as it could spin up irrelevant test nodes).

providers/frame/zombienet.ts
```diff
@@ -58,6 +63,23 @@ export class ZombienetProvider extends FrameProxyProvider {
   zombienet(configPath: string): Promise<Network> {
     return this.zombienetMemo.run(configPath, async () => {
       const tmpDir = Deno.makeTempDirSync({ prefix: `capi_zombienet_` })
+      const watcher = Deno.watchFs(tmpDir)
+      const closeWatcher = () => {
+        if (typeof Deno.resources()[watcher.rid] === "number") watcher.close()
```
tjjfvi, line 68: Why do we need to do this?

providers/frame/zombienet.ts
```diff
@@ -58,6 +63,23 @@ export class ZombienetProvider extends FrameProxyProvider {
...
+          for await (const e of watcher) {
+            if (e.kind === "create" && e.paths.some((path) => path.endsWith(networkManifestPath))) {
+              closeWatcher()
+              return JSON.parse(await Deno.readTextFile(networkManifestPath)) as Network
+            }
+          }
+          return
+        })(),
+        this.timeout,
+      ).catch(closeWatcher)
```
tjjfvi, line 82:
```suggestion
      ).finally(closeWatcher)
```

providers/frame/zombienet.ts
harrysolovay: If we don't, then we're at risk of getting a bad resource ID error (in cases where we attempt to close the watcher twice, such as successful resolution of the the config, and then aborting of the server)

providers/frame/zombienet.ts
harrysolovay: Your way is working

providers/frame/zombienet.ts
tjjfvi: I think it would be cleaner to have something like
```ts
let closed = false
const closeWatcher = () => {
  if(closed) return
  closed = true
  watcher.close()
}
```
instead of relying on the resources

tjjfvi: You need to remove the explicit `closeWatcher()` before the `return JSON.parse`

server/local.ts
```diff
@@ -22,13 +23,15 @@ export function handler(env: Env) {
           return await provider.handle(request, pathInfo)
         } catch (e) {
           if (e instanceof Response) return e
-          return f.serverError(Deno.inspect(e))
+          const eText = Deno.inspect(e)
+          env.dbg && console.error(eText)
```
tjjfvi, line 27:
```suggestion
          if (env.dbg) console.error(eText)
```

providers/frame/zombienet.ts
harrysolovay: The specified watch dir is `/var/...`, yet the events' `paths` get populated with `/private/var/...`. I believe this is related to denoland/deno#10678, although there's no mention of this prefixing behavior elsewhere. `.endsWith` should be a reasonable safeguard, incase "private" is (for some reason) another name on different systems (or not present at all).

tjjfvi: `/var/` is likely symlinked to `/private/var/` on your system (as it is on mine)
```ts
const networkManifestPath = await Deno.realPath(path.join(tmpDir, "zombie.json"))
```

providers/frame/zombienet.ts
```diff
@@ -57,7 +58,26 @@ export class ZombienetProvider extends FrameProxyProvider {
   zombienetMemo = new PermanentMemo<string, Network>()
   zombienet(configPath: string): Promise<Network> {
     return this.zombienetMemo.run(configPath, async () => {
-      const tmpDir = Deno.makeTempDirSync({ prefix: `capi_zombienet_` })
+      const tmpDir = Deno.realPathSync(Deno.makeTempDirSync({ prefix: `capi_zombienet_` }))
```
tjjfvi, line 61: Why are these sync?

harrysolovay: Doesn't make a difference here, does it? Either way, just made both async.

tjjfvi: Not a huge deal, but it does block the whole process on those operations when they're sync.

fluent/Chain.ts
```diff
@@ -0,0 +1,20 @@
+import { Event } from "../primitives/mod.ts"
```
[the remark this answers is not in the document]
harrysolovay: Nope, just deleted

providers/frame/zombienet.ts
```diff
@@ -67,42 +85,32 @@ export class ZombienetProvider extends FrameProxyProvider {
...
-      this.env.signal.addEventListener("abort", async () => {
-        process.kill("SIGINT")
-        await process.status()
+      let processClosed = false
+      const closeProcess = () => {
+        if (processClosed) return
+        processClosed = true
+        process.kill()
         process.close()
+      }
```
tjjfvi, line 100:
```suggestion
      const closeProcess = () => {
        closeProcess = () => {}
        process.kill()
        process.close()
      }
```

server/local.ts
```diff
@@ -22,13 +23,15 @@ export function handler(env: Env) {
           return await provider.handle(request, pathInfo)
         } catch (e) {
           if (e instanceof Response) return e
-          return f.serverError(Deno.inspect(e))
+          const eText = Deno.inspect(e)
+          if (env.dbg) console.error(eText)
+          return f.serverError(eText)
```
tjjfvi, line 28: This is shorter and will give colors
```suggestion
          if (env.dbg) console.error(e)
          return f.serverError(Deno.inspect(e))
```

tjjfvi (approved):
I'm not the happiest with the closing logic but it's at least better than before.

tjjfvi (approved):

PR #555, opened 2023-02-09
feat: multisig proxy pattern

### LINKED ISSUE #525: add multisig proxy example
Add an example that utilizes a mutlisig behind a proxy. the example should be able to change who can access the proxy and thus also access the mutlisg.

ryanleecode:
Resolves #525

### DIFF
--- a/patterns/multisig/proxy.ts
+++ b/patterns/multisig/proxy.ts
@@ -2,8 +2,8 @@ import { types } from "polkadot_dev/mod.ts"
 import { Proxy, Utility } from "polkadot_dev/mod.ts"
 import { Event } from "polkadot_dev/types/pallet_proxy/pallet.ts"
 import { RuntimeEvent } from "polkadot_dev/types/polkadot_runtime.ts"
-import { ArrayRune, MultiAddress, Rune, RunicArgs } from "../mod.ts"
-import { hex, Sr25519 } from "../util/mod.ts"
+import { ArrayRune, MultiAddress, Rune, RunicArgs } from "../../mod.ts"
+import { hex, Sr25519 } from "../../util/mod.ts"
 
 export interface BuildPureProxyMultisigProps {
   sender: Sr25519
--- a/patterns/multisig/proxy.ts
+++ b/patterns/multisig/proxy.ts
@@ -1,9 +1,7 @@
-import { types } from "polkadot_dev/mod.ts"
 import { Proxy, Utility } from "polkadot_dev/mod.ts"
-import { Event } from "polkadot_dev/types/pallet_proxy/pallet.ts"
-import { RuntimeEvent } from "polkadot_dev/types/polkadot_runtime.ts"
-import { ArrayRune, MultiAddress, Rune, RunicArgs } from "../../mod.ts"
-import { hex, Sr25519 } from "../../util/mod.ts"
+
+import { ArrayRune, hex, MultiAddress, Rune, RunicArgs, Sr25519 } from "../../mod.ts"
+import { filterPureCreatedEvents } from "../proxy/mod.ts"
 
 export interface BuildPureProxyMultisigProps {
   sender: Sr25519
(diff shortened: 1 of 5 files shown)


patterns/multisig/proxy.ts
```diff
@@ -0,0 +1,136 @@
+import { types } from "polkadot_dev/mod.ts"
+import { Proxy, Utility } from "polkadot_dev/mod.ts"
+import { Event } from "polkadot_dev/types/pallet_proxy/pallet.ts"
+import { RuntimeEvent } from "polkadot_dev/types/polkadot_runtime.ts"
+import { ArrayRune, MultiAddress, Rune, RunicArgs } from "../../mod.ts"
+import { hex, Sr25519 } from "../../util/mod.ts"
```
harrysolovay, line 6: These are exposed through the root `mod.ts` as well (can move `hex, Sr25519` into the preceding import item list)

patterns/multisig/proxy.ts
```diff
@@ -0,0 +1,136 @@
...
+
+export interface BuildPureProxyMultisigProps {
+  sender: Sr25519
+  admins: Uint8Array[]
+  threshold?: number
+}
+
+export function buildPureProxyMultisig<X1>(props: RunicArgs<X1, BuildPureProxyMultisigProps>) {
```
harrysolovay, line 14: How do we feel about the name `MultiProxy`?

patterns/multisig/proxy.ts
```diff
@@ -0,0 +1,136 @@
...
+    multisig: Rune.rec({
+      signatories: proxySignatoryAccountIds,
+      threshold,
+    }),
+    stashMultiAddress,
+    adminProxies,
+  })
+}
+
+export function addProxy<X>(
```
harrysolovay, line 84: Can we move these proxy-only utilities into a `patterns/proxy/mod.ts`?

harrysolovay: Might be worth creating a simple proxy-only example as well

harrysolovay (changes requested):

patterns/multisig/proxy.ts
ryanleecode: it should be as idiomatic as possible IMO. I should be able to look at the official polkadot/substrate and easily connect it to whats in capi. PureProxyMultisig makes it obvious what it does.

patterns/multisig/proxy.ts
ryanleecode: > Might be worth creating a simple proxy-only example as well

Can do in a separate issue/pr

ryanleecode: https://github.com/paritytech/capi/issues/645

patterns/proxy/mod.ts
```diff
@@ -0,0 +1,59 @@
+import { types } from "polkadot_dev/mod.ts"
+import { Proxy } from "polkadot_dev/mod.ts"
+import { Event } from "polkadot_dev/types/pallet_proxy/pallet.ts"
+import { RuntimeEvent } from "polkadot_dev/types/polkadot_runtime.ts"
+import { hex, MultiAddress, Rune, RunicArgs } from "../../mod.ts"
+
+export function addProxy<X>(
+  ...[proxy, delegate]: RunicArgs<X, [proxy: MultiAddress, delegate: MultiAddress]>
```
harrysolovay, line 8: Perhaps we'd like to turn these into a Rune subclass (`Rune<Uint8Array /* proxy accountId */>`) as to enable the following?

```ts
const proxy = Rune.constant(multiaddr).into(ProxyRune)

await proxy
  .add(alice.address)
  .chain(() => proxy.remove(bob.address))
  .run()
```

harrysolovay: Perhaps even a `transfer` method which encapsulates the two steps.

```ts
await proxy
  .transfer(bob.address)
  .signed({ sender: alice })
  .sent()
  .finalized()
  .run()
```

patterns/proxy/mod.ts
```diff
@@ -0,0 +1,59 @@
...
+    forceProxyType: undefined,
+    call: Proxy.removeProxy({
+      proxyType: "Any",
+      delegate: delegate,
+      delay: 0,
+    }),
+  })
+}
+
+export function getMultiAddress<X1>(
```
harrysolovay, line 35: If I'm not mistaken, #583 will enable use to use avoid this helper?

```ts
const getProxyAddress  = adminProxies
  .access(accountId.map(hex.encode))
  .unhandle(undefined)
  .map(MultiAddress.Id)
  .rehandle(undefined, () => Rune.constant(new Uint8Array()))
```

patterns/proxy/mod.ts
```diff
@@ -0,0 +1,59 @@
...
+) {
+  return function<X2>(...[accountId]: RunicArgs<X2, [Uint8Array]>) {
+    return Rune
+      .tuple([mapping, accountId])
+      .map(([mapping, accountId]) =>
+        MultiAddress.Id(mapping[hex.encode(accountId)] ?? new Uint8Array())
+      )
+  }
+}
+
```
harrysolovay, line 46: Can get rid of this once we address #635

patterns/proxy/mod.ts
harrysolovay: It wouldn't be possible (given that there are multiple signers involved). This is a constraint of the utility pallet's batching methods, not this design.

harrysolovay:
Closing per #661

Issue #517, opened 2023-01-25
rune: add support for DI via env


tjjfvi:
Like #516, I think this should happen outside of #454.

harrysolovay:
For context: this will serve as a means of DI in Runes (for instance, injecting a custom logger, or runtime client)... perhaps a rune-ified paritytech/zones#20 (although may have a very different API, we shall see).

harrysolovay:
Perhaps our DI approach can satisfy the use case of #706: consumers of pattern libs could hook into parts of the pattern's execution at pattern-dev-defined waypoints.

harrysolovay:
Idea for this feature:

```ts
// 1. Define a key, to be used for applying to the placeholder.
const userId = Symbol()

// 2. Define a rune containing a placeholder(s) (the output of the `Rune._` factory).
const free = System.Account
  .value(Rune._(userId))
  .unhandle(undefined)
  .access("data", "free")

// 3. Apply a value/rune to the placeholder at any point in the tree.
const alexaFree = free.apply(userId, alexa.publicKey)
```

It's worth noting that in step 2, the type of `Rune._` is inferred from its context (in the 1st param position of the `value` method call). Typing this could prove convoluted, as we'd need to ensure that the same key cannot placehold items of different types.

This feature will be especially useful for the injection of custom lifecycle handlers into patterns. Consider the virtual multisig pattern: currently, this pattern is a black box. It creates many pure proxies, submits existential deposits, performs ownership swaps and more. In the future, this pattern can expose symbols to be used for (optionally) applying callbacks to hook placeholders.

```ts
await VirtualMultisigRune
  .deployment(chain, {
    founders: [alexa.publicKey, billy.publicKey, carol.publicKey],
    threshold: 2,
  })
  .signed(signature({ sender: alexa }))
  .apply(onOwnershipSwap, (events) => { /* ... */ })
  .run()
```

The root `Rune`'s `run` method would be `this`-typed as containing **no** placeheld items, unless they're typed as unioned with `undefined`. This would ensure that devs didn't accidentally `.run()` their Runes without having applied required types.

PR #585, opened 2023-02-15
feat: `Rune.str`, dbg/log cleanup, transaction sequence example

harrysolovay:
**1) `Rune.str`, a tagged template fn for interpolating Runes in a string**

```ts
const name = Rune.constant("Capi")
const message = Rune.str`Hello ${name}!`
```

**2) runified `valueRune.dbg` and `extrinsicStatusRune.logStatus`**

One can now specify a Rune as a prefix

**3) `rune.log`**

A duplicate of `dbg` (for now) –– I suppose we'll change this behavior upon resolution of #517

**4) transaction sequence example**

- log initial Alice balance
- sequentially do the following for Bob, Charlie and Dave
  - log initial balance
  - send funds from Alice to current address
  - log final balance
- log final Alice balance

### DIFF
--- a/examples/transfer_sequence.ts
+++ b/examples/transfer_sequence.ts
@@ -1,0 +1,36 @@
+import { alice, bob, charlie, dave, Rune, RunicArgs, Sr25519, ValueRune } from "capi"
+import { Balances, System } from "westend_dev/mod.ts"
+
+await aliceFree()
+  .log("Alice balance after transfers")
+  .chain(() =>
+    Object
+      .entries({ bob, charlie, dave })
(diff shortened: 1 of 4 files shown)


tjjfvi (commented):
There's not a PR tax ;)

The `Rune.str` is a nice utility, although I'll have some notes.

The changes to the various logging methods will need rework to avoid getting updates from the prefix.

I think the idea of the transfer sequence is reasonable, but I think having multiple transfers overcomplicates it.

harrysolovay:
Fair enough. Breaking apart now.

harrysolovay:
Closing in favor of #586, #587 and another PR, to be filed shortly thereafter.

PR #586, opened 2023-02-15
chore: log util cleanup

harrysolovay:
- make the `dbg` method on `ValueRune` accept Runes as prefixs
- make the `logStatus` of `extrinsicStatusRune` accept Runes as prefixes
- add a `log` method on `ValueRune` (currently does same thing as `dbg`, which will be reworked upon resolution of #517)

### DIFF
--- a/fluent/ExtrinsicStatusRune.ts
+++ b/fluent/ExtrinsicStatusRune.ts
@@ -15,12 +15,15 @@ export class ExtrinsicStatusRune<out U1, out U2, out C extends Chain = Chain>
   }
 
   logStatus(...prefix: unknown[]): ExtrinsicStatusRune<U1, U2, C> {
-    return this.into(ValueRune).map((rune) =>
-      rune.into(ValueRune).map((value) => {
-        console.log(...prefix, value)
-        return value
-      })
-    ).into(ExtrinsicStatusRune, this.extrinsic)
+    return Rune
+      .tuple([this.into(ValueRune), ...prefix])
+      .map(([rune, ...prefix]) =>
+        rune.into(ValueRune).map((value) => {
+          console.log(...prefix, value)
+          return value
+        })
+      )
+      .into(ExtrinsicStatusRune, this.extrinsic)
   }
 
   terminalTransactionStatuses() {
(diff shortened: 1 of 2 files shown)


tjjfvi (commented):
All of these will have updates from the prefix; you need to use `.lazy()` on prefix.

This still could have problems with filtering regarding the initial value, but this is a niche scenario that we can probably ignore.

fluent/ExtrinsicStatusRune.ts
```diff
@@ -14,13 +14,16 @@ export class ExtrinsicStatusRune<out U1, out U2, out C extends Chain = Chain>
...
-  logStatus(...prefix: unknown[]): ExtrinsicStatusRune<U1, U2, C> {
-    return this.into(ValueRune).map((rune) =>
-      rune.into(ValueRune).map((value) => {
-        console.log(...prefix, value)
-        return value
-      })
-    ).into(ExtrinsicStatusRune, this.extrinsic)
+  logStatus<X>(...prefix: RunicArgs<X, unknown[]>): ExtrinsicStatusRune<U1, U2, C> {
+    return Rune
+      .tuple([this.into(ValueRune), Rune.tuple(prefix).lazy()])
```
harrysolovay, line 19: > All of these will have updates from the prefix; you need to use .lazy() on prefix.

Is this what you mean?

tjjfvi: Yes

fluent/ExtrinsicStatusRune.ts
```diff
@@ -14,13 +14,16 @@ export class ExtrinsicStatusRune<out U1, out U2, out C extends Chain = Chain>
...
+      .map(([rune, prefix]) =>
+        rune.into(ValueRune).map((value) => {
+          console.log(...prefix, value)
+          return value
+        })
```
tjjfvi, line 24: This should use `.log()`

harrysolovay: Done

harrysolovay: Actually, does the placement of the prefix resolution matter (inside one of the inner Runes vs. wrapping `this` and `Rune.tuple(prefixes)` in another `Rune.tuple`, and then mapping their results into the inner runes)?

tjjfvi: Kind of but not sufficiently for me to care at the moment.

tjjfvi: Really I just need a `.trackMap` or `.metaMap` or whatever on `MetaRune` and we can use that.

tjjfvi (approved):

PR #587, opened 2023-02-15
feat: `Rune.str`

harrysolovay:
A tagged template fn for interpolating Runes in a string

```ts
const name = Rune.constant("Capi")
const message = Rune.str`Hello ${name}!`
```

### DIFF
--- a/rune/Rune.ts
+++ b/rune/Rune.ts
@@ -96,6 +96,21 @@ export class Rune<out T, out U = never> {
     return value instanceof Rune ? value.into(ValueRune) : Rune.constant(value)
   }
 
+  static str<X, T extends TemplateStringsArray, R extends unknown[]>(
+    ...args: RunicArgs<X, [T, ...R]>
+  ) {
+    return Rune
+      .tuple(args)
+      .map(([strings, ...values]) =>
+        strings
+          .map((templateString, i) => {
+            const value = values[i]
+            return value !== undefined ? `${templateString}${value}` : templateString
+          })
+          .join("")
+      )
+  }
+
   static tuple<R extends unknown[]>(
     runes: [...R],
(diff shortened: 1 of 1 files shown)


rune/Rune.ts
```diff
@@ -96,6 +96,21 @@ export class Rune<out T, out U = never> {
     return value instanceof Rune ? value.into(ValueRune) : Rune.constant(value)
   }
 
+  static str<X, T extends TemplateStringsArray, R extends unknown[]>(
+    ...args: RunicArgs<X, [T, ...R]>
+  ) {
```
tjjfvi, line 101: No need to take `T` and `R` as generics (this isn't zones :P) and no need to have `TempalteStringsArray` in the runic args
```suggestion
  static str<X>(strings: TemplateStringsArray, ...args: RunicArgs<X, unknown[]>) {
```

rune/Rune.ts
```diff
@@ -96,6 +96,21 @@ export class Rune<out T, out U = never> {
...
+    return Rune
+      .tuple(args)
+      .map(([strings, ...values]) =>
+        strings
+          .map((templateString, i) => {
+            const value = values[i]
+            return value !== undefined ? `${templateString}${value}` : templateString
```
tjjfvi, line 108: This `undefined` checking will have inconsistencies with regular template literals
```suggestion
            return i < values.length ? `${templateString}${values[i]}` : templateString
```

rune/Rune.ts
harrysolovay: Without extracting `T` and `R` into type params/runic args, the `U` doesn't propagate into the `Rune.str`-resulting Rune.

tjjfvi: Ah, because it was an `unknown[]`, an explicit `RunicArgs.resolve` was needed. I've pushed this fix.

tjjfvi (dismissed):

tjjfvi (approved):

Issue #589, opened 2023-02-16
option to transform `RunicArgs`


harrysolovay:
Let's say that we have the following Rune factory:

```ts
function greeting<X>(...args: RunicArgs<X, [name: string, isCool: boolean]>) {
  const [name, isCool] = RunicArgs.resolve(args)
  // ... do something with these Runes
}
```

What do we do if we want to transform one of the supplied Runes?

```ts
function greeting<X>(...args: RunicArgs<X, [name: string, isCool: boolean]>) {
  const name = Rune.resolve(args[0]).map(upperCaseFirst)
  const isCool = Rune.resolve(args[1])
  // ...
}
```

^ this isn't super clean. One place where this is especially apparent is in [the `instantiate` method of `InkMetadataRune`](https://github.com/paritytech/capi/blob/5d37aac773b38c4801e1b536e6e294c7f3ff3abc/patterns/ink/InkMetadataRune.ts#L71-L80), in which we want to fallback to sensible defaults for certain arguments when not provided.

## Solution

I propose that `RunicArgs.resolve` accept an optional second argument, a partial mapping from index to transform.

```ts
function greeting<X>(...args: RunicArgs<X, [name: string, isCool: boolean]>) {
  const [name, isCool] = RunicArgs.resolve(args, { 0: upperCaseFirst })
  //
}
```

In the case of the aforementioned `instantiate` method:

```diff
instantiate<X>(props: RunicArgs<X, InstantiateProps>) {
- const salt = Rune
-   .resolve(props.salt)
-   .unhandle(undefined)
-   .rehandle(undefined, this.salt)
- const value = Rune
-   .resolve(props.value)
-   .unhandle(undefined)
-   .rehandle(undefined, () => Rune.constant(0n))
+ const { salt, value, ... } = RunicArgs.resolve(props, {
+   salt: (v) => v ?? this.salt(),
+   value: (v) => v ?? 0n,
+ })

  // ...
```

PR #590, opened 2023-02-16
chore: transfer sequence example

harrysolovay:
An example that showcases sequential txs

### DIFF
--- a/examples/transfer_sequence.ts
+++ b/examples/transfer_sequence.ts
@@ -1,36 +1,45 @@
-// import { alice, bob, charlie, dave, Rune, RunicArgs, Sr25519, ValueRune } from "capi"
-// import { Balances, System } from "westend_dev/mod.ts"
+import { alice, bob, charlie, dave, Rune, RunicArgs, Sr25519, ValueRune } from "capi"
+import { Balances, System } from "westend_dev/mod.ts"
+import { upperFirstCase } from "../deps/case.ts"
 
-// await aliceFree()
-//   .log("Alice balance after transfers")
-//   .chain(() =>
-//     Object
-//       .entries({ bob, charlie, dave })
-//       .reduce<ValueRune<any, any>>((acc, args) =>
-//         acc
-//           .chain(() => beforeThenTxThenAfter(...args))
-//           .into(ValueRune), Rune.constant(undefined))
-//   )
-//   .chain(() => aliceFree().log("Alice balance after transfers"))
-//   .run()
+await aliceFree()
+  .log(msg("Alice", "before"))
+  .chain(() =>
+    Object
+      .entries({ bob, charlie, dave })
+      .reduce<ValueRune<any, any>>((acc, args) =>
+        acc
+          .chain(() => beforeThenTxThenAfter(...args))
+          .into(ValueRune), Rune.constant(undefined))
+  )
+  .chain(() => aliceFree().log(msg("Alice", "after")))
+  .run()
 
-// function aliceFree() {
-//   return System.Account.entry([alice.publicKey]).access("data", "free")
-// }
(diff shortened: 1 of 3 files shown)


examples/transfer_sequence.ts
```diff
@@ -0,0 +1,45 @@
+import { alice, bob, charlie, dave, Rune, RunicArgs, Sr25519, ValueRune } from "capi"
+import { Balances, System } from "westend_dev/mod.ts"
+import { upperFirstCase } from "../deps/case.ts"
+
+await aliceFree()
+  .log(msg("Alice", "before"))
+  .chain(() =>
+    Object
+      .entries({ bob, charlie, dave })
+      .reduce<ValueRune<any, any>>((acc, args) =>
```
harrysolovay, line 10: This isn't the prettiest (both the `reduce` call, and the `ValueRune` type args. How shall we clean this up? Is a new Rune utility required?

examples/transfer_sequence.ts
```diff
@@ -0,0 +1,45 @@
...
+  return Rune.str`${name} balance ${stage}:`
+}
+
+function aliceFree() {
+  return System.Account.entry([alice.publicKey]).access("data", "free")
+}
+
+function beforeThenTxThenAfter<X>(...args: RunicArgs<X, [name: string, pair: Sr25519]>) {
+  const pair = Rune.resolve(args[1])
+  const name = Rune.resolve(args[0]).map(upperFirstCase)
```
harrysolovay, line 28: #589

ryanleecode:
Why do it this way instead of via batch call?

harrysolovay:
> Why do it this way instead of via `batchAll` call?

Because the goal here is to show off modeling sequential txs with Capi's Rune system (not making use of the `Utility` pallet). Could be worth adding a designated `batchAll` example as well though.

ryanleecode:
```ts
import { alice, bob, charlie, dave, Rune, RunicArgs, Sr25519, ValueRune } from "capi"
import { Balances, System } from "westend_dev/mod.ts"
import { upperFirstCase } from "../deps/case.ts"

await getBalance(alice.publicKey)
  .log(msg("Alice", "before"))
  .chain(() =>
    Object
      .entries({ bob, charlie, dave })
      .reduce<ValueRune<any, any>>((acc, args) =>
        acc
          .chain(() => transfer(...args))
          .into(ValueRune), Rune.constant(undefined))
  )
  .chain(() => getBalance(alice.publicKey).log(msg("Alice", "after")))
  .run()

function msg<X>(...[name, stage]: RunicArgs<X, [name: string, stage: "before" | "after"]>) {
  return Rune.str`${name} balance ${stage}:`
}

function transfer<X>(...args: RunicArgs<X, [name: string, pair: Sr25519]>) {
  const pair = Rune.resolve(args[1])
  const name = Rune.resolve(args[0]).map(upperFirstCase)
  return getBalance(pair.access("publicKey"))
    .log(msg(name, "before"))
    .chain(() =>
      Balances
        .transfer({
          dest: pair.access("address"),
          value: 1000123n,
        })
        .signed({ sender: alice })
        .sent()
        .logStatus(Rune.str`Transfer to ${name} from alice:`)
        .finalized()
    )
    .chain(() => getBalance(pair.access("publicKey")).log(msg(name, "after")))
}

function getBalance<X>(...[pubKey]: RunicArgs<X, [pubKey: Uint8Array]>) {
  return System.Account.entry(Rune.tuple([pubKey])).access("data", "free")
}
```

Rewrote it a little and change the some naming to be more obvious

harrysolovay:
The only different I notice is that you've parameterized the public key of the balance read... not sure if that's necessary for this example. Less code / more legible if the public key is supplied directly... if I missed any other changes, want to suggest them as commits within the GH review UI?

tjjfvi (approved):

PR #603, opened 2023-02-19
ci: fix dnt & udd tasks

harrysolovay:


_tasks/dnt.ts
```diff
@@ -50,5 +52,5 @@ await Promise.all([
     typeCheck: false,
   }),
   fs.copy("LICENSE", path.join(outDir, "LICENSE")),
-  fs.copy("Readme.md", path.join(outDir, "Readme.md")),
+  fs.copy("help.txt", path.join(outDir, "help.txt")),
```
tjjfvi, line 55: We should probably still copy the readme

util/branded.ts
```diff
@@ -1,9 +1,4 @@
...
-
-declare const _hex: unique symbol
-export type Hex = Branded<string, typeof _hex>
-
-declare const _hash: unique symbol
-export type Hash = Branded<Uint8Array, typeof _hash>
-
-export type HexHash = Hex & Hash
+// TODO: delete these upon re-rpc merge
+export type Hex = string
```
tjjfvi, line 2: Why this, here?

_tasks/dnt.ts
harrysolovay: If we do, it'll contain dead links (given its in a different directory)

util/branded.ts
harrysolovay: DNT started throwing an error regarding serialization of symbols `hex_` and `hash_` (even if we export them from the root mod). This was the fix.

_tasks/dnt.ts
tjjfvi: Better to have a readme with dead links than no readme at all, IMO

tjjfvi, line 55:
```suggestion
  fs.copy("Readme.md", path.join(outDir, "Readme.md")),
  fs.copy("help.txt", path.join(outDir, "help.txt")),
```

tjjfvi (approved):

PR #605, opened 2023-02-19
chore: move multisig account derivation util into pattern

harrysolovay:
Moves the `multisigAddress` util into the multisig pattern and renames it to `multisigAccountId`.


patterns/mod.ts
```diff
@@ -1,3 +1,4 @@
 export * as consensus from "./consensus/mod.ts"
```
tjjfvi, line 1: I thought we got rid of this file

patterns/mod.ts
```diff
@@ -1,3 +1,4 @@
...
 export * as ink from "./ink/mod.ts"
+export * from "./multisigAccountId.ts"
 export * from "./MultisigRune.ts"
```
tjjfvi, line 4: These should be colocated in one directory.

harrysolovay: I agree (exposed through `patterns/multisig/mod.ts`), but there are several other changes we should merge first (otherwise, we'll end up rewriting many imports).

tjjfvi (approved):

PR #607, opened 2023-02-19
feat: `client.ifVersionGt`

harrysolovay:
Enables users to conditionally redirect to different runes based on current runtime version.

```ts
client.ifVersionGt(
  "0.9.36",
  Balances.transfer({
    value: 12345n,
    dest: bob.address,
  }),
  Balances.transfer({
    value: 56789n,
    dest: bob.address,
  }),
)
```


tjjfvi:
Per offline discussion, we'll want a more holistic approach.

PR #622, opened 2023-02-21
fix: enable dprint dep to be used offline

### LINKED ISSUE #423: import dprint dependency without runtime `fetch`

harrysolovay:
Fixes #423 

... sadly, we now depend on Shiki, which also stands in the way of offline runs :/ to-be-addressed in a another PR


harrysolovay:
> We need to hex- or base64- encode the wasm and export it from a ts file.

I've updated the `_tasks/downloads.ts` script to generate a `util/tsFormatter.ts`, which includes the text-encoded, hex-encoded string of wasm... it's quite large (10_299_088 characters)... browsers will likely panic given that it's not instantiated via streaming. Feels like I'm complicating things. Gonna table this for now and return to it at a later date.

tjjfvi:
> browsers will likely panic given that it's not instantiated via streaming

We only use dprint in the codegen server, so this shouldn't be an issue. Similar with `shiki` (hence why we're using `npm:shiki` for it).

harrysolovay:
> We only use dprint in the codegen server, so this shouldn't be an issue

Very good point... we can revisit the browser-compat concern if we ever want to utilize the formatter for anything in `capi.dev`'s front end.

Meanwhile, `shiki` attempts to fetch the theme asset from a unpkg; this is a blocker to running the server offline. Any ideas for temporary fixes? (I'd imagine we'll think about alternative highlighting tools / potentially monaco in the longer term?)

tjjfvi (commented):
Looks good, but the `deps/dprint_ts.wasm` should be deleted.

tjjfvi (approved):

tjjfvi (approved):

tjjfvi (approved):

Issue #481, opened 2022-12-22
Storage Fluent API


kratico:
Improve storage fluent API by narrowing to storage value and storage map operations

For example
```
// specific to a storage value
const storageValue = createStorageValue(client, pallet, name)
storageValue(blockHash?) // or storageValue.read(blockHash?)
storageValue.raw(blockHash?)
storageValue.watch(listener)

// specific to a storage NMap
const storageMap = createStorageMap(client, pallet, name)
storageMap(...keys).keys(count, start?, blockHash?)
storageMap(...keys).values(count, start?, blockHash?) // note: the return value will need the last key to request the next key page
storageMap(...keys).rawValues(count, start?, blockHash?)
storageMap(...keys).watchValues(listener, count, start?)

// not specific but could be an alternative because storage values and NMap are storage items
const storage = createStorageItem(client, pallet, name)
storage.value(blockHash?)
storage.rawValue(blockHash?)
storage.watchValue(listener)
storage.map(...keys).keys(count, start?, blockHash?)
storage.map(...keys).values(count, start?, blockHash?)
storage.map(...keys).rawValues(count, start?, blockHash?)
storage.map(...keys).watchValues(listener, count, start?)
```

Note: the [StorageNMap](https://paritytech.github.io/substrate/master/frame_support/storage/types/struct.StorageNMap.html) key is encoded as follows in substrate
```
Twox128(Prefix::pallet_prefix())
  ++ Twox128(Prefix::STORAGE_PREFIX)
  ++ Hasher1(encode(key1))
  ++ Hasher2(encode(key2))
  ++ ...
  ++ HasherN(encode(keyN))
```

And the [StorageValue](https://paritytech.github.io/substrate/master/frame_support/storage/types/struct.StorageValue.html)
```
Twox128(Prefix::pallet_prefix()) ++ Twox128(Prefix::STORAGE_PREFIX)
```

harrysolovay:
I believe we've tackled most of this functionality in a fairly expressive manner. For any future readers: if you see limitations in the current storage API, please open shortcoming-specific issues.

PR #641, opened 2023-02-22
feat: metadata rework

### LINKED ISSUE #386: partial key codec

### LINKED ISSUE #412: use codegen codecs instead of derived codecs
Ideally chain metadata should not be requested when using codegen.

### LINKED ISSUE #499: multiple chains x same module
Let's say we're using a Capi pattern lib `multisig_util`.

```ts
import * as multisigUtil from "multisig_util"
```

Let's use an import map to specify the chain.

`import_map.json`

```json
{
  "scopes": {
    "multisig_util/": {
      "#chain": "<the-compass>"
    }
  }
}
```

What if we want to utilize this lib twice in the same file, with two different chains? Given this approach, we cannot.

On one hand, we could design libs such that the "compass" needs to be injected into instances at runtime.

```ts
import { MultisigUtil } from "multisig_util"

const multisigUtil = new MultisigUtil(client)
```

But this is lackluster, as one does not get the type-checking against the target chain (not ideal).

I believe we'll need to tackle this via the codegen server (#471)

```ts
import * as multisigUtil from "http://localhost:8000/...compass.../lib/multisig_util/mod.ts"
```

How would this look with the import-mapped lead?

```ts
import * as multisigUtil from "#polkadot/lib/multisig_util/mod.ts"
```

I suppose any valid import spec could follow `/lib` (as well as anything defined within the import map)

### LINKED ISSUE #514: use custom metadata format based on scale codecs

### LINKED ISSUE #558: generating `.js` and `.d.ts` instead of `.ts`
Currently, Capi's codegen outputs raw TypeScript files. Although this poses no problem for Deno users, other users must incorporate additional steps into their workflow (for instance, import-mapping to the generated TS and then using a build tool such as esbuild). Users could forgo this step if Capi's codegen were to output plain JavaScript and corresponding declaration files.

### LINKED ISSUE #562: redundancy in codegen typing
Let's say we're utilizing a codegened `client`.

```ts
import { client } from "polkadot_dev/mod.ts"
```

We can get the strongly-typed events.

```ts
;(await client.latestBlock
  .events()
  .run()).forEach((event) => {
    event.event.type // narrow / string-literals of pallet names
  })
```

However, other bindings of `client` are untyped. For instance, we can bind to non-existent pallet names.

```ts
const result = await client
  .metadata()
  .pallet("J Cole is a great rapper")
  .run()
```

Granted, we want devs to make use of the generated pallet bindings, which safely wrap pallet access... still, unless we're going to hide the `pallet` method from the generated client, this feels inconsistent.

Unless the checker performance cost is too high, I propose that we do both (chain-specific types passed through client-rooted API + explicitly re-typed bindings that wrap common access points).

### LINKED ISSUE #563: chain-specific extrinsic info reflected in codegen + codec
@tjjfvi previous implemented [the beginnings of this](https://github.com/paritytech/capi/blob/9a296a91d0d3890655337601fa847344366febc0/codegen/genMetadata.ts#L118-L131).

### LINKED ISSUE #564: how should we expose metadata types?
Perhaps to be folded into #514:

Currently, metadata types/utils are exported from the root `mod.ts` under the `M` namespace. Meanwhile, we `*` export from `scale_info/mod.ts`. This is wonky. How do we want to approach these names/renames + exposing from root without conflicts?

Ideally the solution doesn't involve bloated names (such as changing `MapStorageEntryType` to `MapStorageEntryTypeMetadata`). Thoughts?

### LINKED ISSUE #583: ability to specify runes as arguments to `rune.access`
```ts
declare const v: ValueRune<{ a: { b: { c: "d" } } }, never>
declare const aK: Rune<"a", never>
declare const bK: Rune<"c", never>
const d = v.access(aK, "b", bK)
```

### LINKED ISSUE #588: make optional call data fields omitable
For instance.

```ts
import { MultiAddress } from "capi"
import { Balances, Proxy } from "polkadot_dev/mod.ts"

declare const real: Uint8Array
declare const dest: Uint8Array

Proxy.proxy({
  real,
  call: Balances.transfer({
    dest,
    value: 1_000_000n,
  }),
  forceProxyType: undefined,
})
```

In this example, removing the `forceProxyType` field will unnecessarily provoke a type error.

### LINKED ISSUE #627: star task not exiting in CI
This has happened before (we were forgetting to explicitly close specific `Deno.run`-spawned processes. Now it seems there are many timer resources left open (preventing the unload event from firing). We could do a brute `Deno.exit`... but that's an admission of defeat to a leak.

### LINKED ISSUE #631: make `star` task run at light speed
`deno task star` used to type-check the entire codebase in a matter of milliseconds. Given the addition of the codegen/testing server, and subsequent usage of large amounts of generated artifacts throughout the codebase, star now takes as long as 16 minutes. Thanks to merge queues, this is less of an immediate concern... although it certainly hinders productivity. Let's aim to have star speedy again before `0.1.0`.

### LINKED ISSUE #637: support offline tx creation
Should be relatively simple. Requires the ability to manually specify some metadata. The rest is up to the developer (running the `SignedExtrinsicRune` to get the encoded tx).

### LINKED ISSUE #665: Concurrent execution of `zombienet` sometimes fail
## Bug Report

Concurrent `zombienet` bootstrapping seems to trigger this error https://github.com/paritytech/capi/actions/runs/4279205187/jobs/7449680066#step:8:45
which is related to how `zombienet` is packaged with `pkg` https://github.com/vercel/pkg/issues/1829

### Current Behavior

In `CI` the `Star` workflow timeouts. 

### Expected Behavior

In `CI` the `Star` workflow should not timeout or it should handle the error

### LINKED ISSUE #703: clean up `ChainError` type
The stack trace is unnecessary + might be good to add a message that clarifies the error is not from the JS env.

221949314-eb99d26b-4de0-46b4-92dd-c65a2e505bda

### LINKED ISSUE #730: zombienet / statemine weirdness
```
SyntaxError: Unterminated string in JSON at position 524288
    at JSON.parse (<anonymous>)
    at ZombienetProvider.network (file:///github/workspace/providers/frame/zombienet.ts:44:21)
    at async file:///github/workspace/providers/frame/zombienet.ts:32:14
SyntaxError: Unterminated string in JSON at position 524288
    at JSON.parse (<anonymous>)
    at ZombienetProvider.network (file:///github/workspace/providers/frame/zombienet.ts:44:21)
    at async file:///github/workspace/providers/frame/zombienet.ts:32:14
SyntaxError: Unterminated string in JSON at position 524288
    at JSON.parse (<anonymous>)
    at ZombienetProvider.network (file:///github/workspace/providers/frame/zombienet.ts:44:21)
    at async file:///github/workspace/providers/frame/zombienet.ts:32:14
SyntaxError: Unterminated string in JSON at position 524288
    at JSON.parse (<anonymous>)
    at ZombienetProvider.network (file:///github/workspace/providers/frame/zombienet.ts:44:21)
    at async file:///github/workspace/providers/frame/zombienet.ts:32:14
SyntaxError: Unterminated string in JSON at position 524288
    at JSON.parse (<anonymous>)
    at ZombienetProvider.network (file:///github/workspace/providers/frame/zombienet.ts:44:21)
    at async file:///github/workspace/providers/frame/zombienet.ts:32:14
SyntaxError: Unterminated string in JSON at position 524288
    at JSON.parse (<anonymous>)
    at ZombienetProvider.network (file:///github/workspace/providers/frame/zombienet.ts:44:21)
    at async file:///github/workspace/providers/frame/zombienet.ts:32:14
Download http://localhost:4646/mod.ts
Download http://localhost:4646/frame/zombienet/zombienets/statemine.toml/alice/@latest/mod.ts
Download http://localhost:4646/frame/zombienet/zombienets/statemine.toml/alice/@latest/types/pallet_xcm/pallet.ts
Download http://localhost:4646/frame/zombienet/zombienets/statemine.toml/alice/@latest/types/rococo_runtime/mod.ts
Download http://localhost:4646/frame/zombienet/zombienets/statemine.toml/collator/@latest/mod.ts
Download http://localhost:4646/frame/zombienet/zombienets/statemine.toml/collator/@latest/types/cumulus_pallet_parachain_system/pallet.ts
Download http://localhost:4646/frame/zombienet/zombienets/statemine.toml/collator/@latest/types/statemine_runtime.ts
error: Import 'http://localhost:4646/frame/zombienet/zombienets/statemine.toml/alice/@latest/mod.ts' failed: 500 Internal Server Error
    at file:///github/workspace/examples/xcm_asset_teleportation.ts:2:34
```

Saw this in some logs... not sure if its a concern or not.

https://github.com/paritytech/capi/actions/runs/4380576960/jobs/7667795286

tjjfvi:
Resolves #386 
Resolves #412
Resolves #499
Resolves #514
Resolves #558
Resolves #562
Resolves #563
Resolves #564 
Resolves #583 
Resolves #588
Resolves #627
Resolves #631
Resolves #637 
Resolves #665
Resolves #703
Resolves #730


examples/derived.ts
```diff
@@ -1,12 +1,11 @@
...
+import { ArrayRune, ValueRune } from "capi"
 import { Paras } from "polkadot/mod.ts"
 
 const result = await Paras
   .Parachains
-  .entry([])
   .unhandle(undefined)
   .into(ArrayRune)
-  .mapArray((id) => Paras.Heads.entry(Rune.tuple([id])))
+  .mapArray((id) => Paras.Heads.value(id))
```
harrysolovay, line 8:
```suggestion
  .mapArray(Paras.Heads.value)
```

util/normalize.ts
```diff
@@ -0,0 +1,8 @@
+export function normalizeIdent(ident: string) {
```
harrysolovay, line 1: Should we move these into the `codegen` dir?

scale_info/transformTys.ts
```diff
@@ -0,0 +1,108 @@
...
+
+/**
+ * All derived codecs for ZSTs will use this exact codec,
+ * so `derivedCodec === $null` is true iff the type is a ZST.
+ */
+export const $null = $.constant(null)
+
+export function transformTys(tys: Ty[]): [Codec<any>[], Record<string, Codec<any>>] {
+  const memo = new Map<number, Codec<any>>()
+  const paths: Record<string, Codec<any>> = {}
```
harrysolovay, line 16: - Assuming you're leaving populating this record for later?
- Where will be utilize `paths`?

scale_info/overrides/overrides.ts
```diff
@@ -0,0 +1,36 @@
...
+      ),
+    )
+  },
+  BTreeMap: (ty, visit) => {
+    return $.map(visit(ty.params[0]!.ty!), visit(ty.params[1]!.ty!))
+  },
+  BTreeSet: (ty, visit) => {
+    return $.set(visit(ty.params[0]!.ty!))
+  },
+  "frame_support::traits::misc::WrapperOpaque": (ty, visit) => {
```
harrysolovay, line 27: Nit: would we want a `fmtPath` util from `scale_info`, which we could use like so:

```ts
fmtPath("frame_support", "traits", "misc, "WrapperOpaque")
```

frame_metadata/raw/v14.ts
```diff
@@ -0,0 +1,163 @@
...
+  $.field("pallets", $.array($pallet)),
+  $.field("extrinsic", $extrinsicDef),
+)
+
+export function transformMetadata(metadata: $.Native<typeof $metadata>): FrameMetadata {
+  const [types, paths] = transformTys(metadata.tys)
+  return {
+    types: paths,
+    pallets: Object.fromEntries(metadata.pallets.map((pallet): [string, Pallet] => [pallet.name, {
+      id: pallet.i,
```
harrysolovay, line 96: Is it worth also renaming `i` to `id` in the raw pallet metadata?

frame_metadata/raw/v14.ts
```diff
@@ -0,0 +1,163 @@
...
+
+export const $metadata = $.object(
+  $.field("magicNumber", $.constant<typeof magicNumber>(magicNumber, $.u32)),
+  $.field("version", $.constant<14>(14, $.u8)),
+  $.field("tys", $.array($ty)),
+  $.field("pallets", $.array($pallet)),
+  $.field("extrinsic", $extrinsicDef),
+)
+
+export function transformMetadata(metadata: $.Native<typeof $metadata>): FrameMetadata {
```
harrysolovay, line 91: Why is this contained within `raw/v14`? Should it be? After all, this gives us the non-raw/processed metadata.

frame_metadata/FrameMetadata.ts
```diff
@@ -0,0 +1,44 @@
+import * as $ from "../deps/scale.ts"
+
+export interface FrameMetadata {
```
harrysolovay, line 3: This processed metadata shape corresponds to v14 –– perhaps instead of `raw/v14.ts`, it might make more sense to have a `v14` directory, within which we could place both the raw codec and normalized declaration, along with the transformation from raw to normalized?

scale_info/transformTys.ts
tjjfvi: Later in this PR, yes.

`paths` will be a e.g.
```ts
{
  "sp_runtime.multiaddress.MultiAddress": $.taggedUnion(...),
  ...
}
```
and we'll primarily use this to populate the `types/` directory in the codegen.

scale_info/overrides/overrides.ts
tjjfvi: This `::` format is only used here, so I don't think this is worthwhile.

I don't think
```ts
[fmtPath("frame_support", "traits", "misc", "WrapperOpaque")]: ...
```
is cleaner than
```ts
"frame_support::traits::misc::WrapperOpaque": ...
```
anyway.

frame_metadata/raw/v14.ts
tjjfvi: Because this transformation logic is specific to `raw/v14` – or in other words, this ensures that the only code that cares about the raw format is contained in the `raw` directory.

frame_metadata/FrameMetadata.ts
tjjfvi: This processed metadata shape is for *all* versions. As new versions come out, we'll add more (likely optional) fields to this, and different raw frame metadata versions will all transform into this one shape.

providers/frame/FrameProvider.ts
```diff
@@ -63,26 +64,29 @@ export abstract class FrameProvider extends Provider {
...
-        )
-        return metadata
-      },
-    )
+    return this.metadataMemo.run(this.cacheKey(pathInfo), async () => {
+      const raw = await this.env.cache.getRaw(
+        `${this.cacheKey(pathInfo)}/metadata`,
+        async () => {
+          if (pathInfo.vRuntime !== await this.latestVersion(pathInfo)) {
+            throw f.serverError("Cannot get metadata for old runtime version")
```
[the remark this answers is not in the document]
tjjfvi: (to be clear, this error was here before, just at a different indent level)

harrysolovay:
Will also close #481 if I'm not mistaken?

util/normalize.ts
tjjfvi: No; these are primarily used not in the codegen but in the metadata preprocessing.

providers/frame/common.ts
```diff
@@ -1,8 +1,7 @@
-import { File } from "../../codegen/frame/mod.ts"
 import { ss58, testUser } from "../../crypto/mod.ts"
 import * as $ from "../../deps/scale.ts"
 
-const DEFAULT_TEST_USER_COUNT = 100_000
+const DEFAULT_TEST_USER_COUNT = 1_000
```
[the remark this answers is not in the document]
tjjfvi: #741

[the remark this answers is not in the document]
kratico: Interesting, locally I didn't noticed a meaningful difference.

What do you mean by pre-compute?

Was the blocking related to this `for` loop?
https://github.com/paritytech/capi/blob/da4d7a94036b2da54e0ba981d70df01379195fbb/providers/frame/common.ts#L17-L22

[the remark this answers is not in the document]
tjjfvi: Yes, that for-loop was taking about 7 seconds on my machine, and it seems about 30 on ci.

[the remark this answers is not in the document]
tjjfvi: > What do you mean by pre-compute?

All of the test-users are deterministic, so we should be able to take the 100_000 test user addresses, compute them, and store them in a file somewhere, rather than computing them every time.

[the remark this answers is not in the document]
kratico: perfect!

harrysolovay (approved):

Issue #567, opened 2023-02-12
key page pagination happy path


harrysolovay:
How might we approach simplifying iterating over key pages / pagination? Ideally, one wouldn't need manually track the offset. What's the ideal DX? Maybe...

```ts
const count = 100
const p = paginator(System.Account, count)

p.next() // `Promise<[Uint8Array][]>`

for await (const keyPage of p) {
  keyPage // [Uint8Array][]
}
```

In this case, `p` stores state about the current offset (outside of the effect system).

tjjfvi:
```ts
const count = 100
const pages = System.Account.keyPages(count);
pages // Rune<[Uint8Array][]>

for await (const keyPage of pages.watch()) {
  keyPage // [Uint8Array][]
}

const allPages = await pages.collect().run()
allPages // [Uint8Array][][]
```

harrysolovay:
Ah, so this will retain the run state (aka. last position)? If so, is `watch` the correct method name? Perhaps we should rename this to `iter`?

tjjfvi:
`.watch()` is the method on `Rune`, though I don't have an objection to it being renamed to `.iter()`.

harrysolovay:
I don't think utilizing `iter` for pagination is the right way forward. An example that illustrates why:

Here, we would expect for the evaluated `page` to contain the keypages corresponding to two different blocks

```ts
const iter = Rune.rec({
  page: System.Account.keyPage(50),
  events: block.events(),
}).iter()

await iter.next() // { page: [Uint8Array][]; events: EventRecord[] }
await iter.next() // { page: [Uint8Array][]; events: EventRecord[] }
```

With this in mind, the rename of `watch` to `iter` isn't quite as fitting as we thought.

A better approach might be to revert the name change and create an explicit `KeyPageRune`/`EntryPageRune`, each of which could implement an `iter` method, returning a `MetaRune`, with the custom pagination logic boiled in.

Thoughts @tjjfvi?

PR #644, opened 2023-02-22
feat: entry page retrieval

### LINKED ISSUE #643: Fetch all the items from a StorageMap
## Feature Request

I have the following FRAME Storage Item: A StorageMap with an id as key and a struct as a value.

```rust
#[pallet::storage]
#[pallet::getter(fn projects)]
pub type Projects<T: Config> =
    StorageMap<_, Blake2_128Concat, T::ProjectIdentifier, ProjectOf<T>>;
```
It's very easy to fetch a single value using Capi:

```typescript
const result = await Funding.Projects.entry([project_identifier]).run()
```

On polkadot.js you can also *omit* the key and get back **all the content of the storage item**. e.g. 

I'm trying to replicate the same behaviour using Capi, without success. I tried:

```typescript
const assets = await Assets.Asset.entry([]).run()
```

but the parameter of the `entry(...)` method should be a `[number] | Rune<[number], never>`.

I also tried 

```typescript
const assets = await Assets.Asset.entry([] as any).run()
```

But I'm getting  `undefined` as result.

It would be very useful to be able to fetch an unknown quantity of elements as is done on polkadot.js, perhaps even implementing some kind of pagination. I am available for further details or necessary tests.

harrysolovay:
Resolves #643

Allows one to retrieve a list of entries as they do a list of keys.

```ts
import { Uniques } from "statemint/mod.ts"

const collections = Uniques.Class.entryPage(100)

const result = await collections.run()

console.log(result)
```

Will soon be complemented by #567.

> Note: was confused by the name of the result, storage changeset. Doesn't represent a diff after all.

### DIFF
--- a/examples/entry_page.ts
+++ b/examples/entry_page.ts
@@ -1,6 +1,6 @@
 import { Uniques } from "statemint/mod.ts"
 
-const collections = Uniques.Class.entryPage(100)
+const collections = Uniques.Class.entryPage(10)
 
 const result = await collections.run()
 
--- a/examples/entry_page.ts
+++ b/examples/entry_page.ts
@@ -1,0 +1,7 @@
+import { Uniques } from "statemint/mod.ts"
+
+const collections = Uniques.Class.entryPage(100)
+
+const result = await collections.run()
+
+console.log(result)
--- a/fluent/StorageRune.ts
+++ b/fluent/StorageRune.ts
@@ -52,6 +52,7 @@ export class StorageRune<in out K extends unknown[], out V, out U> extends Rune<
           })
         )
       })
+      .unsafeAs<[[K, V][]]>()
   }
 
   entryRaw<X>(...[key, blockHash]: RunicArgs<X, [key: K, blockHash?: HexHash]>) {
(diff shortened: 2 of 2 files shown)


tjjfvi (changes requested):
[`state_queryStorageAt` isn't the right method](https://github.com/paritytech/capi/blob/main/rpc/known/state.ts#L200-L207). Looks like we'll either need to use `mapArray` or give up paging and use `state_getPairs`.

harrysolovay:
This is the same approach used by pjs apps.

> Try reading statemine's uniques classes storage map. Same output.

getPairs is deprecated and calling it results in a server error.

This approach allows for retrieval with 2 round trips... that being said, yea: it's not the prettiest. Not sure if we have any better alternative though.

harrysolovay:
> The first entry doesn't, but subsequent entries do.

This is interesting. Yea, even if we did go this route, we'd need to skip decoding the subsequent diffs.

tjjfvi (approved):
Ah, I confused `state_queryStorageAt` with `state_queryStorage`. The former does a single block, the latter is the history-diffing one.

Eventually we should rename `.entry` to `.get` or `.value`, but for now I think we should leave it as-is.

PR #625, opened 2023-02-21
fix: debugging star exit

harrysolovay:

PR #648, opened 2023-02-23
fix: process resource cleanup in start task

kratico:
Based on trial and error, it seems that sometimes `zombienet` Deno resources are not properly released.
I haven't been able to identify the root cause.
The `unload` event is not fired in these cases.
```
self.addEventListener("unload", () => Deno.exit(status.code))
```

This PR 
- closes child processes started with `Deno.run`
- closes child processes `piped` file descriptors
- upgrades polkadot, polkadot-parachain and zombienet versions

Note: I've observed that the `star` task sometimes keeps running even after closing fd and processes 

Another theory to explore is that this could be correlated to a recent Deno version upgrade.

This is related to https://github.com/paritytech/capi/pull/625

### DIFF
--- a/providers/frame/zombienet.ts
+++ b/providers/frame/zombienet.ts
@@ -98,6 +98,8 @@ export class ZombienetProvider extends FrameProxyProvider {
         closeProcess = () => {}
         process.kill()
         process.close()
+        process.stdout.close()
+        process.stderr.close()
       }
       this.env.signal.addEventListener("abort", async () => {
         closeWatcher()
--- a/providers/frame/zombienet.ts
+++ b/providers/frame/zombienet.ts
@@ -97,7 +97,7 @@ export class ZombienetProvider extends FrameProxyProvider {
       let closeProcess = () => {
         closeProcess = () => {}
         console.log("closing zombienet rid:", process.rid)
-        process.kill("SIGKILL")
+        process.kill("SIGINT")
         process.stdout.close()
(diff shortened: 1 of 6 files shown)


providers/frame/zombienet.ts
```diff
@@ -96,7 +96,10 @@ export class ZombienetProvider extends FrameProxyProvider {
       })
       let closeProcess = () => {
         closeProcess = () => {}
-        process.kill()
+        console.log("closing zombienet rid:", process.rid)
+        process.kill("SIGINT")
```
kratico, line 100: zombienet handles `SIGINT`

https://github.com/paritytech/zombienet/blob/8b421410d803177abe866457443c601dc8b25aae/javascript/packages/cli/src/cli.ts#L298

harrysolovay (approved):
LGTM

harrysolovay (changes requested):
Let's actually hold off on merging this, as we will likely want to capture dev process std out for signaling boot progress #649

kratico:
> Let's actually hold off on merging this, as we will likely want to capture dev process std out for signaling boot progress #649

@harrysolovay 
Do you mean something like `await copy(process.stderr, Deno.stderr)` for the `PolkadotDevProvider`?

harrysolovay:
Both. I suppose we can tackle that in a subsequent PR. This looks good.

harrysolovay (approved):

PR #655, opened 2023-02-25
fix: make patterns consumable by node projects

harrysolovay:
Following this PR + publish, one will be able to utilize pattens in node projects.

> Note: this will require the project's `package.json` to contain a `type` of `"module"` and its tsconfig to have a `moduleResolution` strategy of `"esnext"`

```ts
import { MultisigRune } from "capi/patterns/multisig"
```


patterns/identity.ts
```diff
@@ -1,4 +1,7 @@
-import { Data, IdentityInfo as IdentityInfoRaw } from "polkadot_dev/types/pallet_identity/types.ts"
+import type {
```
tjjfvi, line 1: This `import type` won't be resolved for TS users of the NPM package.

tjjfvi (dismissed):

tjjfvi (approved):

PR #661, opened 2023-02-26
feat: virtual multisigs

harrysolovay:
Closes #658
Closes #555

This PR takes the learnings from #555 and integrates them into a `VirtualMultisigRune`. One can deploy a `VirtualMultisigRune` like so:

```ts
import { alice, bob, charlie, hex } from "capi"
import { VirtualMultisig } from "capi/patterns/multisig/mod.ts"
import { chain } from "polkadot_dev/mod.ts"

const state = await VirtualMultisigRune
  .deployment(chain, {
    founders: [alice.publicKey, bob.publicKey, charlie.publicKey],
    threshold: 2,
    deployer: alice,
  })
  .hex
  .run()

await Deno.writeTextFile("persist_this.txt", state)
```

To hydrate from the state hex, we do the following.

```ts
const vMultisig = VirtualMultisigRune.hydrate(chain, state)
```

To ratify a call from the stash account (signatory -> signatory proxy -> multisig -> stash), we can use virtual multisig's ratify method.

```ts
await vMultisig
  .ratify(bob.publicKey, proposal)
  .signed({ sender: bob })
  .sent()
  .dbgStatus("Bob ratify:")
  .finalized()
  .run()
```

> Note: ensure that you've funded the proxy (as it's initialized with no more than the existential deposit –– `fundMemberProxy`)

Open questions:

- Should `VirtualMultisig` depend on mappings from the member accounts to the proxy accounts? This would allow us to avoid the DX of passing signatory index into the `ratify` method.
- Voting signatories in and out (how might we elegantly persist this data?)
- Validation such as threshold and existing member of vmultisig

### DIFF
--- a/patterns/multisig/VirtualMultisigRune.ts
+++ b/patterns/multisig/VirtualMultisigRune.ts
@@ -19,9 +19,9 @@ import { Multisig, MultisigRune } from "./MultisigRune.ts"
 export interface VirtualMultisig extends Multisig {
   stash: Uint8Array
 }
-export const $virtualMultisig = $.object(
+export const $virtualMultisig: $.Codec<VirtualMultisig> = $.object(
   $.field("signatories", $.array($.sizedUint8Array(32))),
-  $.field("threshold", $.option($.u8)),
+  $.optionalField("threshold", $.u8),
   $.field("stash", $.sizedUint8Array(32)),
 )
 
--- a/patterns/multisig/VirtualMultisigRune.ts
+++ b/patterns/multisig/VirtualMultisigRune.ts
@@ -118,7 +118,7 @@ export class VirtualMultisigRune<out U, out C extends Chain = Chain>
       .signed({ sender: configurator })
       .sent()
       .dbgStatus("Proxy creation:")
-      .txEvents()
+      .finalizedEvents()
       .pipe(filterPureCreatedEvents)
       .map((events) =>
         events
--- a/patterns/multisig/VirtualMultisigRune.ts
+++ b/patterns/multisig/VirtualMultisigRune.ts
@@ -21,7 +21,7 @@ export interface VirtualMultisig extends Multisig {
 }
 export const $virtualMultisig = $.object(
   $.field("signatories", $.array($.sizedUint8Array(32))),
-  $.field("threshold", $.u8),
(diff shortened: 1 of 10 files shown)


patterns/multisig/VirtualMultisigRune.ts
```diff
@@ -0,0 +1,244 @@
...
+export interface VirtualMultisigDeploymentProps extends Multisig {
+  deployer: ExtrinsicSender
+  deposits?: bigint
+}
+
+export function filterPureCreatedEvents<X>(...[events]: RunicArgs<X, [Event[]]>) {
+  return Rune.resolve(events).map((events) =>
+    events
+      .map((e) => e.event)
+      .filter((event): event is RuntimeEvent.Proxy => event.type === "Proxy")
```
ryanleecode, line 240: Dont we have codegen guard for these?

tjjfvi:
> Closes https://github.com/paritytech/capi/pull/658
> Closes https://github.com/paritytech/capi/pull/555

Can we go ahead and close those now, for housekeeping purposes?

patterns/multisig/VirtualMultisigRune.ts
harrysolovay: We should avoid using codegened runtime code in patterns, as it makes them unusable to consuming NodeJS projects (for now).

patterns/multisig/VirtualMultisigRune.ts
```diff
@@ -0,0 +1,244 @@
...
+    const v = this.into(ValueRune)
+    this.signatories = v.access("signatories")
+    this.threshold = v.access("threshold")
+    this.stash = v.access("stash")
+    this.encoded = v.map((m) => $virtualMultisig.encode(m))
+    this.hex = this.encoded.map(hex.encode)
+  }
+
+  ratify<X>(...[senderIndex, call]: RunicArgs<X, [number, Chain.Call<C>]>) {
+    const call_ = this.chain.extrinsic(Rune.rec({
```
ryanleecode, line 50: Proxy call needs to happen outside of ratify. We can't assume every ratify call will interact with the stash

ryanleecode (approved):

PR #671, opened 2023-02-27
fix: smoldot connection race condition

kratico:
- Fixes subscription handler setup (`Connection.handle` was invoked before the subscription handler was set)
- Add unit test for Smoldot

### DIFF
--- a/rpc/Connection.ts
+++ b/rpc/Connection.ts
@@ -99,9 +99,8 @@ export abstract class Connection {
     if (typeof message.id === "number") {
       this.callResultPendings[message.id]?.resolve(message)
       delete this.callResultPendings[message.id]
-      if (!message.error && this.subscriptionPendingInits[message.id]) {
-        this.subscriptionPendingInits[message.id]!(message.result as string)
-      }
+      const init = this.subscriptionPendingInits[message.id]
+      if (!message.error && init) init(message.result as string)
     } else if (message.params) {
       this.subscriptionHandlers[message.params.subscription]?.(message)
     } else {
--- a/rpc/Connection.ts
+++ b/rpc/Connection.ts
@@ -51,10 +51,14 @@ export abstract class Connection {
 
   protected abstract close(): void
 
-  callResultPendings: Record<number, Deferred<RpcCallMessage>> = {}
   async call(method: string, params: unknown[]) {
-    await this.ready()
     const id = this.nextId++
+    return this.#call(id, method, params)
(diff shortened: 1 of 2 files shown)


harrysolovay (dismissed):
LGTM

rpc/Connection.ts
```diff
@@ -51,39 +51,56 @@ export abstract class Connection {
...
-    const message = await this.call(subscribe, params)
-    if (signal.aborted) return
-    if (message.error) return handler(message)
-    const subscriptionId = message.result as string
-    this.subscriptionHandlers[subscriptionId] = handler
-    signal.addEventListener("abort", () => {
-      delete this.subscriptionHandlers[subscriptionId]
-      this.send(this.nextId++, unsubscribe, [subscriptionId])
-    })
+    const id = this.nextId++
```
tjjfvi, line 77: Why were these changes to subscription reverted?

tjjfvi (changes requested):

rpc/Connection.ts
kratico: @harrysolovay @tjjfvi this is surfaced with smoldot because it `Connection.handle` is invoked before the subscription handler is set.
It may happen with WebSockets too but, most of the time, the first subscription message in WS is not very fast as in smoldot.

kratico: I believe that for this particular subscription, the `chainHead_unstable_follow` responds with 2 messages very quickly
- the response for the subscription id
- the response for the first subscription message

It's a microtask race condition

For the above subscription, this smoldot loop run 2 times 
```
    while (true) {
      try {
        const response = await Promise.race([
          this.listening,
          chain.nextJsonRpcResponse(),
        ])
        if (!response) break
        this.handle(JSON.parse(response))
      } catch (_e) {}
    }
```
before this `connection.subscription`
```
    const message = await this.#call(id, subscribe, params)
    if (signal.aborted) {
      delete this.subscriptionPendingInits[id]
      return
    }
    if (message.error) {
      delete this.subscriptionPendingInits[id]
      return handler(message)
    }
```

tjjfvi: Ah yes, I believe this is because `deferred.resolve` uses two microtask ticks (first to resolve the argument, then to resolve the real promise). I plan to report this to deno std (this, along with another related issue with deferred), but this is a good solution for now.

tjjfvi (approved):

Issue #687, opened 2023-03-01
multisig example with only stash as proxy


ryanleecode:
Example of using a multisig where the only pure proxy is the stash and all the signors are regular users

PR #689, opened 2023-03-01
fix: add tsFormatter to moderate excludes

### LINKED ISSUE #688: ts-formatter should be removed from util/mod.ts
ts-formatter should be removed from `util/mod.ts` and added to the `exclude` moderate otherwise, when the code is bundled with `esbuild`, it cannot be used in the browser.

ryanleecode:
Resolves #688

### DIFF
--- a/util/mod.ts
+++ b/util/mod.ts
@@ -1,4 +1,4 @@
-// moderate --exclude cache port.ts
+// moderate --exclude cache port.ts tsFormatter.ts"
 
 export * from "./case.ts"
 export * from "./clock.ts"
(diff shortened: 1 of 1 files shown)


harrysolovay (approved):

util/mod.ts
```diff
@@ -1,10 +1,9 @@
-// moderate --exclude cache port.ts
+// moderate --exclude cache port.ts tsFormatter.ts"
```
tjjfvi, line 1: This trailing quote isn't right.

harrysolovay: Whoops... @ryanleecode will you please remove the quote along with #687 ?

tjjfvi: #690

PR #666, opened 2023-02-27
chore: use `Deno.Command`, tweak provider timeouts

kratico:
Resolves https://github.com/paritytech/capi/issues/650

Migrate to Deno.Command from Deno.run

This is the same as #664 but with longer deadline to execute `polkadot` or `zombienet`

PR #695, opened 2023-03-02
feat: error unhandling method + decoding

harrysolovay:
Partially addresses #677

PR #701, opened 2023-03-03
fix: star task

kratico:
Revert a bad copy/paste from #666 and add extra logs for error debugging

### DIFF
--- a/_tasks/star.ts
+++ b/_tasks/star.ts
@@ -76,7 +76,7 @@ for (const [file, deps] of entries.slice(1)) {
   if (done.has(file)) continue
   console.log(file)
   const command = new Deno.Command(Deno.execPath(), {
-    args: ["info", "-r=http://localhost:4646/", "--json", "target/star.ts"],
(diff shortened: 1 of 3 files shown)


_tasks/star.ts
```diff
@@ -76,10 +76,14 @@ for (const [file, deps] of entries.slice(1)) {
   if (done.has(file)) continue
   console.log(file)
   const command = new Deno.Command(Deno.execPath(), {
-    args: ["info", "-r=http://localhost:4646/", "--json", "target/star.ts"],
+    args: ["cache", "--check", file],
```
kratico, line 79: bad copy paste from #666

tjjfvi:
🎉 Woo, there are type errors! 🎉 

(never thought I'd say that)

harrysolovay:
The type errors have to do with #695 –– `EventsRune` doesn't have a `map` method, hence we need to convert to `ValueRune` in various places. Pushing the fix here soon.

Issue #515, opened 2023-01-25
add server api to get unique test users


tjjfvi:
We can make as many test users as we want when spawning a dev chain, e.g.:
```ts
import { Blake2b } from "./deps/wat_the_crypto.ts"
import { Sr25519, ss58 } from "./util/mod.ts"

const chainSpecProcess = Deno.run({ cmd: ["polkadot", "build-spec", "--dev"], stdout: "piped" })

const chainSpec = JSON.parse(new TextDecoder().decode(await chainSpecProcess.output()))

const initialFunds = 10000000000000000
const testUserCount = 100

const testUsers = Array.from({ length: testUserCount }, (_, i) => testUser(i))

chainSpec.genesis.runtime.balances.balances = testUsers.map((x) => [
  ss58.encode(42, x.publicKey),
  initialFunds,
])

const chainSpecFile = await Deno.makeTempFile()

await Deno.writeTextFile(chainSpecFile, JSON.stringify(chainSpec))

await Deno.run({ cmd: ["polkadot", "--chain", chainSpecFile] }).status()

function testUser(i: number) {
  const hasher = new Blake2b(32)
  hasher.update(new TextEncoder().encode(`capi-test-user-${i}`))
  const seed = hasher.digest()
  return Sr25519.fromSeed(seed)
}
```

PR #725, opened 2023-03-08
feat: add custom chain spec

kratico:
Resolves https://github.com/paritytech/capi/issues/515

Update `PolkadotDevProvider`/`ContractsDevProvider` providers with a custom chain spec and add the `/user_i` endpoints

The custom chain spec add test accounts with an initial balance.

How to test?

Run the following example to validate that a test account has an initial balance.

```typescript
import { System, users } from "polkadot_dev/mod.ts"

const [user0, _user1, _user2] = await users(3)

const result = await System.Account.entry([user0.publicKey]).run()

console.log(result)

```

To test the API

```bash
curl -X post http://localhost:4646/frame/dev/polkadot/@latest/user_i -d '{"count":2}'     
{"index":4}

curl -X post http://localhost:4646/frame/contracts_dev/@latest/user_i -d '{"count":2}'
{"index":2}
```


kratico:
@harrysolovay @tjjfvi 

I explored how to get unique tests users doing something like
```
import users from http://localhost:4646/frame/dev/polkadot/@v0.9.37/test_users.ts
```
But, this is a bit challenging given that
- `deno` caches the imports
- in the CAPI server, the `polkadot` instance is shared by all imports to the same provider

There are a few ways to bypass the `deno` cache, for example we can tweak the CAPI server to respond differently based on
- `authorization` header, `DENO_AUTH_TOKENS=a1b2c3d4e5f6@localhost:4646 deno run script.ts`, but it will need a different token on every run
- `import {} from "http://localhost:4646/frame/dev/polkadot/@v0.9.37/test_users.ts?cacheId=1234`, but it will need a different query string on every import

tjjfvi:
Yeah, imports probably won't work well.

You'll want to add logic to the provider itself (override `handle`) to handle a request to a filePath of e.g. `/testUsers?count=5` and return the relevant user ids.

Then, you'll want to add a custom function to the codegen to call this route. The easiest way will be to override `chainFile` to add a function like
```ts
type ArrayOfLength<
  T,
  L extends number,
  A extends T[] = [],
> = number extends L ? T[]
  : L extends A["length"] ? A
  : ArrayOfLength<T, L, [...A, T]>

export async function getTestUsers<N extends number>(count: N): Promise<ArrayOfLength<C.Sr25519, N>>
export async function getTestUsers(count: number): Promise<C.Sr25519[]> {
  const userIds: number[] = await fetch(`.../testUsers?count=${count}`).then(r => r.json())
  return userIds.map(C.testUser)
}
```

Then, it can be used like
```ts
import { getTestUsers } from "polkadot_dev"

const [alex, bill, carol] = await getTestUsers(3)
```

kratico:
@tjjfvi I've just spoke with @harrysolovay and he suggested a similar approach.

Thank you both

harrysolovay:
Should we model this as a Rune factory instead?

```ts
import { System, user } from "polkadot_dev/mod.ts"

const alice = user()

const result = await System.Account.value(alice.publicKey).run()

console.log(result)
```

And let's also make sure to include an `ExceededMaximumTestUsersError`.

tjjfvi:
We need it to persist state between runs, so it wouldn't work well as a rune at the moment.

kratico:
@tjjfvi @harrysolovay Would you give it a quick pass to validate the provider `override handle` and `override chainFile` implementations?

I'd like to get a quick feedback before refactoring them for the `ContractsDevProvider`

providers/frame/polkadot_dev.ts
```diff
@@ -1,13 +1,27 @@
...
+import { createCustomChainSpec } from "./utils/mod.ts"
+
+const DEV_RUNTIME_PREFIXES = {
+  polkadot: 0,
+  kusama: 2,
+  westend: 42,
+  rococo: 42,
+} as const
+
+const TEST_USER_COUNT = 10
```
kratico, line 16: I'll update this after testing

providers/frame/polkadot_dev.ts
```diff
@@ -16,15 +30,100 @@ export class PolkadotDevProvider extends FrameBinProvider {
...
+        }
+        const currentCount = this.#testUserCountCache[this.#getRuntime(pathInfo)] ?? 0
+        if (count + currentCount > TEST_USER_COUNT) {
+          throw new Error("Maximum test user count reached")
+        }
+        const from = currentCount
+        const to = currentCount + count
+        this.#testUserCountCache[this.#getRuntime(pathInfo)] = to
+        return new Response(
+          JSON.stringify({ from, to }),
```
kratico, line 88: Would it make sense to return the seed instead?

providers/frame/utils/createCustomChainSpec.ts
```diff
@@ -0,0 +1,68 @@
+import { ss58, testUser } from "../../../crypto/mod.ts"
+
+const DEFAULT_TEST_USER_COUNT = 10
+const DEFAULT_TEST_USER_INITIAL_FUNDS = 1000000000000000000
```
kratico, line 4: Validated that this is safe for `JSON.stringify`

Note: it's not valid for integer operations because it's greater than `Number.MAX_SAFE_INTEGER`
```
> 1000000000000000000 + 1 === 1000000000000000000 + 2
true
```

providers/frame/utils/createCustomChainSpec.ts
```diff
@@ -0,0 +1,68 @@
+import { ss58, testUser } from "../../../crypto/mod.ts"
+
+const DEFAULT_TEST_USER_COUNT = 100_000
+const DEFAULT_TEST_USER_INITIAL_FUNDS = 1_000_000_000_000_000_000
```
kratico, line 4: Validated that this is safe for `JSON.stringify`

Note: it's not valid for integer operations because it's greater than `Number.MAX_SAFE_INTEGER`
```
> 1000000000000000000 + 1 === 1000000000000000000 + 2
true
```

providers/frame/polkadot_dev.ts
harrysolovay: Neither: just needs to return an index. Then––the front-end of the utility can instantiate the associated `Sr25519`s for the range from the index -> index + count.

kratico: so the current implementation, which returns a range, is fine, right?

providers/frame/utils/createCustomChainSpec.ts
```diff
@@ -0,0 +1,68 @@
...
+
+export interface CreateCustomChainSpecProps {
+  binary: string
+  chain: string
```
harrysolovay, line 8: Appears to be `dev` for both contracts and polkadot providers. Is this prop necessary?

providers/frame/utils/createCustomChainSpec.ts
```diff
@@ -0,0 +1,68 @@
...
+  testUserAccountProps: {
```
harrysolovay, line 9: Same question about these props. Let's not preemptively parameterize this chain-spec writing util.

providers/frame/utils/createCustomChainSpec.ts
```diff
@@ -0,0 +1,68 @@
...
+  )
+  const chainSpecRaw = JSON.parse(
+    new TextDecoder().decode((await buildSpecRawCmd.output()).stdout),
+  )
+  const customChainSpecRawPath = await Deno.makeTempFile({
+    prefix: `custom-${chain}-chain-spec-raw`,
+    suffix: ".json",
+  })
+  await Deno.writeTextFile(customChainSpecRawPath, JSON.stringify(chainSpecRaw, undefined, 2))
+  return customChainSpecRawPath
```
harrysolovay, line 67: Before returning, should we add an abort controller "abort" listener to trigger file cleanup?

providers/frame/polkadot_dev.ts
```diff
@@ -1,13 +1,27 @@
...
+} as const
+
+const TEST_USER_COUNT = 10
 
 export interface PolkadotDevProviderProps {
   polkadotPath?: string
 }
 
 export class PolkadotDevProvider extends FrameBinProvider {
+  #testUserCountCache: Record<string, number> = {}
```
tjjfvi, line 23: We use `Cache` in other places; let's rename this to avoid conflation
```suggestion
  #testUserCount: Record<string, number> = {}
```

providers/frame/polkadot_dev.ts
```diff
@@ -16,15 +30,100 @@ export class PolkadotDevProvider extends FrameBinProvider {
...
+        count: TEST_USER_COUNT,
+      },
+    })
+    const args: string[] = ["--tmp", "--alice", "--ws-port", port.toString(), "--chain", chainSpec]
     await this.runBin(args)
     return port
   }
+
+  #getRuntime(pathInfo: PathInfo) {
+    return pathInfo.target || "polkadot"
```
tjjfvi, line 125:
```suggestion
    const runtime = pathInfo.target
    $.assert($devRuntimeName, runtime)
    return runtime
```

providers/frame/polkadot_dev.ts
```diff
@@ -16,15 +30,100 @@ export class PolkadotDevProvider extends FrameBinProvider {
...
+            },
+          },
+        )
+      } catch (error) {
+        console.log("Error", request.url, error)
+        const message = error instanceof Error
+          ? error.message
+          : "Unknown error"
+        return new Response(message, { status: 400 })
+      }
```
tjjfvi, line 102: No need for this catch; the root handler has something that will suffice.

providers/frame/polkadot_dev.ts
```diff
@@ -16,15 +30,100 @@ export class PolkadotDevProvider extends FrameBinProvider {
...
+          "${url}",
+          {
+            method: "POST",
+            headers: {
+              "content-type": "application/json",
+            },
+            body: JSON.stringify({ count }),
+          },
+        )
+        if (response.status >= 400) {
```
tjjfvi, line 60:
```suggestion
        if (!response.ok) {
```

providers/frame/polkadot_dev.ts
```diff
@@ -16,15 +30,100 @@ export class PolkadotDevProvider extends FrameBinProvider {
...
+        L extends number,
+        A extends T[] = [],
+      > = number extends L ? T[]
+        : L extends A["length"] ? A
+        : ArrayOfLength<T, L, [...A, T]>
+
+      export async function users<N extends number>(count: N): Promise<ArrayOfLength<C.Sr25519, N>>
+      export async function users(count: number): Promise<C.Sr25519[]> {
+        const response = await fetch(
+          "${url}",
```
tjjfvi, line 51:
```suggestion
          ${JSON.stringify(url)},
```

providers/frame/polkadot_dev.ts
```diff
@@ -16,15 +30,100 @@ export class PolkadotDevProvider extends FrameBinProvider {
...
+        for (let i = from; i < to; i++) {
+          userIds.push(C.testUser(i))
+        }
+        return userIds
+      }
+    `)
+  }
+
+  override async handle(request: Request, pathInfo: PathInfo): Promise<Response> {
+    if (request.method.toUpperCase() === "POST" && pathInfo.filePath === "test-users") {
```
harrysolovay, line 74: I think a better way to approach this may be to use the pathInfo –– we can have "magic" file names. Maybe something like...

```ts
if (pathInfo.filePath === "user_i")
```

providers/frame/polkadot_dev.ts
```diff
@@ -16,15 +30,100 @@ export class PolkadotDevProvider extends FrameBinProvider {
...
+      try {
+        const { count } = await request.json()
+        if (!(count > 0)) {
+          throw new Error("'count' should be greater than 0")
+        }
+        const currentCount = this.#testUserCountCache[this.#getRuntime(pathInfo)] ?? 0
+        if (count + currentCount > TEST_USER_COUNT) {
+          throw new Error("Maximum test user count reached")
```
harrysolovay, line 82: No need to throw an error on the server. Best to return something that indicates it's out of bounds. The front-end util can then throw.

providers/frame/polkadot_dev.ts
```diff
@@ -16,15 +30,100 @@ export class PolkadotDevProvider extends FrameBinProvider {
...
+        networkPrefix: DEV_RUNTIME_PREFIXES[runtimeName],
+        count: TEST_USER_COUNT,
+      },
+    })
+    const args: string[] = ["--tmp", "--alice", "--ws-port", port.toString(), "--chain", chainSpec]
     await this.runBin(args)
     return port
   }
+
+  #getRuntime(pathInfo: PathInfo) {
```
harrysolovay, line 124: Why is extracting this necessary?

providers/frame/polkadot_dev.ts
```diff
@@ -16,15 +30,100 @@ export class PolkadotDevProvider extends FrameBinProvider {
...
+  }
+
+  override async handle(request: Request, pathInfo: PathInfo): Promise<Response> {
+    if (request.method.toUpperCase() === "POST" && pathInfo.filePath === "test-users") {
+      try {
+        const { count } = await request.json()
+        if (!(count > 0)) {
+          throw new Error("'count' should be greater than 0")
+        }
+        const currentCount = this.#testUserCountCache[this.#getRuntime(pathInfo)] ?? 0
```
harrysolovay, line 80: If `pathInfo.target` is an empty string, that should work fine too Any reason that we'd need to normalize?

harrysolovay (changes requested):

providers/frame/utils/mod.ts
```diff
@@ -0,0 +1 @@
+export * from "./createCustomChainSpec.ts"
```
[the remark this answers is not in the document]
kratico: yes

providers/frame/utils/createCustomChainSpec.ts
kratico: these are temp files, so adding the abort to clean up temporary files seems unnecessary

providers/frame/utils/createCustomChainSpec.ts
kratico: ok

providers/frame/polkadot_dev.ts
tjjfvi: I disagree; throwing a server error for this rare edge case is IMO a lot better than making the return something discriminated and introducing additional complexity.

providers/frame/polkadot_dev.ts
tjjfvi: I agree that the current extraction is unnecessary, but I think extracting the assert (as per my review comment on this) is warranted.

providers/frame/utils/createCustomChainSpec.ts
kratico: for contracts chain is `dev` but for the polkadot binary chain could be
- `polkadot-dev`
- `kusama-dev`
- `${runtime}-dev`

providers/frame/polkadot_dev.ts
kratico: it is to normalize the user count cache key 
as suggested previously, I'll remove the normailization

providers/frame/polkadot_dev.ts
kratico: no special reason, I'll remove the normalization

providers/frame/polkadot_dev.ts
harrysolovay: No reason to even prefix with `test` imo. Should be clear without.

providers/frame/polkadot_dev.ts
harrysolovay: Why do we need a range? The front-end knows the desired count. Returning just the start index should suffice.

kratico: You are right.
If the front-end request more users than the maximum an error will be thrown

providers/frame/polkadot_dev.ts
harrysolovay: But then we don't see the error from the perspective of the client––but rather from that of the server. I don't think this is preferable; from the server's perspective, there's no error. The server dulls out current test user indices until it reaches the max, at which point it could return a `-1`. The test scripts / clients are responsible for not passing the limit. Ie., they're the ones who step on the land mine / should incur the wrath of a throw.

harrysolovay: Will we want a custom error on the client side? If so, doesn't make a huge difference whether it's signaled via a 500 or a -1 body. Either way, triggering this error will happen rarely. Fine either way.

tjjfvi: > Will we want a custom error on the client side?

Not worth the additional complexity IMO. With 10000 if you hit the limit you're doing something very wrong.

providers/frame/polkadot_dev.ts
```diff
@@ -16,12 +27,84 @@ export class PolkadotDevProvider extends FrameBinProvider {
...
+          },
+        },
+      )
+    }
+    return super.handle(request, pathInfo)
+  }
+
   async launch(pathInfo: PathInfo) {
-    const runtimeName = pathInfo.target
+    const runtimeName = pathInfo.target || "polkadot"
```
tjjfvi, line 100: If there's no target, it should error.
```suggestion
    const runtimeName = pathInfo.target
```

providers/frame/polkadot_dev.ts
```diff
@@ -16,12 +27,84 @@ export class PolkadotDevProvider extends FrameBinProvider {
...
+      }
+    `)
+  }
+
+  override async handle(request: Request, pathInfo: PathInfo): Promise<Response> {
+    if (pathInfo.filePath === "user_i") {
+      const { count } = await request.json()
+      if (!(count > 0)) {
+        throw new Error("'count' should be greater than 0")
+      }
```
tjjfvi, line 78:
```suggestion
      const body = await request.json()
      $.assert($.field("count", $.u32), body)
      const { count } = body
```

providers/frame/polkadot_dev.ts
```diff
@@ -1,13 +1,24 @@
...
+  westend: 42,
+  rococo: 42,
+} as const
+
 export interface PolkadotDevProviderProps {
   polkadotPath?: string
 }
 
 export class PolkadotDevProvider extends FrameBinProvider {
+  #userCount: Record<string, number> = {}
```
tjjfvi, line 20: Let's narrow the key, and IMO there's no need for this to be #private
```suggestion
  userCount: Record<DevRuntimeName, number | undefined> = {}
```

providers/frame/polkadot_dev.ts
```diff
@@ -16,12 +27,84 @@ export class PolkadotDevProvider extends FrameBinProvider {
...
+    `)
+  }
+
+  override async handle(request: Request, pathInfo: PathInfo): Promise<Response> {
+    if (pathInfo.filePath === "user_i") {
+      const { count } = await request.json()
+      if (!(count > 0)) {
+        throw new Error("'count' should be greater than 0")
+      }
+      let index = this.#userCount[pathInfo.target!] ?? 0
```
tjjfvi, line 79:
```suggestion
      $.assert($devRuntimeName, pathInfo.target)
      let index = this.#userCount[pathInfo.target] ?? 0
```

providers/frame/polkadot_dev.ts
kratico: this is optional in the `PathInfo` which is parsed using this regex

If I follow correctly, `target` is optional in the regex too

https://github.com/paritytech/capi/blob/3aee893724fd73ff451986bd6eb2c21fd2b42f24/server/PathInfo.ts#L1-L11

providers/frame/polkadot_dev.ts
kratico: Any suggestion to extract the runtime names from 
```
const $devRuntimeName = $.literalUnion(["polkadot", "kusama", "westend", "rococo"])
```
or should I just do
```
keyof typeof DEV_RUNTIME_PREFIXES
```

providers/frame/polkadot_dev.ts
kratico: nvm, I got the suggestion

kratico:
@tjjfvi @harrysolovay Shall I move these `.handle` and `.chainFile` overrides to the `FrameBinProvider`?

That way the `ContractsDevProvider` will have the test user API too

Thoughts?

kratico:
@harrysolovay @tjjfvi I just realized that the `ProjectProvider` and the `ZombienetProvider` also extends from `FrameBinProvider`

So adding the overrides to `.chainFile` and `.handle` will work only for substrate based nodes because the assumption is that using the binary the chain spec could be generated with `binary build-spec ...`.

Maybe I need to change the design by extracting the user testing stuff in a new class and then use composition instead of inheritance.

Thoughts?

tjjfvi:
Perhaps we should just merge the dev and contracts providers

providers/frame/common.ts
```diff
@@ -0,0 +1,120 @@
...
+  const buildSpecCmd = new Deno.Command(bin, {
+    args: [
+      "build-spec",
+      "--disable-default-bootnode",
+      "--chain",
+      chain,
+    ],
+  })
+  const chainSpec = JSON.parse(new TextDecoder().decode((await buildSpecCmd.output()).stdout))
+  const balances: any[] = chainSpec.genesis.runtime.balances.balances
```
harrysolovay, line 22:
```suggestion
  const balances: [string, number][] = chainSpec.genesis.runtime.balances.balances
```

providers/frame/common.ts
```diff
@@ -0,0 +1,120 @@
...
+}
+
+export async function chainFileWithUsers(
+  file: File,
+  url: string,
+): Promise<File> {
+  return new File(`
+    ${file.codeRaw}
+
+    type ArrayOfLength<
```
harrysolovay, line 61: Should we expose this from `util` and then import it from this generated file?

providers/frame/common.ts
```diff
@@ -0,0 +1,120 @@
...
+      T,
+      L extends number,
+      A extends T[] = [],
+    > = number extends L ? T[]
+      : L extends A["length"] ? A
+      : ArrayOfLength<T, L, [...A, T]>
+
+    export async function users<N extends number>(count: N): Promise<ArrayOfLength<C.Sr25519, N>>
+    export async function users(count: number): Promise<C.Sr25519[]> {
```
harrysolovay, line 70: Let's extract this into a `userFactory` fn, exposed from within `test_pairs`. `userFactory` can accept the `url` as an argument and return the `users` fn. This way, we generate less code:

```ts
export const users = C.testUserFactory(url)
```

harrysolovay (changes requested):

util/ArrayOfLength.ts
```diff
@@ -0,0 +1,7 @@
+export type ArrayOfLength<
+  T,
+  L extends number,
+  A extends T[] = [],
+> = number extends L ? T[]
+  : L extends A["length"] ? A
+  : ArrayOfLength<T, L, [...A, T]>
```
kratico, line 7: @harrysolovay Would this be the right place for this utility type?

harrysolovay: Sure :)

crypto/test_pairs.ts
```diff
@@ -30,3 +32,36 @@ export const bobStash = pair(
...
+export function testUserFactory(url: string) {
+  return async function users<N extends number>(count: N): Promise<ArrayOfLength<Sr25519, N>> {
+    const response = await fetch(
+      url,
+      {
+        method: "POST",
+        headers: {
+          "content-type": "application/json",
+        },
+        body: JSON.stringify({ count }),
```
harrysolovay, line 51: Why does the count need to be contained within an object? Wanna just have the count be the body?

crypto/test_pairs.ts
```diff
@@ -30,3 +32,36 @@ export const bobStash = pair(
...
+      },
+    )
+    if (!response.ok) {
+      throw new Error(await response.text())
+    }
+    const { index }: { index: number } = await response.json()
```
harrysolovay, line 57: Same question for this

crypto/test_pairs.ts
kratico: If we eventually need to add more data, I think that it's easier to augment the body with an extra attribute.
The above assumes that the CAPI server could eventually be deployed as a cloud service.

If that assumption is not accurate, then, this and the response could be raw text.

Shall I do the switch to raw text?

kratico: for a backward compatible change.
For example, the client code is cached by the deno cache and the server is updated keeping the old object attribute and adding a new one

tjjfvi (dismissed):

providers/frame/common.ts
```diff
@@ -0,0 +1,61 @@
...
+}
+
+export async function handleCount(request: Request, cache: { count: number }): Promise<Response> {
+  const body = await request.json()
+  $.assert($.field("count", $.u32), body)
+  const { count } = body
+  let index = cache.count
+  const newCount = index + count
+  if (newCount < DEFAULT_TEST_USER_COUNT) cache.count = newCount
+  else index = -1
```
harrysolovay, line 56: Last thing: per @tjjfvi's recommendation, let's actually return an error response here. Then we can delete the `if (index === -1)` check in the client `users` fn.

harrysolovay (approved):

PR #731, opened 2023-03-10
Feat/example tool testing vscode less docker

kratico:


kratico:
cc @ryanleecode 
commenting `vscode` related things in the `Dockerfile` but ended up with the same error as before
https://github.com/paritytech/capi/actions/runs/4384799803/jobs/7676708318#step:3:13717

PR #740, opened 2023-03-13
fix: xcm_asset_teleportation not being ignored

ryanleecode:


tjjfvi (approved):
I don't believe this ignore is doing anything atm but I suppose there's no harm in fixing it.

ryanleecode:
> I don't believe this ignore is doing anything atm but I suppose there's no harm in fixing it. 

im using it in my tool

PR #743, opened 2023-03-14
chore: wrap error unhandling example in assertRejects

ryanleecode:


examples/error_unhandling.ts
```diff
@@ -1,14 +1,16 @@
...
+    .transfer({
+      value: 1_000_000_000_000_000_000_000_000_000_000_000_000n,
+      dest: bob.address,
+    })
+    .signed({ sender: alice })
+    .sent()
+    .dbgStatus()
+    .finalizedEvents()
+    .unhandleFailed()
+    .run(), "unhandleFailed")
```
tjjfvi, line 16: I don't think the `"unhandleFailed"` here is right.

tjjfvi:
```suggestion
    .run())
```

ryanleecode: mmm gimme a suggestion. not sure how these should be worded. "should undandleFailed` ?

ryanleecode: oh just remove it ok

tjjfvi: Ohh, sorry, I misunderstood that parameter. In that case, perhaps `"Should fail due to `InsufficientBalance`"`?

tjjfvi (approved):

ryanleecode:
@tjjfvi override merge plz

PR #744, opened 2023-03-15
chore: lint examples ignore file

ryanleecode:
Check the `.ignore` file in the `examples` directory to make sure that it doesn't specify any files that don't exist.


harrysolovay:
This check could be part of the example-testing tool that you're building (at which point I think this would make more sense as an example run preflight).

PR #751, opened 2023-03-15
feat: improve ss58 encoding/decoding

### LINKED ISSUE #619: add `multiAddressToSs58`
Perhaps we can learn something from [this](https://github.com/airgap-it/airgap-coin-lib/blob/44a9acec71418c9f1bb253a9189e8833e2657317/packages/core/src/protocols/substrate/common/data/scale/type/SCALEMultiAddress.ts)?

kratico:
Resolves #619 

Add ss58 encoding and decoding for different payload and checksum lengths (see [ss58-format](https://docs.substrate.io/reference/address-formats/)).

Note: `MultiAdress.Index` use formats whose checksum length is `1` provided that the index is of type `u8`, `u16`, `u32` or `u64` (`1`, `2`, `4` or `8` byte payload).

image

### DIFF
--- a/crypto/ss58.ts
+++ b/crypto/ss58.ts
@@ -19,10 +19,9 @@ export function encodeRaw(
   payload: Uint8Array,
   props?: EncodeProps,
 ): Uint8Array {
-  const checksumLength = props?.checksumLength ?? VALID_PAYLOAD_CHECKSUM_LENGTHS[payload.length]
+  const checksumLength = props?.checksumLength ?? DEFAULT_PAYLOAD_CHECKSUM_LENGTHS[payload.length]
   if (!checksumLength) throw new InvalidPayloadLengthError()
-  const isValidNetworkPrefix = !props?.validNetworkPrefixes
-    || props.validNetworkPrefixes.includes(prefix)
+  const isValidNetworkPrefix = props?.validNetworkPrefixes?.includes(prefix) ?? true
   if (!isValidNetworkPrefix) throw new InvalidNetworkPrefixError()
   const prefixBytes = prefix < 64
     ? Uint8Array.of(prefix)
@@ -109,7 +108,7 @@ const VALID_ADDRESS_CHECKSUM_LENGTHS: Record<number, number | undefined> = {
   36: 2,
   37: 2,
 }
-const VALID_PAYLOAD_CHECKSUM_LENGTHS: Record<number, number | undefined> = {
+const DEFAULT_PAYLOAD_CHECKSUM_LENGTHS: Record<number, number | undefined> = {
   1: 1,
   2: 1,
   4: 1,
--- a/crypto/ss58.ts
+++ b/crypto/ss58.ts
@@ -1,22 +1,28 @@
 import * as base58 from "../deps/std/encoding/base58.ts"
 import { Blake2b } from "../deps/wat_the_crypto.ts"
 
+export interface EncodeProps {
+  validNetworkPrefixes?: readonly number[]
+  checksumLength?: number
+}
+
 export function encode(
   prefix: number,
   payload: Uint8Array,
-  validNetworkPrefixes?: readonly number[],
+  props?: EncodeProps,
 ): string {
-  return base58.encode(encodeRaw(prefix, payload, validNetworkPrefixes))
+  return base58.encode(encodeRaw(prefix, payload, props))
 }
 
 export function encodeRaw(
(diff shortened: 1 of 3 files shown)


crypto/ss58.ts
```diff
@@ -99,14 +83,44 @@ export class InvalidAddressChecksumError extends Error {
...
+  36: 2,
+  37: 2,
+}
+const VALID_PAYLOAD_CHECKSUM_LENGTHS: Record<number, number | undefined> = {
+  1: 1,
+  2: 1,
+  4: 1,
+  8: 1,
+  32: 2,
+  33: 2,
```
kratico, line 112: These libs support `33 byte` payloads with `2 byte` checksum

- https://github.com/polkascan/py-scale-codec/blob/master/scalecodec/utils/ss58.py#L121
- https://github.com/wyhaines/base58.cr/blob/main/src/base58/ss58.cr#L284

tjjfvi:
> Note: MultiAdress.Index use formats whose checksum length is 1 provided that the index is of type `u8`, `u16`, `u32` or `u64` (`1`, `2`, `3` or `4` byte payload).

It would be `1`/`2`/`4`/`8` byte payload, no?

kratico:
> > Note: MultiAdress.Index use formats whose checksum length is 1 provided that the index is of type `u8`, `u16`, `u32` or `u64` (`1`, `2`, `3` or `4` byte payload).
> 
> It would be `1`/`2`/`4`/`8` byte payload, no?

Yes!

tjjfvi (commented):
Could we get a test for every valid address length and every valid payload length to make sure they always round-trip correctly?

kratico:
> Could we get a test for every valid address length and every valid payload length to make sure they always round-trip correctly?

I could add decoding tests for the 16 formats.
But for encoding, I will need to change the `encode(...)` signature to accept the desired checksum or I can leave it as is.
@tjjfvi  WDYT?

crypto/ss58.ts
```diff
@@ -99,14 +89,44 @@ export class InvalidAddressChecksumError extends Error {
...
+  13: 4,
+  14: 5,
+  15: 6,
+  16: 7,
+  17: 8,
+  35: 2,
+  36: 2,
+  37: 2,
+}
+const VALID_PAYLOAD_CHECKSUM_LENGTHS: Record<number, number | undefined> = {
```
tjjfvi, line 112:
```suggestion
const DEFAULT_PAYLOAD_CHECKSUM_LENGTHS: Record<number, number | undefined> = {
```

crypto/ss58.ts
```diff
@@ -1,54 +1,52 @@
...
+  props?: EncodeProps,
 ): Uint8Array {
-  const isValidPublicKeyLength = !!VALID_PUBLIC_KEY_LENGTHS[pubKey.length]
-  if (!isValidPublicKeyLength) throw new InvalidPublicKeyLengthError()
-
-  const isValidNetworkPrefix = !validNetworkPrefixes || validNetworkPrefixes.includes(prefix)
+  const checksumLength = props?.checksumLength ?? VALID_PAYLOAD_CHECKSUM_LENGTHS[payload.length]
+  if (!checksumLength) throw new InvalidPayloadLengthError()
+  const isValidNetworkPrefix = !props?.validNetworkPrefixes
+    || props.validNetworkPrefixes.includes(prefix)
```
tjjfvi, line 25:
```suggestion
  const isValidNetworkPrefix = props?.validNetworkPrefixes?.includes(prefix) ?? false
```

crypto/ss58.ts
```diff
@@ -1,54 +1,52 @@
 import * as base58 from "../deps/std/encoding/base58.ts"
 import { Blake2b } from "../deps/wat_the_crypto.ts"
 
+export interface EncodeProps {
+  validNetworkPrefixes?: readonly number[]
```
tjjfvi, line 5: Why is this parameter necessary?

kratico: I think we wanted to let the user specify`validNetworkPrefixes` if any.
@harrysolovay Do you remember?

tjjfvi: In that case, couldn't the user just validate the network prefix before passing it to encode?

kratico: makes sense, I've removed the `validNetworkPrefixes`

tjjfvi (approved):

PR #752, opened 2023-03-15
chore: make use of users util in examples

harrysolovay:
Thanks to #725


kratico (dismissed):
🎉 LGTM

tjjfvi (commented):
Given that `alice`/`bob`/`charlie`/`dave`/`eve` have specific meaning in the polkadot context, perhaps for these test users we should adopt a different set of names. Maybe
```
alexa, billy, carol, david, emily
```
... or just `a, b, c, d, e`, but that seems less fun.

tjjfvi (approved):

Issue #709, opened 2023-03-06
pre-package capi dependencies


harrysolovay:
Create a repo that packages all of the Capi+provider deps (polkadot, cumulus, zombienet, misc.) for different systems, for more seamless downloads / auto-installations. #651

tjjfvi:
[`paritytech/capi-binary-builds`](https://github.com/paritytech/capi-binary-builds)

PR #753, opened 2023-03-15
feat: pre-compute test users

### LINKED ISSUE #741: pre-compute test users
See https://github.com/paritytech/capi/pull/641#discussion_r1135484606

kratico:
Resolves #741 

The `test_users_public_keys` file was generated with

```typescript
import { testUser } from "./crypto/mod.ts"
import * as $ from "./deps/scale.ts"

const DEFAULT_TEST_USER_COUNT = 100_000

const publicKeys = []
for (let i = 0; i < DEFAULT_TEST_USER_COUNT; i++) {
  publicKeys.push(testUser(i).publicKey)
}

await Deno.writeFile(
  `./providers/frame/test_users_public_keys`,
  $.array($.sizedUint8Array(32)).encode(publicKeys),
)

```

Note that the `star.yml` workflow duration is not impacted by this change

image


providers/frame/common.ts
```diff
@@ -14,11 +14,21 @@ export async function createCustomChainSpec(
...
-    balances.push([
-      ss58.encode(networkPrefix, testUser(i).publicKey),
-      DEFAULT_TEST_USER_INITIAL_FUNDS,
-    ])
+  try {
+    balances.push(
+      ...JSON.parse(
+        await Deno.readTextFile(
+          new URL(`./test_users/network_${networkPrefix}_user_balances.json`, import.meta.url),
+        ),
```
kratico, line 22: Each `network_<prefix>_user_balances.json` weights ~7mb.
I'm not sure what would be best
- keep them in Git
- upload them the something like S3 and cache them in a CDN

WDYT?

tjjfvi: Hmm, I didn't think about how large the files would be... Can you test to see how much time each part of the generation takes? (text encoding / hashing / sr25519 from seed / ss58 generation) If the text encoding and/or hashing are expensive, we could adopt a different approach for the seed generation, but if the majority of the time is the sr25519 or ss58, we might be out of luck.

tjjfvi:
And yeah, star is taking a while to cache... I think maybe the `Cache deno dependencies` step isn't doing as much as it should. I suspect given that the polkadot js dependencies in `examples/polkadot_js_signer.ts` aren't in `deps`, that they weren't included in the initial cache cache.

tjjfvi:
I just evicted the deno cache for main, we'll see if that fixes it

tjjfvi:
Ah, no, I think it's taking a long time again because @harrysolovay [changed the specifiers back to `@latest`](https://github.com/paritytech/capi/commit/ff0b79f450fcee461f8c438a70ab1a165a31b72f#diff-92d6086f7630dd58a9bf2713de3849ccde8ce4bf11d6cce8278f34c071bcad82), so now star has to spin up chains again.

harrysolovay:
> Ah, no, I think it's taking a long time again because @harrysolovay [changed the specifiers back to `@latest`](https://github.com/paritytech/capi/commit/ff0b79f450fcee461f8c438a70ab1a165a31b72f#diff-92d6086f7630dd58a9bf2713de3849ccde8ce4bf11d6cce8278f34c071bcad82), so now star has to spin up chains again.

Correct: I didn't realize the pinning was intentional / for caching purposes. Apologies. Feel free to revert.

providers/frame/common.ts
tjjfvi: Ah, ok. In that case, instead of pre-computing the full addresses, could you just pre-compute the public keys, and then save them into a file using `$.array($.sizedUint8Array(32))`? That way it can be the same file across chains, and it should be only around 3.2 MB.

tjjfvi: And if the file is only around 3.2 megs, you can put that in git for now, like we do with the dprint formatter. Later, once I take a look at #709, that file can potentially be moved from git into an s3 cache.

cspell.json
```diff
@@ -18,6 +18,6 @@
     "**/__snapshots__/*.snap",
     "**/*.contract",
     "examples/smart_contract/metadata.json",
-    "providers/frame/test_users/*.json"
+    "providers/frame/test_users_public_keys"
```
tjjfvi, line 21: Can you rename this file to use a `.scale` extension?
```suggestion
    "**/*.scale"
```

tjjfvi (changes requested):
Looks good. Can you put the code you used to generate that file into `_tasks/download.ts`? (at which point it can be renamed to something like `generate_artifacts.ts`)

tjjfvi (dismissed):

providers/frame/common.ts
```diff
@@ -1,8 +1,9 @@
 import { ss58 } from "../../crypto/mod.ts"
 import * as $ from "../../deps/scale.ts"
-import { DEFAULT_TEST_USER_COUNT } from "../../util/generateTestUserPublicKeys.ts"
 
+export const DEFAULT_TEST_USER_COUNT = 100_000
 const DEFAULT_TEST_USER_INITIAL_FUNDS = 1_000_000_000_000_000_000
+export const publicKeysPath = new URL("./test_users_public_keys.scale", import.meta.url)
```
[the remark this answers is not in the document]
tjjfvi: Yeah, you'll need to use `fetch` for the reading (for when capi is loaded from `deno.land`, not locally), and use `path.fromFileUrl` for the writing.

providers/frame/common.ts
```diff
@@ -3,7 +3,7 @@ import * as $ from "../../deps/scale.ts"
 
 export const DEFAULT_TEST_USER_COUNT = 100_000
 const DEFAULT_TEST_USER_INITIAL_FUNDS = 1_000_000_000_000_000_000
-export const publicKeysPath = new URL("./test_users_public_keys.scale", import.meta.url)
+export const publicKeysFileUrl = import.meta.resolve("./test_users_public_keys.scale")
```
tjjfvi, line 6: It's not always a file url:```suggestion
export const publicKeysUrl = import.meta.resolve("./test_users_public_keys.scale")
```

tjjfvi (approved):

PR #757, opened 2023-03-17
feat: `from` method on all rune classes

harrysolovay:
Currently, we instantiate Rune subclasses from constants as follows.

```ts
const multisig = Rune
  .constant({
    signatories: [alexa, billy, carol]
      .map(({ publicKey }) => publicKey),
    threshold: 2,
  })
  .into(MultisigRune, chain)
```

Following this PR we can express this as...

```ts
const multisig = MultisigRune.from({
  signatories: [alexa, billy, carol]
    .map(({ publicKey }) => publicKey),
  threshold: 2,
}, chain)
```

### DIFF
--- a/patterns/multisig/MultisigRune.ts
+++ b/patterns/multisig/MultisigRune.ts
@@ -3,6 +3,7 @@ import { MultiAddress } from "polkadot/types/sp_runtime/multiaddress.js"
 // const MultiAddress = null! as any
 import * as bytes from "../../deps/std/bytes.ts"
 import { Chain, ChainRune, ExtrinsicRune, Rune, RunicArgs, ValueRune } from "../../mod.ts"
+import { PolkadotSignatureChain } from "../signature/polkadot.ts"
 import { multisigAccountId } from "./multisigAccountId.ts"
 
 export interface MultisigRatifyProps<C extends Chain> {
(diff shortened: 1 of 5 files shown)


patterns/multisig/MultisigRune.ts
```diff
@@ -21,7 +22,7 @@ export interface Multisig {
 }
 
 // TODO: swap out `Chain` constraints upon subset gen issue resolution... same for other patterns
-export class MultisigRune<out C extends Chain, out U> extends Rune<Multisig, U> {
+export class MultisigRune<out C extends PolkadotSignatureChain, out U> extends Rune<Multisig, U> {
```
harrysolovay, line 25: This is a temporary suppression of the resulting type error (in lieu of proper constraints).

patterns/multisig/VirtualMultisigRune.ts
```diff
@@ -28,7 +28,9 @@ export const $virtualMultisig: $.Codec<VirtualMultisig> = $.object(
   $.field("stash", $.sizedUint8Array(32)),
 )
 
-export class VirtualMultisigRune<out C extends Chain, out U> extends Rune<VirtualMultisig, U> {
+export class VirtualMultisigRune<out C extends PolkadotSignatureChain, out U>
```
harrysolovay, line 31: Same thing here.

patterns/multisig/VirtualMultisigRune.ts
```diff
@@ -114,7 +116,7 @@ export class VirtualMultisigRune<out C extends Chain, out U> extends Rune<Virtua
       .into(VirtualMultisigRune, chain)
   }
 
-  static deployment<C extends Chain, U, X>(
+  static deployment<C extends PolkadotSignatureChain, U, X>(
```
harrysolovay, line 119: And here.

tjjfvi:
Discussed offline: the typing of this won't work as we hoped it would; we'll leave it to pattern libraries to write their own analogs (including variants like `fromHex` etc.)

PR #765, opened 2023-03-20
feat: add reserve transfer example

kratico:
Add [reserve transfer asset](https://wiki.polkadot.network/docs/learn-xcm-usecases#reserve-asset-transfer) example between Statemine and Trappist.

Notes
- for this example, Statemine acts as the source and reserve chain.
- the example uses [trappist#22755b2](https://github.com/paritytech/trappist/commit/22755b22cb13ff562df50e73872f444066dba3fd) which is compatible with `polkadot v0.9.37` and `polkadot-parachain v0.9.370`

The example steps are
- In Rococo, create a sufficient asset on Statemine (reserve parachain)
  Notes
  - asset should be sufficient to be able to pay fees in Statemine
  - root is needed to create a sufficient asset
  - in a common good parachains, root is accessed through the relay chain
- In Statemine, mint the new reserve asset
- In Trappist, create an asset to pair it with the Statemeine reserve asset 
- In Trappist, register Trappist AssetId to Statemine AssetId
- In Statemine, make a reserve transfer asset to Trappist


kratico:
@harrysolovay @tjjfvi I need to setup docker to debug the GH action issue
In the meantime, feedback is welcome

tjjfvi:
Let's wait to merge this until after #823

tjjfvi:
@kratico Devnets has been merged, but it'll need to be modified to set up the hrmp channels.

devnets/startNetwork.ts
```diff
@@ -177,7 +181,7 @@ async function spawnChain(
     if (bootnodes) {
       args.push("--bootnodes", bootnodes)
     } else {
-      const { nodeKey, peerId } = await generateNodeKey(binary)
+      const { nodeKey, peerId } = await generateNodeKey(generateNodeKeyBinary)
```
kratico, line 184: The `trappist` binary does not have the utility to generate the node key.

Ideally, we should figure out the algo to generate the node key and not use a binary to compute it.

.trunignore
```diff
@@ -1,3 +1,3 @@
-examples/xcm/asset_teleportation.eg.ts
+examples/xcm/*.eg.ts
```
tjjfvi, line 1: Do these error currently? Or are they just slow?

devnets/startNetwork.ts
```diff
@@ -216,3 +220,15 @@ async function spawnNode(tempDir: string, binary: string, args: string[], signal
...
 }
+
+function addXcmHrmpChannels(genesisConfig: GenesisConfig, paraIds: number[]) {
+  if (!genesisConfig.hrmp) {
+    genesisConfig.hrmp = { preopenHrmpChannels: [] }
+  }
+  for (const senderParaId of paraIds) {
+    for (const recipientParaId of paraIds) {
+      if (senderParaId === recipientParaId) continue
+      genesisConfig.hrmp.preopenHrmpChannels.push([senderParaId, recipientParaId, 8, 512])
```
tjjfvi, line 231: Can the 8 and 512 constants be extracted? (Should they be configurable?)

devnets/startNetwork.ts
```diff
@@ -216,3 +220,15 @@ async function spawnNode(tempDir: string, binary: string, args: string[], signal
     }
   })
 }
+
+function addXcmHrmpChannels(genesisConfig: GenesisConfig, paraIds: number[]) {
+  if (!genesisConfig.hrmp) {
+    genesisConfig.hrmp = { preopenHrmpChannels: [] }
+  }
```
tjjfvi, line 227:
```suggestion
 genesisConfig.hrmp ??= { preopenHrmpChannels: [] }
```

devnets/startNetwork.ts
kratico: I'll update the function signature and set these values as default.

Given that these values can be different for each permutation, the alternative would be to code each permutation in the `capi.config.ts`

@tjjfvi  WDYT?

.trunignore
harrysolovay: Let's ignore the XCM examples for now. We can re-enable upon configurability of block times.

devnets/startNetwork.ts
harrysolovay: What resulting behavior does this impact? Is it meaningful to JS devs?

kratico: I guess that we could see an impact if there are many messages, but given that this is for dev, I think we should be fine with these values

words.txt
```diff
@@ -84,3 +89,5 @@ unhandle
 unioned
 vadimcn
 westend
+xcmp
+zombienet
```
harrysolovay, line 93:
```suggestion
```

examples/xcm/reserve_transfer.eg.ts
```diff
@@ -0,0 +1,223 @@
...
+ * @description Perform an XCM reserve asset transfer, in which two chains, which
+ * do not trust one another, rely on a third chain to store assets and facilitate
+ * the exchange.
+ */
+
+import * as Rococo from "@capi/rococo-dev-xcm"
+import * as Statemine from "@capi/rococo-dev-xcm/statemine"
+import * as Trappist from "@capi/rococo-dev-xcm/trappist"
+import { assert, assertNotEquals } from "asserts"
+import { $, alice as root, hex, Rune } from "capi"
```
kratico, line 13: @harrysolovay for `sudo` calls I switched `alexa` to `alice (or root)` because the default sudo key for these runtimes is `alice` key

harrysolovay (dismissed):
LGTM

capi.config.ts
```diff
@@ -58,5 +58,21 @@ export const config: CapiConfig = {
         },
       },
     },
+    rococoDevXcm: {
```
tjjfvi, line 61: Can this network be merged with the above once #888 is merged?

examples/xcm/reserve_transfer.eg.ts
```diff
@@ -0,0 +1,223 @@
...
+console.log("Statemine(Billy): asset balance:", billyStatemintBalanceFinal)
+assertNotEquals(billyStatemintBalanceInitial, billyStatemintBalanceFinal)
+
+// Retrieve the statemint sovereign account balance.
+const statemintSovereignAccountBalance = await Statemine.Assets.Account
+  // Sovereign address on sibling chain
+  // b"sibl" + $.u32.encode(2000) + 0...0
+  .value([
+    RESERVE_ASSET_ID,
+    hex.decode("0x7369626cd0070000000000000000000000000000000000000000000000000000"),
```
tjjfvi, line 215: Can you make a codec for this and use that instead?

capi.config.ts
kratico: It can't be merged because `trappist` hasn't been updated to XCM v3.
Therefore this example needs a network with `polkadot v0.9.37`, `polkadot v0.9.370` and `trappist#...`

examples/xcm/reserve_transfer.eg.ts
kratico: Would something like `$sovereignAccount.encode(["sibl", 2000])` work?

kratico: @tjjfvi Where shall I leave this codec?

harrysolovay: Depends, is this sovereign account constant specific to this setup? If so, `patterns/constants.ts`?

kratico: Adding more info, parachain sovereign account addresses can be prefixed with the binary representation of `sibl` or `para`

So this would be a codec for parachain addresses.

There are some details on how to compute the address in https://substrate.stackexchange.com/questions/1200/how-to-calculate-sovereignaccount-for-parachain/1210

harrysolovay: In that case, how about a `patterns/para_id.ts`, which contains two codecs: `$siblId` and `$paraId`?

examples/xcm/reserve_transfer.eg.ts
```diff
@@ -0,0 +1,224 @@
...
+// Retrieve Billy's balance on statemint.
+const billyStatemintBalanceFinal = await billyStatemintBalance.run()
+
+// Ensure the balance is different from the initial.
+console.log("Statemine(Billy): asset balance:", billyStatemintBalanceFinal)
+assertNotEquals(billyStatemintBalanceInitial, billyStatemintBalanceFinal)
+
+// Retrieve the statemint sovereign account balance.
+const statemintSovereignAccountBalance = await Statemine.Assets.Account
+  // Sovereign address on sibling chain
```
harrysolovay, line 212: We can probably get rid of these two lines of comments. As well as have `214-217` on a single line.

harrysolovay: ... unless, do you think it's important to describe this codec usage?

harrysolovay (dismissed):

examples/xcm/reserve_transfer.eg.ts
kratico: you are right, these comments are not necessary

examples/xcm/reserve_transfer.eg.ts
```diff
@@ -0,0 +1,219 @@
...
+  maxTimeout: 2 * 60 * 1000,
+}
+
+// Create a sufficient asset with Sudo. When targeting a common good
+// parachain, access root instead through the relay chain.
+await Rococo.Sudo
+  .sudo({
+    call: Rococo.ParasSudoWrapper.sudoQueueDownwardXcm({
+      id: RESERVE_CHAIN_ID,
+      xcm: Rune.rec({
```
harrysolovay, line 40: We can use the codegened type factory here.

examples/xcm/reserve_transfer.eg.ts
```diff
@@ -0,0 +1,219 @@
...
+        type: "V2",
+        value: Rune.tuple([
```
harrysolovay, line 42: And here

examples/xcm/reserve_transfer.eg.ts
```diff
@@ -0,0 +1,219 @@
...
+          Rococo.Instruction.Transact({
+            originType: "Superuser",
+            requireWeightAtMost: 1000000000n,
+            call: Rune.rec({
```
harrysolovay, line 46: And here

examples/xcm/reserve_transfer.eg.ts
```diff
@@ -0,0 +1,219 @@
...
+  .dbgStatus("Trappist(root): Create derived asset")
+  .finalized()
+  .run()
+
+// Register Trappist parachain asset id to reserve asset id.
+await Trappist.Sudo
+  .sudo({
+    call: Trappist.AssetRegistry.registerReserveAsset({
+      assetId: TRAPPIST_ASSET_ID,
+      assetMultiLocation: Rune.rec({
```
harrysolovay, line 118: And here

examples/xcm/reserve_transfer.eg.ts
```diff
@@ -0,0 +1,219 @@
...
+    Junctions,
+    AssetId,
+    NetworkId,
+    WeightLimit,
+    RuntimeEvent,
+    CumulusPalletXcmpQueueEvent: { isXcmpMessageSent },
+  } = Statemine
+  const events = await Statemine.PolkadotXcm
+    .limitedReserveTransferAssets({
+      dest: VersionedMultiLocation.V1(Rune.rec({
```
harrysolovay, line 151: Here

examples/xcm/reserve_transfer.eg.ts
```diff
@@ -0,0 +1,219 @@
...
+        parents: 1,
+        interior: Junctions.X1(
+          XcmV1Junction.Parachain(TRAPPIST_CHAIN_ID),
+        ),
+      })),
+      beneficiary: VersionedMultiLocation.V1(Rune.rec({
```
harrysolovay, line 157: Here

examples/xcm/reserve_transfer.eg.ts
```diff
@@ -0,0 +1,219 @@
...
+        parents: 0,
+        interior: Junctions.X1(
+          XcmV1Junction.AccountId32({
+            network: NetworkId.Any(),
+            id: billy.publicKey,
+          }),
+        ),
+      })),
+      assets: VersionedMultiAssets.V1(Rune.array([Rune.rec({
```
harrysolovay, line 166: Here

examples/xcm/reserve_transfer.eg.ts
```diff
@@ -0,0 +1,219 @@
...
+        id: AssetId.Concrete(Rune.rec({
```
harrysolovay, line 167: Here

devnets/startNetwork.ts
```diff
@@ -288,3 +292,23 @@ function addAuthorities(genesisConfig: GenesisConfig, count: number) {
...
+const hrmpChannelMaxMessageSize = 512
+function addXcmHrmpChannels(
+  genesisConfig: GenesisConfig,
+  paraIds: number[],
+) {
+  genesisConfig.hrmp ??= { preopenHrmpChannels: [] }
+  for (const senderParaId of paraIds) {
+    for (const recipientParaId of paraIds) {
+      if (senderParaId === recipientParaId) continue
+      genesisConfig.hrmp.preopenHrmpChannels.push([
```
tjjfvi, line 306: Do hrmp channels need to be opened for both directions? I.e. 1 -> 2 and 2 -> 1?

fluent/ExtrinsicRune.ts
```diff
@@ -31,18 +31,20 @@ export type SignatureDataFactory<C extends Chain, CU, SU> = (
...
   static readonly PROTOCOL_VERSION = 4
 
-  hash = this.chain
-    .into(ValueRune)
-    .access("metadata", "extrinsic", "call")
-    .map((x) => blake2_256.$hash<any>(x))
-    .into(CodecRune)
-    .encoded(this)
+  $call = this.chain.into(ValueRune).access("metadata", "extrinsic", "call").into(CodecRune)
+  call = this.$call.encoded(this)
```
tjjfvi, line 35:
```suggestion
  callData = this.$call.encoded(this)
```

patterns/para_id.ts
```diff
@@ -0,0 +1,26 @@
...
+      const prefix_ = new TextDecoder().decode($.sizedUint8Array(4)._decode(buffer))
+      const paraId = $.u32._decode(buffer)
+      buffer.index += 24
+      if (prefix != prefix_) throw new Error("Invalid prefix")
+      return paraId
+    },
+    _assert(assert) {
+      assert.typeof(this, "number")
+    },
+  })
```
tjjfvi, line 25:
```suggestion
  const $prefix = $.constant(null, new TextEncoder().encode(prefix))
  const $postfix = $.constant(null, new Uint8Array(24))
  return $.createCodec<number>({
    _metadata: $.metadata(`$${prefix}Id`),
    _staticSize: $prefix._staticSize + $.u32._staticSize + $postfix._staticSize,
    _encode(buffer, value) {
      $prefix._encode(buffer, null)
      $.u32.encode(buffer, value)
      $postfix._encode(buffer, null)
    },
    _decode(buffer) {
      $prefix._decode(buffer)
      const paraId = $.u32._decode(buffer)
      $postfix._decode(buffer)
      return paraId
    },
    _assert(assert) {
      $.u32._assert(assert)
    },
  })
}
```

patterns/signature/statemint.ts
tjjfvi, line 1: Now that #815 is merged I believe a copy of this file exists already?

devnets/startNetwork.ts
kratico: yes

patterns/signature/statemint.ts
kratico: There is [`westmint.ts`](https://github.com/paritytech/capi/blob/d875ffa0c1a69051d10292a959ac08eaa67fd0ef/patterns/signature/westmint.ts)

Nevertheless, I will update to use `SignatureProps` from `patterns/signature/polkadot.ts`

tjjfvi: Don't westmint and statemine have the same signature types? If so, we only need one file. Probably best named statemine than westmint, though.

kratico, line 1: @tjjfvi @harrysolovay `westmint.ts` has been tweaked so it can be used for statemint/statemine too.

What would be a better filename?

harrysolovay: `statemint` should be good 👍

tjjfvi (approved):

PR #768, opened 2023-03-21
fix: entry page and key page usage

harrysolovay:
Fixes storage entry and key page retrieval

### DIFF
--- a/fluent/StorageRune.ts
+++ b/fluent/StorageRune.ts
@@ -31,7 +31,7 @@ export class StorageRune<
       blockHash?: string,
     ]>
   ) {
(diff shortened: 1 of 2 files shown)


fluent/StorageRune.ts
```diff
@@ -31,7 +31,7 @@ export class StorageRune<
       blockHash?: string,
     ]>
   ) {
-    const storageKey = this.$key.encoded(key).map(hex.encode)
+    const storageKey = this.$partialKey.encoded(key).map(hex.encode)
```
tjjfvi, line 34: This is a full key, not a partial key
```suggestion
    const storageKey = this.$key.encoded(key).map(hex.encode)
```

PR #769, opened 2023-03-21
feat: trun example tooling

ryanleecode:
## Summary

"Trun" is an tool to test our examples both in Deno and the Browser. It does this by running as a github action/workflow inside of a particular repo -- in this case capi. For running examples in Deno, it spawns separate `Deno.Commands`, collects the console output and aggregates the exitcodes to produce a test summary. It runs in the browser by compiling each example with `esbuild` and running a puppeteer chromium instance with the compiled script injected. 

Trun will eventually be migrated to https://github.com/paritytech/trun but to start off it will live inside of the capi repo.

__Note: the tests will fail since not all the examples are working... but the workflow isnt required for merge so this isn't an issue.__

## Capi Specific Changes

- Updates dev container so it can run puppetter headlessly
- Removes old test_examples script
- Adds two new workflows that run on trun -- one for the browser and one for deno


.github/actions/trun/main.ts
```diff
@@ -0,0 +1,90 @@
+import * as core from "./deps/actions/core.ts"
+import { PQueue } from "./deps/pqueue.ts"
+import puppeteer from "./deps/puppeteer.ts"
+import { deferred } from "./deps/std/async.ts"
+import { parse } from "./deps/std/flags.ts"
+import * as path from "./deps/std/path.ts"
+import { run, runWithBrowser, runWithDeno } from "./run.ts"
+
+const flags = parse(Deno.args, {
+  string: ["dir", "concurrency", "ignore", "importMap", "filter"],
```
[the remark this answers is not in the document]
harrysolovay: Is `dir` necessary? Why not assume cwd?

.github/actions/trun/main.ts
```diff
@@ -0,0 +1,90 @@
...
+const results: [fileName: string, exitCode: number][] = []
+const runner = useBrowser
+  ? await runWithBrowser({ createBrowser, importMapURL, logger, results })
+  : await runWithDeno({ logger, reloadUrl: "http://localhost:4646", results })
+const paths = sourceFileNames.map((fileName) => [dir, fileName] as const)
+
+await run({ paths, runner, concurrency })
+console.log("test results", results)
+
+const isFailed = results.find(([_, exitCode]) => exitCode === 1)
```
harrysolovay, line 74: Would `!!exitCode` suffice?

_tasks/validate_ignore.ts
harrysolovay, line 1: Does this merit a task unto itself? Why not place within trun as a preflight step?

.github/actions/trun/main.ts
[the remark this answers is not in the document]
ryanleecode: The tool runs from the root of the capi repo unless the expectation is for it to `cd` into the examples directory

.github/actions/trun/main.ts
ryanleecode: it could but is that really better than an equals operator?

.github/actions/trun/run.ts
```diff
@@ -0,0 +1,132 @@
...
+      entryPoints: [path.join(dir, fileName)],
+      bundle: true,
+      write: false,
+      format: "esm",
+    })
+
+    const code = wrapCode(`${consoleJS}\n${result.outputFiles[0]?.text!}`)
+
+    await page.exposeFunction("__trun_injected_log", (...args: unknown[]) => {
+      outputQueue.add(() => console.log(args))
```
tjjfvi, line 57: Instead of pushing to a `PQueue`, can you write to a `Buffer`?

.github/actions/trun/main.ts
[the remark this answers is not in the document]
ryanleecode: And ideally we shouldn't need to go into the examples directory since we also specific the `import_map.json` which is in the root. hence a dir flag is necessary here.

.github/actions/trun/run.ts
```diff
@@ -0,0 +1,132 @@
...
+
+async function pipeThrough(reader: Deno.Reader, writer: Deno.Writer) {
+  const encoder = new TextEncoder()
+  for await (const line of readLines(reader)) {
+    await writeAll(writer, encoder.encode(`${line}\n`))
+  }
+}
+
+const wrapCode = (code: string) => `
+BigInt.prototype.toJSON = function() { return this.toString() }
```
tjjfvi, line 123: Shouldn't be necessary with the `console.js`, right?

.github/actions/trun/main.ts
```diff
@@ -0,0 +1,89 @@
...
+const sourceFileNames = Array.from(Deno.readDirSync(dir))
+  .filter((e) =>
+    e.name.match(/^.*\.ts$/g)
+    && e.isFile
+    && !ignoredFiles.has(e.name)
+    && ((filter.size > 0 && filter.has(e.name)) || filter.size === 0)
+  )
+  .map((f) => f.name)
+
+const shutdownTasks = new PQueue()
```
tjjfvi, line 36: Can you use an `AbortController` instead of this `deferred`/`PQueue` combination?

.github/actions/trun/main.ts
```diff
@@ -0,0 +1,89 @@
...
+async function shutdown(exitCode: number) {
+  console.log(`\nshutting down with exitcode ${exitCode}`)
+  browserCloseSignal.resolve()
+  await shutdownTasks.onIdle()
+  Deno.exit(exitCode)
```
tjjfvi, line 41: Here we can let deno exit naturally with a custom code like https://github.com/paritytech/capi/blob/main/main.ts#L77

.github/actions/trun/main.ts
tjjfvi: Yes, since any non-zero exit codes is failing.

tjjfvi (changes requested):
Across all JS files, variables should use the convention of treating abbreviations like words in camelCase (i.e. `fooUrl`, not `fooURL`, and `fooJs`, not `fooJS`)

.github/actions/trun/run.ts
ryanleecode: Like `JSON.stringify` each line into a buffer and then when reading `JSON.parse`? Reason its done this way is because you lose the "console richness" if you just convert it into a string and throw it into a buffer and then read from it.

.devcontainer/Dockerfile
```diff
@@ -17,10 +17,60 @@ RUN mkdir -p ${DENO_INSTALL} \
...
+ENV PUPPETEER_EXECUTABLE_PATH=/usr/bin/chromium
+
+# Ripped from https://github.com/lucacasonato/deno-puppeteer/blob/main/Dockerfile
+RUN apt-get -qq update \
+    && apt-get -qq install -y --no-install-recommends \
+    curl \
+    ca-certificates \
+    unzip \
+    fonts-liberation \
+    libappindicator3-1 \
```
kratico, line 29: If possible, I think we should have a docker image with the minimum set of packages so it doesn't get bigger and bigger.

Also, I'd suggest to merge `RUN` instructions to minimize created layers

.github/actions/trun/run.ts
tjjfvi: The arguments you receive from `console.js` are of the form `(log: string, level: number)`; the first argument has color baked in already, and the second argument indicates stdout/stderr (which isn't really relevant here). So if you `textEncoder.encode(log)`, and write that into a `Buffer`, when you print that buffer out in the end it'll have all the relevant color and formatting.

.github/actions/trun/main.ts
```diff
@@ -0,0 +1,89 @@
...
+}
+
+Deno.addSignalListener("SIGINT", () => shutdown(1))
```
kratico, line 44: I'd suggest to handle `SIGTERM` too

.github/actions/trun/run.ts
```diff
@@ -0,0 +1,132 @@
...
+      plugins: [
+        denoPlugin({
+          importMapURL,
+        }) as any,
+      ],
+      entryPoints: [path.join(dir, fileName)],
+      bundle: true,
+      write: false,
+      format: "esm",
+    })
```
kratico, line 52: It would be nice to run this using the `dnt` build.
That way we would be testing that the packaged build works in the browser too.

.github/actions/trun/validate_ignore.ts
```diff
@@ -0,0 +1,30 @@
...
+  string: ["dir", "ignore"],
+  default: { ignore: ".ignore" },
+})
+if (!dir) {
+  throw new Error("dir flag missing")
+}
+
+const decoder = new TextDecoder("utf-8")
+const files = decoder
+  .decode(await Deno.readFile(`${dir}/${ignore}`))
```
kratico, line 13: `Deno.readTextFile`

.github/actions/trun/run.ts
```diff
@@ -0,0 +1,132 @@
...
+  importMapURL?: URL
+  results: [fileName: string, exitCode: number][]
+}
+
+export async function runWithBrowser(
+  { createBrowser, importMapURL, results }: RunWithBrowserOptions,
+) {
+  const browser = await createBrowser()
+  const currentDir = new URL(import.meta.url).pathname.split("/").slice(0, -1).join("/")
+  const consoleJS = new TextDecoder().decode(await Deno.readFile(`${currentDir}/console.js`))
```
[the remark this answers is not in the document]
ryanleecode: oo didnt know this trick. very nice

.github/actions/trun/action.yml
```diff
@@ -0,0 +1,23 @@
+name: "trun"
+description: "asdf"
```
tjjfvi, line 2: This should be updated

.devcontainer/Dockerfile
ryanleecode: Aye to the 2nd comment. For the first one does it really matter? Since its just running a dev environment anyways

.github/actions/trun/run.ts
```diff
@@ -0,0 +1,133 @@
...
+
+  return (async (dir: string, fileName: string) => {
+    console.log(`running ${fileName}`)
+    const outputQueue = new PQueue({ concurrency: 1, autoStart: false })
+
+    outputQueue.add(() => console.log(`${fileName} output:`))
```
tjjfvi, line 39: This is now unused
```suggestion
```

.github/actions/trun/run.ts
```diff
@@ -0,0 +1,133 @@
...
+    })
+
+    await page.addScriptTag({ content: code, type: "module" })
+
+    const exitCode = await exit
+
+    if (exitCode != 0) {
+      for await (const line of readLines(outputBuffer)) {
+        console.log(line)
+      }
```
tjjfvi, line 74:
```suggestion
      console.log(`${fileName} output:`)
      console.log(new TextDecoder().decode(outputBuffer.bytes()))
```

.github/actions/trun/run.ts
```diff
@@ -0,0 +1,133 @@
...
+    const exit = deferred<number>()
+    await page.exposeFunction("exit", (args: string) => {
+      exit.resolve(Number.parseInt(args))
+    })
+
+    await page.addScriptTag({ content: code, type: "module" })
+
+    const exitCode = await exit
+
+    if (exitCode != 0) {
```
tjjfvi, line 71:
```suggestion
    if (exitCode !== 0) {
```

.devcontainer/Dockerfile
kratico: it would be to ease the experience for new contributors

.github/actions/trun/main.ts
```diff
@@ -0,0 +1,89 @@
...
+  : undefined
+
+const results: [fileName: string, exitCode: number][] = []
+const runner = browser
+  ? await runWithBrowser({ createBrowser, importMapURL, results })
+  : await runWithDeno({ reloadUrl: "http://localhost:4646", results })
+const paths = sourceFileNames.map((fileName) => [dir, fileName] as const)
+
+await run({ paths, runner, concurrency })
+console.log("test results", results)
```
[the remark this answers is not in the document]
ryanleecode: image

looks nice

.github/actions/trun/run.ts
ryanleecode: does our linter not automatically catch these kinds of errors? typically eslint would catch these and it would get fixed automatically if you have format on saveon.

.github/actions/trun/main.ts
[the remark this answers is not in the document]
ryanleecode: nope it works b/c they are separate

```
console.log("test results")
console.table(results.map(([fileName, exitCode]) => ({ fileName, exitCode })))

const isFailed = results.some(([_, exitCode]) => exitCode !== 0)
if (Deno.env.get("GITHUB_STEP_SUMMARY")) {
  await core.summary
    .addHeading("Failures")
    .addTable([
      [{ data: "file", header: true }],
      ...results.filter(([_, exitCode]) => exitCode !== 0).map(([name]) => [name]),
    ])
    .write({ overwrite: true })
}

```

.github/actions/trun/run.ts
tjjfvi: Hmm, not sure if deno lint has a rule to ban `==`. But auto fixing it wouldn't be good, since it does change the semantics.

.github/actions/trun/main.ts
```diff
@@ -0,0 +1,90 @@
...
+
+const results: [fileName: string, exitCode: number][] = []
+const runner = browser
+  ? await runWithBrowser({ createBrowser, importMapURL, results })
+  : await runWithDeno({ reloadUrl: "http://localhost:4646", results })
+const paths = sourceFileNames.map((fileName) => [dir, fileName] as const)
+
+await run({ paths, runner, concurrency })
+console.log("test results")
+console.table(results.map(([fileName, exitCode]) => ({ fileName, exitCode })))
```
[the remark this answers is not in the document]
ryanleecode: done

tjjfvi:
Everywhere `parseInt(x)` is used, `+x` or `Number(x)` should be used instead; `parseInt` has many unexpected and undesired behaviors.

.github/actions/trun/main.ts
```diff
@@ -0,0 +1,96 @@
+import * as core from "./deps/actions/core.ts"
```
harrysolovay, line 1: While the GH table artifact is cool, I'm not so sure utilizing actions-specific functionality in the trun CLI is merited. Better as a general CLI tool imo.

harrysolovay (changes requested):
This tool can be simplified. It doesn't need to be packaged as an action. Nor should it depend on GH core. The table artifact is unnecessary. If you're able to, please remove any non-critical environment setup and distill this into its most minimal form.

.github/actions/trun/validate_ignore.ts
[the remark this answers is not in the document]
ryanleecode: https://github.com/paritytech/capi/blob/feat/trun/.github/actions/trun/entrypoint.sh#L10

i put it here in the entrypoint script to run b4 everything else

.github/actions/trun/main.ts
ryanleecode: How about we add an output flag to trun that emits the test results as csv file w/ columns name & exitcode. and then whoever is using it has the option to read  from that csv file and package it up as a gh action.

i.e. we have the option of making the action / table summary local to capi itself, _if_ we choose to.

harrysolovay: If you're passionate about it, an output flag is fine. Instead of a csv, how about plain JSON. The flag can be `--json` (much like `deno bench`).

Regarding the table... I don't think it adds much value.

- When one sees the check passed, they'll ignore output entirely (as there should be none).
- When one sees the check failed, they'll click into the run to see the run-specific logs (and therefore the list (not table) of failures).

ryanleecode: yup thats fair. -- and json it is then

.github/actions/trun/validate_ignore.ts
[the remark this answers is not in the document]
ryanleecode: Was part of the gh actions setup -- will be removed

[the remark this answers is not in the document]
ryanleecode: on second though we do still need this dockerfile entrypoint script combination simply because you cant run the workflow in CI for browser in headless mode without installing all the necessary dependencies.

it can be just a part of the capi repo though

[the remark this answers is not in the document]
ryanleecode: it would be beneficial to users to have an example of how to run it in CI in the trun repo IMO. otherwise they will do the same digging i had to do to figure out how to run it headlessly in gh actions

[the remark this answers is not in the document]
ryanleecode: Yeah. I'm going to backtrack here. I strongly believe making this a github actions repo is essential otherwise its basically useless because the whole point of the tool is to be able to run things in CI.

ryanleecode:
will merge what is necessary to the https://github.com/paritytech/trun repo... closing this fo rnow

PR #774, opened 2023-03-22
feat: simplify dynamic dx

### LINKED ISSUE #770: dynamic metadata retrieval
The dynamic DX is broken as of #641. Needs to fall back in the case that a phantom is provided to the `Chain` used to initialize the `ChainRune`.

harrysolovay:
Resolves #770 

```ts
import { ChainRune, WsConnection } from "capi"

const chain = ChainRune.from(WsConnection, "wss://rpc.polkadot.io")

const accountInfo = await chain
  .pallet("System")
  .storage("Account")
  .entryPage(10, null)
  .run()

console.log(accountInfo)
```


fluent/ChainRune.ts
```diff
@@ -50,9 +51,41 @@ export namespace Chain {
...
+    props: RunicArgs<X, { connection: Connection; metadata?: M }>,
+  ) {
+    return Rune
+      .rec(props)
+      .unsafeAs<{ connection: Connection; metadata: M }>()
+      .into(ChainRune)
+  }
+
+  static dynamic<D>(
+    connectionCtor: {
```
harrysolovay, line 64: Simpler way to represent this type? The following won't work

- `typeof Connection` –– type compat error
- `new(discovery: D) => Connection` –– we need access to the `connect` static
- `{ connect: (discovery: D, signal: AbortSignal) => Connection }` –– for some reason infers `D` as unknown

fluent/ChainRune.ts
```diff
@@ -50,6 +51,21 @@ export namespace Chain {
...
 export class ChainRune<out C extends Chain, out U> extends Rune<C, U> {
+  static from<D, M extends FrameMetadata>(
+    connectionCtor: ConnectionCtorLike<D>,
+    discovery: D,
+    staticMetadata?: M,
+  ) {
+    const connection = ConnectionRune.from(async (signal) =>
+      connectionCtor.connect(discovery, signal)
+    )
+    const metadata = staticMetadata || Rune
```
tjjfvi, line 62:
```suggestion
    const metadata = staticMetadata ?? Rune
```

tjjfvi (approved):

PR #789, opened 2023-03-23
docs: node-first readme

harrysolovay:


Readme.md
```diff
@@ -14,49 +14,38 @@ interactions without compromising either performance or ease of use.
 - [API Reference &rarr;](https://deno.land/x/capi/mod.ts)<br />A generated API
   reference, based on type signatures and TSDocs.
 
-## At a Glance
+## Installation
 
-Run the Capi development server.
+We can substitute the URL's target (in this case the WebSocket URL of a proxy).
```
tjjfvi, line 19: What does this mean?

Readme.md
```diff
@@ -14,49 +14,38 @@ interactions without compromising either performance or ease of use.
...
+### [Deno](https://deno.land/)
 
-```json
-{
-  "imports": {
-    "@capi/polkadot/": "http://localhost:4646/frame/wss/rpc.polkadot.io/@latest/"
-  }
-}
+```ts
+import {} from "https://capi.dev/frame/wss/rpc.polkadot.io/@latest/mod.js"
```
tjjfvi, line 30: This should use the import map IMO

Readme.md
harrysolovay: Meaning that it can be swapped out with another URL / target

Readme.md
harrysolovay: Good point

harrysolovay: Actually, that isn't the case (bc the extension).

tjjfvi: True. Still, I think it's cleaner as an import map – we want our readme to show the expected use.

harrysolovay: Sounds good. In that case, is it worth import-mapping directly to the `mod.js`? That way we can keep it extension-free after all.

harrysolovay: Or I suppose this isn't the intended use, as we want people to be able to easily access files within the codegened dir (such as guards, factories, misc.)

Readme.md
```diff
@@ -14,49 +14,44 @@ interactions without compromising either performance or ease of use.
...
 {
   "imports": {
-    "@capi/polkadot/": "http://localhost:4646/frame/wss/rpc.polkadot.io/@latest/"
+    "@capi/polkadot": "http://localhost:4646/frame/wss/rpc.polkadot.io/@latest/"
   }
 }
 ```
 
-Then, open your editor and import from the mapped chain module.
+## At A Glance
```
tjjfvi, line 37:
```suggestion
## At a Glance
```

tjjfvi (approved):

PR #799, opened 2023-03-25
feat: add `storageRune.default` and re-add constants codegen

harrysolovay:
`storageRune.default`

```ts
import { System } from "polkadot_dev/mod.js"

assertEquals(
  await System.Account.default().run(),
  {
    nonce: 0,
    consumers: 0,
    providers: 0,
    sufficients: 0,
    data: { free: 0n, reserved: 0n, miscFrozen: 0n, feeFrozen: 0n }
  }
)
```

Constant codegen

```ts
import { System } from "polkadot_dev/mod.js"

System.SS58Prefix // `number`
```


harrysolovay:
Is there a path that would spare devs of the extra network interaction (using the hash to retrieve the block)? Thoughts on `finalized` producing a new `BlockHashRune`, which would have a `block` method on it? (same for `inBlock`).

tjjfvi (approved):

kratico (approved):

PR #800, opened 2023-03-25
chore: rework `finalized`/`inBlock` with `BlockHashRune`

harrysolovay:
Given that we have `inBlockHash`/`finalizedHash` and `inBlockEvents`/`finalizedEvents`, there's a chance that `finalized` alone for describing block would be confusing.


tjjfvi:
I think instead we should rename them back to `.finalized()` and use `.finalized().hash` instead of `.finalizedHash()`

harrysolovay:
Is there a path that would spare devs of the extra network interaction (using the hash to retrieve the block)? Thoughts on `finalized` producing a new `BlockHashRune`, which would have a `block` method on it? (same for `inBlock`).

tjjfvi (dismissed):

tjjfvi (approved):

PR #801, opened 2023-03-25
feat: constraints, round 2

harrysolovay:
Heavily WIP


harrysolovay:
Round 2 was fun, but alas, we'll be taking a different approach.

PR #804, opened 2023-03-27
chore: hide several internal syms from mod

harrysolovay:


tjjfvi (changes requested):

harrysolovay:
We can pick this up later as part of #323.

PR #806, opened 2023-03-27
ci: merge check workflows

tjjfvi:


.github/workflows/checks.yml
```diff
@@ -0,0 +1,84 @@
...
+    name: Formatting
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: dprint/check@2f1cf31537886c3bfb05591c031f7744e48ba8a1 # v2.2
+
+  lint:
+    name: Lint
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
```
harrysolovay, line 19: We can get rid of this timeout

.github/workflows/checks.yml
```diff
@@ -0,0 +1,84 @@
...
+            target/capi/**/_chainName
+          key: ${{ runner.os }}-capi-metadata-${{ hashFiles('import_map.json') }}
+      - run: deno task run _tasks/star.ts
+      - run: deno task cache target/star.ts
+      - run: deno task check target/star.ts
+
+  test:
+    name: Tests
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
```
harrysolovay, line 64: Can get rid of this timeout as well, no? Should we have a single top-level timeout for `checks.yml`?

Issue #790, opened 2023-03-23
simple match function on rune


nythrox:
If you have a param that's a or Rune<T | Uint8Array | undefined> or any other typescript union, and you'd like to treat each case separately, you'd have to do some combination of `.unhandle(isUint8Array)` and `.rehandle(isUint8Array, $encoder.encode)` for each separate type. 

Would be nice to have a simple match that does the same thing, but with a more understandable control flow:
```
const value: Rune<T | undefined> =
  param.match(isUint8Array, $encoder.encode)
```

related to #624 , but for general conditional control flows based on any arbitrary value

tjjfvi:
Yes, a `.handle` method is planned to do exactly that. Since `Uint8Array` is a class, you could simply write `.handle(Uint8Array, $encoder.encode)`. We'll probably also add a simple `not` around the same time, to enable `.handle(not(undefined), $encoder.encode)`

harrysolovay:
Good point, especially since `handle` will accept guards too. Shall we close #624 in favor of this issue (and rename this issue to `rune.handle`)?

harrysolovay:
In addition to a `not` util, we might want a simple curried `isTagged` for easier creation of guards to `handle` tagged unions.

```ts
const isType = isTagged("type")

rune
  .handle(isType("Id"), () => {})
  .handle(isType("Address20"), () => {})
```

tjjfvi:
> Shall we close https://github.com/paritytech/capi/issues/624 in favor of this issue

No, #624 is also important, because repeated calls to `.handle` could flow into each other.

nythrox:
I'm experimenting with this on my branch, I a simple `match` (soon to be handle) to see if it improves my control flow. Impl was suprisingly simple

```ts
  match<T2 extends T, T3, U3>(
    fn: Guard<T, T2>,
    alt: (rune: ValueRune<T2, never>) => Rune<T3, U3>,
  ): ValueRune<T | T3, U | U3> {
    return this.unhandle(fn).rehandle(fn as Guard<T2 | U, T2>, alt)
  }
```

tjjfvi:
The issue with that implementation is that it may unintentionally rehandle values; part of the reason to have a `.handle` at all is to make that pattern less common.

PR #814, opened 2023-03-27
feat: handle rune

nythrox:
#790 

possible changes: 
- change name of `RunMatch` (RunHandle is already taken, maybe refactor)
- handleError function

### DIFF
--- a/rune/ValueRune.ts
+++ b/rune/ValueRune.ts
@@ -60,7 +60,7 @@ export class ValueRune<out T, out U = never> extends Rune<T, U> {
     return ValueRune.new(
       RunHandle,
       this,
-      (x: T): x is T2 => checkGuard(x, guard),
+      guard,
       alt(this as never),
     )
   }
--- a/rune/ValueRune.ts
+++ b/rune/ValueRune.ts
@@ -58,10 +58,10 @@ export class ValueRune<out T, out U = never> extends Rune<T, U> {
     alt: (rune: ValueRune<T2, never>) => Rune<T3, U2>,
   ): ValueRune<Exclude<T, T2> | T3, U | U2> {
     return ValueRune.new(
-      RunMatch,
+      RunHandle,
       this,
       (x: T): x is T2 => checkGuard(x, guard),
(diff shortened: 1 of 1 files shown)


rune/ValueRune.ts
```diff
@@ -53,6 +53,18 @@ export class ValueRune<out T, out U = never> extends Rune<T, U> {
...
   }
 
+  handle<T2 extends T, T3, U2>(
+    guard: Guard<T, T2>,
+    alt: (rune: ValueRune<T2, never>) => Rune<T3, U2>,
+  ): ValueRune<Exclude<T, T2> | T3, U | U2> {
+    return ValueRune.new(
+      RunMatch,
+      this,
+      (x: T): x is T2 => checkGuard(x, guard),
```
tjjfvi, line 63: This should pass `guard` directly like in `RunUnhandle`

rune/ValueRune.ts
```diff
@@ -53,6 +53,18 @@ export class ValueRune<out T, out U = never> extends Rune<T, U> {
...
+      ValueRune.new(RunChain, this, alt(this as never)),
```
tjjfvi, line 64:
```suggestion
      alt(this as never),
```

rune/ValueRune.ts
```diff
@@ -151,6 +163,28 @@ class RunMap<T1, U, T2> extends Run<T2, U> {
   }
 }
 
+class RunMatch<T, T2 extends T, T3, U, U2> extends Run<Exclude<T, T2> | T3, U | U2> {
```
tjjfvi, line 166: Yeah, this should be called `RunHandle`, and the existing one should be renamed to `RunRehandle`

rune/ValueRune.ts
tjjfvi: This wasn't resolved

nythrox: oops my bad, I updated the Run function for this and forgot to update it here

tjjfvi (approved):

PR #818, opened 2023-03-28
chore: asset teleportation example clean up

harrysolovay:


kratico (approved):
LGTM

examples/xcm/asset_teleportation.ts
```diff
@@ -0,0 +1,67 @@
...
+    feeAssetItem: 0,
+    weightLimit: WeightLimit.Unlimited(),
+  })
+  .signed(signature({ sender: alice }))
+  .sent()
+  .dbgStatus()
+  .finalized()
+  .run()
+
+// TODO: have the recipient associate the downward message with the sender
```
kratico, line 56: @hbulgarini Would you know if there is a standard around this?

Is it possible to correlate a message from the sender to the recipient?

tjjfvi (approved):

examples/xcm/asset_teleportation.ts
hbulgarini: not sure specifically what you mean by standard but my recommendation is that for building a teleporting message you can basically copy what the extrinsic `teleport_assets` in the `pallet_xcm` is doing. Specifically this is the sequence of XCM messages built for teleporting: https://github.com/paritytech/polkadot/blob/master/xcm/pallet-xcm/src/lib.rs#L1202:L1207

> Is it possible to correlate a message from the sender to the recipient?

Not AFAIK. The only way that i can think of is setting the recipient as the caller of the extrinsic.

PR #823, opened 2023-03-29
feat: devnets

### LINKED ISSUE #805: star not exiting cleanly on ci
Introduced in #777 – seems to exit cleanly locally... perhaps related to how zombienet is killing the processes ([badly](https://github.com/paritytech/zombienet/blob/aeac842843a070f3454942489ca0a6c41e795f86/javascript/packages/orchestrator/src/providers/native/nativeClient.ts#L121-L146))

### LINKED ISSUE #810: support all providers on capi.dev

### LINKED ISSUE #817: `capi.config.ts`
Instead of specifying all our codegen in URLs, we're going to move to a config-file-based setup. Something like this:
```ts
// capi.config.ts

import { CapiConfig, fetchBinary } from "capi"

const polkadot = fetchBinary("polkadot", "v0.9.40")
const polkadotParachain = fetchBinary("polkadot-parachain", "v0.9.380")
const substrateContractsNode = fetchBinary("substrate-contracts-node", "v0.9.40")

export const config: CapiConfig = {
  server: "http://localhost:4646/", // default "https://capi.dev/"
  chains: {
    polkadot: {
      url: "wss://rpc.polkadot.io/",
      version: "v0.9.40",
    },
    polkadotDev: {
      binary: polkadot,
    },
    westendDev: {
      binary: polkadot,
      runtime: "westend",
    },
    contractsDev: {
      binary: substrateContractsNode,
    },
  },
  networks: {
    rococoDev: {
      relay: {
        binary: polkadotParachain,
        runtime: "rococo-local",
        nodes: 4,
      },
      parachains: {
        statemine: {
          id: 1000,
          binary: polkadotParachain,
          runtime: "statemine-local",
        },
        contracts: {
          id: 2000,
          binary: polkadotParachain,
          runtime: "rococo-contracts-local",
        },
      },
    },
  },
}
```

Given this config file, running `capi install` (name tbd) will generate an import map as follows:
```json
{
  "imports": {
    "@capi/polkadot/": "http://localhost:4646/polkadot-abcdef1234567890/",
    "@capi/polkadot-dev/": "http://localhost:4646/polkadot-dev-abcdef1234567890/",
    "@capi/westend-dev/": "http://localhost:4646/westend-dev-abcdef1234567890/",
    "@capi/contracts-dev/": "http://localhost:4646/contracts-dev-abcdef1234567890/",
    "@capi/rococo-dev-relay/": "http://localhost:4646/rococo-dev-relay-abcdef1234567890/",
    "@capi/rococo-dev-statemine/": "http://localhost:4646/rococo-dev-statemine-abcdef1234567890/",
    "@capi/rococo-dev-contracts/": "http://localhost:4646/rococo-dev-contracts-abcdef1234567890/",
  }
}
```
Here, `abcdef1234567890` stands for an automatically generated hash (which would be different for each). For Node.js, it would generate a similar thing in the `package.json`, pointing to `pkg.tar`s.

Under the hood, the `capi install` command would get the metadata for each node, upload it to the server (be it local or `capi.dev`), and get back the hash to use in the URL.

This would work with all chains for both the local server (for our development purposes) and `capi.dev` (for all consumers).

At runtime, instead of relying on the existence of the local server for the chains, we'll adopt another approach, perhaps involving a daemon. Capi will still orchestrate spawning nodes, generating test users etc., but separately from the local server.

This new direction will solve #810, and overall simplify the way we specify codegen.

This will entail rewriting much of the way we handle the codegen, including the provider setup (although the `codegen/` directory proper will be largely untouched). @tjjfvi will take on the initial rewrite, after which we can identify subtasks and load-balance.

An important part of this configuration lies in the `networks` section – this configuration will be an alternative to the zombienet configuration files, aiming to identify a more minimal subset that matches our needs. An open question is whether we'll convert this to a zombienet config, or orchestrate the network ourselves – given the degree of customization we may want, taking on our own orchestration may be the best path. @kratico will be experimenting with this.

tjjfvi:
~~Capi Automated Pseudo-Networks reporting for duty~~

- Resolves #805
- Resolves #810
- Resolves #817


_tasks/await_deployment.ts
```diff
@@ -0,0 +1,9 @@
+import { _getDeploymentUrl } from "../server/capi.dev/delegator.ts"
+
+const sha = Deno.env.get("CAPI_SHA")!
+
+const deploymentPollInterval = 5_000
+
+while (!(await _getDeploymentUrl(sha))) {
+  await new Promise((r) => setTimeout(r, deploymentPollInterval))
```
harrysolovay, line 8: Nit: can use `delay` from `deps/std/async.ts`

.github/workflows/checks.yml
```diff
@@ -33,29 +37,53 @@ jobs:
...
+        with:
+          path: ~/.cache/deno
+          key: cache-${{ env.CAPI_SHA }}
+      - run: deno run -A _tasks/await_deployment.ts
+        timeout-minutes: 1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      - run: deno run -A _tasks/use_remote.ts
+      - run: deno task run _tasks/star.ts
+      - run: deno cache target/star.ts || deno cache target/star.ts || deno cache target/star.ts
```
harrysolovay, line 59: Add comment description why + reference #856

_tasks/use_remote.ts
harrysolovay, line 1: Top-level description of what this task does

capn/CapiConfig.ts
```diff
@@ -0,0 +1,25 @@
...
+  parachains?: Record<string, {
+    binary: Binary
+    chain: string
+    id: number
+    nodes?: number
+  }>
+}
+
+export interface CapiConfig {
+  server: string
```
harrysolovay, line 23: Default to `capi.dev` (once we add version-checking)? #859

harrysolovay (dismissed):

Issue #821, opened 2023-03-29
Examples not running


wirednkod:
## Bug Report

Tried to follow the tutorials to run some examples; but stumbled upon some issues

### Current Behavior

After cloning CAPI repo, in a terminal window I run:
```bash
~ deno run -A https://deno.land/x/capi/main.ts

Warning Implicitly using latest version (v0.1.0-beta.31) for https://deno.land/x/capi/main.ts
Capi server listening at "http://localhost:4646/"
```
in another terminal window (same dir) I run:
```bash
~ deno task run ./examples/bonded.ts

Task run deno task capi -- deno run -A -r=http://localhost:4646/ "./examples/bonded.ts"
Task capi deno run -A main.ts "--" "deno" "run" "-A" "-r=http://localhost:4646/" "./examples/bonded.ts"
Reusing existing Capi server at "http://localhost:4646/"
error: Import 'http://localhost:4646/frame/dev/polkadot/@v0.9.39-1/mod.js' failed: 500 Internal Server Error
    at file:///home/wirednkod/Documents/repos/parity/ts/capi/examples/bonded.ts:2:25
inner command exited with status code 1
```

### Expected Behavior

Example should execute

### Steps To Reproduce

See above

### Environment

- Deno: Deno 1.32.1
- OS: Ubuntu

tjjfvi:
Looks like our [timeout](https://github.com/paritytech/capi/blob/main/providers/frame/FrameBinProvider.ts#L8) is too short for downloading the binary for the first time.

kratico:
@tjjfvi I tried to repro this issue in a fresh dev container and it seems to be a binary issue
```
vscode ➜ /workspaces/capi $ deno task capi bin polkadot v0.9.39-1 -- --dev
Task capi deno run -A main.ts "bin" "polkadot" "v0.9.39-1" "--" "--dev"
Downloading polkadot/v0.9.39-1/x86_64-unknown-linux-gnu
<jemalloc>: MADV_DONTNEED does not work (memset will be used instead)
<jemalloc>: (This is the expected behaviour if you are running under QEMU)
2023-03-29 12:51:30 Parity Polkadot    
2023-03-29 12:51:30 ✌️  version 0.9.39-1-298b4aefe1b    
2023-03-29 12:51:30 ❤️  by Parity Technologies <admin@parity.io>, 2017-2023    
2023-03-29 12:51:30 📋 Chain specification: Development    
[...]

PR #824, opened 2023-03-29
feat: ed25519 signature example

### LINKED ISSUE #682: examples utilizing misc. kinds of signing
The basic transfer example, only using a different signature type / means of signing. Perhaps https://deno.land/x/ed25519

Research involves looking into which test chains support a ed25519 scheme and how to run it w/ zombienet. 

implementation could involve modifying https://github.com/paritytech/capi/blob/main/crypto/Sr25519.ts to become a subclass that inherits a base class and adding a ED25519 subclass.

ryanleecode:
Resolves #682

## Changes
- ~~Updates the test user public keys list to also have ed25119 keys~~
- ~~Abstract Sr25519 into a Keypair class and adds Ed25519~~
- Adds transfer example using ED2519 signatures
- ~~Adds script for recreating test user public keys~~

## Notes
- ~~import map change is a requirement from https://github.com/paulmillr/noble-ed25519/tree/1.7.3~~
- ~~Not using latest version of noble-ed25519 b/c it doesn't work~~
- ~~Using older version of `main` because `main` is currently broken on linux machines. See #821~~


tjjfvi (changes requested):
> - Not using latest version of noble-ed25519 b/c it doesn't work

Can you elaborate?

> - import map change is a requirement from https://github.com/paulmillr/noble-ed25519/tree/1.7.3

Can you use `https://esm.sh/@noble/ed25519` instead?

> - Adds script for recreating test user public keys

There is already a script for this: <https://github.com/paritytech/capi/blob/main/_tasks/generate_artifacts.ts#L30-L37>

> - Updates the test user public keys list to also have ed25119 keys
> - Abstract Sr25519 into a Keypair class and adds Ed25519

We don't use ed25519 anywhere else, so I'd like to keep the relevant logic contained to the example.

Instead of putting this logic into the crypto directory, can you write the signing logic etc. into the example? (I.e. without using a `Ed25519` class). It will then also be a good example of how the signing works under the hood.

Then, rather than add ed25519 test users, you can have an existential deposit from `alexa` to the ed25519 account (perhaps `eddie`), and then do a transfer from `eddie` to `billy` using the ed25519 signing.

ryanleecode:
> Can you elaborate?

Running noble ed25519 on latest version gives me an error about sha512sync

```
vscode ➜ /workspaces/capi (feat/ed25519-signature-example) $ deno task run examples/ed25519_signature.ts
Task run deno task capi -- deno run -A -r=http://localhost:4646/ "examples/ed25519_signature.ts"
Task capi deno run -A main.ts --dbg "--" "deno" "run" "-A" "-r=http://localhost:4646/" "examples/ed25519_signature.ts"
Reusing existing Capi server at "http://localhost:4646/"
error: Uncaught (in promise) Error: etc.sha512Sync not set
const err = (m = ''): never => { throw new Error(m); }; // error helper, messes-up stack trace
                                       ^
    at err (https://deno.land/x/ed25519@2.0.0/index.ts:12:40)
    at sha512s (https://deno.land/x/ed25519@2.0.0/index.ts:193:47)
    at getExtendedPublicKey (https://deno.land/x/ed25519@2.0.0/index.ts:208:55)
    at Module.getPublicKey (https://deno.land/x/ed25519@2.0.0/index.ts:211:44)
    at Function.fromSecret (file:///workspaces/capi/crypto/Keypair.ts:32:26)
    at Sr25519.toEd25519 (file:///workspaces/capi/crypto/Keypair.ts:70:38)
    at file:///workspaces/capi/examples/ed25519_signature.ts:6:28
```

> Can you use https://esm.sh/@noble/ed25519 instead?

Same error as above

> There is already a script for this: https://github.com/paritytech/capi/blob/main/_tasks/generate_artifacts.ts#L30-L37

didn't see that mb

> We don't use ed25519 anywhere else, so I'd like to keep the relevant logic contained to the example.

sounds good

tjjfvi:
> > Can you use https://esm.sh/@noble/ed25519 instead?
> 
> Same error as above

In that case, what about https://esm.sh/@noble/ed25519@1.7.3

examples/ed25519_signature.ts
```diff
@@ -0,0 +1,42 @@
+import { Balances, users } from "westend_dev/mod.js"
+import { MultiAddress } from "westend_dev/types/sp_runtime/multiaddress.js"
+import * as ed from "../deps/ed25519.ts"
```
tjjfvi, line 3: Since this is only a dependency of the example, let's inline the url here and delete `deps/ed25519.ts`.

ryanleecode:
> https://esm.sh/@noble/ed25519@1.7.3

```
vscode ➜ /workspaces/capi (feat/ed25519-signature-example) $ deno task run examples/ed25519_signature.ts
Task run deno task capi -- deno run -A -r=http://localhost:4646/ "examples/ed25519_signature.ts"
Task capi deno run -A main.ts --dbg "--" "deno" "run" "-A" "-r=http://localhost:4646/" "examples/ed25519_signature.ts"
Reusing existing Capi server at "http://localhost:4646/"
error: Uncaught (in promise) Error: utils.sha512Sync must be set to use sync methods
    at Y (https://esm.sh/v113/@noble/ed25519@1.7.3/deno/ed25519.mjs:2:10289)
    at tt (https://esm.sh/v113/@noble/ed25519@1.7.3/deno/ed25519.mjs:2:10444)
    at Object.Zt [as sign] (https://esm.sh/v113/@noble/ed25519@1.7.3/deno/ed25519.mjs:2:10813)
    at Object.sign (file:///workspaces/capi/examples/ed25519_signature.ts:16:20)
    at Codec._encode (file:///workspaces/capi/frame_metadata/Extrinsic.ts:63:40)
    at Codec._encode (https://deno.land/x/scale@v0.11.0/codecs/lenPrefixed.ts:14:14)
    at Codec.encodeAsync (https://deno.land/x/scale@v0.11.0/common/codec.ts:122:10)
    at RunMap.fn (file:///workspaces/capi/fluent/CodecRune.ts:9:26)
    at RunMap._evaluate (file:///workspaces/capi/rune/ValueRune.ts:150:40)
    at async RunMap.evaluate (file:///workspaces/capi/rune/Rune.ts:222:14)
```

examples/ed25519_signature.ts
```diff
@@ -0,0 +1,42 @@
...
+import { Rune } from "../mod.ts"
+import { signature } from "../patterns/signature/polkadot.ts"
+
+const [alexa, billy, eddie] = await users(3)
+
+const ed25519PrivateKey = eddie.secretKey.slice(0, 32)
+const ed25519PublicKey = await ed.getPublicKey(ed25519PrivateKey)
```
tjjfvi, line 10:
```suggestion
const { alexa, billy } = await users()

const eddieSecretKey = crypto.getRandomValues(new Uint8Array(32))
const eddiePublicKey = await ed.getPublicKey(eddieSecretKey)
```

examples/ed25519_signature.ts
```diff
@@ -0,0 +1,42 @@
...
+
+const ed25519Eddie = Rune.rec({
```
tjjfvi, line 12:
```suggestion
const eddie = Rune.rec({
```

tjjfvi (commented):
Looks a lot better; just a few nits

tjjfvi (approved):
Beautiful
image

Issue #827, opened 2023-03-30
fix multisig_pure_proxy_stash.ts


ryanleecode:
```
Fund Multisig: ready
Fund Multisig: { inBlock: "0x673eae1bd13347115fd7f1c82cd9f5892b4cc80dd6e1a172e392b88c656b788c" }
Fund Multisig: { finalized: "0x673eae1bd13347115fd7f1c82cd9f5892b4cc80dd6e1a172e392b88c656b788c" }
Alice Ratify: ready
Alice Ratify: { inBlock: "0xaf56cde5d93954983e0b0836d468463829f20cd6d03308b28837713a3e863d85" }
Alice Ratify: { finalized: "0xaf56cde5d93954983e0b0836d468463829f20cd6d03308b28837713a3e863d85" }
Bob Ratify: ready
Bob Ratify: { inBlock: "0x27f6075dab996c8459ce93276b993b26a2f6d1d0097779aa5a02e13e4a714753" }
Bob Ratify: { finalized: "0x27f6075dab996c8459ce93276b993b26a2f6d1d0097779aa5a02e13e4a714753" }
error: Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'pure')
        value = value[key]
                     ^
    at RunMap.fn (file:///workspaces/capi/rune/ValueRune.ts:50:22)
    at RunMap._evaluate (file:///workspaces/capi/rune/ValueRune.ts:162:40)
    at async RunMap.evaluate (file:///workspaces/capi/rune/Rune.ts:222:14)
    at async ValueRune.iter (file:///workspaces/capi/rune/Rune.ts:79:23)
    at async ValueRune.run (file:///workspaces/capi/rune/Rune.ts:66:22)
    at async file:///workspaces/capi/examples/multisig_pure_proxy_stash.ts:42:22

```

ryanleecode:
```
dispatchError { index: 30, error: Uint8Array(4) [ 12, 0, 0, 0 ] }
```

```
{
  id: 30,
  name: "Multisig",
  storagePrefix: "Multisig",
  storage: {
    Multisigs: {
      singular: false,
      name: "Multisigs",
      key: $storageKey("Multisig", "Multisigs", $.tuple([Array])),
      partialKey: $storageKey("Multisig", "Multisigs", $partialMultiKey([Array])),
      value: $.object($.field([Array]), $.field([Array]), $.field([Array]), $.field([Array])),
      docs: " The set of open multisig operations.",
      default: undefined
[...]

PR #829, opened 2023-03-30
fix: multisig rune proof size

ryanleecode:
resolves: #827

https://github.com/paritytech/substrate/pull/13268

broke the weight calculation causing the proof size requirement to exceed 0 

need to wait for this to be merged and released is my understanding

https://github.com/paritytech/substrate/pull/13766

### DIFF
--- a/patterns/multisig/MultisigRune.ts
+++ b/patterns/multisig/MultisigRune.ts
@@ -58,7 +58,9 @@ export class MultisigRune<out C extends Chain, out U>
           call,
           otherSignatories: this.otherSignatories(sender),
           storeCall: false,
-          maxWeight: call.feeEstimate().access("weight"),
+          maxWeight: call.feeEstimate().access("weight")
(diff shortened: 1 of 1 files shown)


patterns/multisig/MultisigRune.ts
```diff
@@ -58,7 +58,9 @@ export class MultisigRune<out C extends Chain, out U>
           call,
           otherSignatories: this.otherSignatories(sender),
           storeCall: false,
-          maxWeight: call.feeEstimate().access("weight"),
+          maxWeight: call.feeEstimate().access("weight")
+            // TODO: revert when this is merged https://github.com/paritytech/substrate/pull/13766
+            .map((weight) => ({ ...weight, proofSize: 200_000_000n })),
```
ryanleecode, line 63: this number is arbitrary and random

harrysolovay (approved):
Just tried it & surely enough, seems to fix the example!

PR #832, opened 2023-03-31
feat: add script/main-network.ts

kratico:


kratico:
will be implemented in #823

PR #833, opened 2023-03-31
chore: update feat estimate api

ryanleecode:


fluent/ExtrinsicRune.ts
```diff
@@ -46,20 +49,35 @@ export class ExtrinsicRune<out C extends Chain, out U> extends PatternRune<Chain
...
-      .map(({ weight, ...rest }) => ({
-        ...rest,
-        weight: {
-          proofSize: BigInt(typeof weight === "number" ? 0 : weight.proof_size),
-          refTime: BigInt(typeof weight === "number" ? weight : weight.ref_time),
-        },
-      }))
+    const $queryInfoCallArgs = Rune.fn($extrinsic)
+      .call(this.chain.into(ValueRune).access("metadata"))
+      .map((ext) => $.tuple(ext, $.u32))
```
harrysolovay, line 60: I wonder if we could/should tweak [SCALE's `lenPrefixed` codec](https://github.com/paritytech/scale-ts/blob/dc64f6564044b8626535d717c9c2bc8e24cc6a9a/codecs/lenPrefixed.ts) to support postfixing instead / specificity of direction?

fluent/ExtrinsicRune.ts
```diff
@@ -46,20 +49,35 @@ export class ExtrinsicRune<out C extends Chain, out U> extends PatternRune<Chain
...
+      .into(CodecRune)
+      .encoded(Rune.tuple([
+        Rune.rec({
+          protocolVersion: ExtrinsicRune.PROTOCOL_VERSION,
+          call: this,
+        }),
+        this.encoded().access("length"),
+      ]))
+
+    const $transactionPaymentApiQueryInfoResult = $.object(
```
harrysolovay, line 72: Is there a consistent path at which this type exists? If so, you could access the codec from the metadata's types instead of manually hard-coding its shape. `metadata.access("types", path)`.

harrysolovay (changes requested):

fluent/ExtrinsicRune.ts
```diff
@@ -46,20 +49,32 @@ export class ExtrinsicRune<out C extends Chain, out U> extends PatternRune<Chain
       .call(this.chain.into(ValueRune).access("metadata"))
```
ryanleecode, line 49: .call(this.chain.metadata)

ryanleecode:
It now works before wasn't working because the codec wasn't using `compact` for `u64` types

fluent/ExtrinsicRune.ts
```diff
@@ -1,5 +1,7 @@
 import { blake2_256, hex } from "../crypto/mod.ts"
+import { concat } from "../deps/std/bytes.ts"
 import { $extrinsic, Signer } from "../frame_metadata/Extrinsic.ts"
+import { $ } from "../mod.ts"
```
tjjfvi, line 4:
```suggestion
import * as $ from "../deps/scale.ts"
```

tjjfvi (approved):

harrysolovay (approved):

tjjfvi (approved):

PR #834, opened 2023-03-31
fix: cors server errors

### LINKED ISSUE #793: cors errors when running examples in browser
```
Access to fetch at 'http://localhost:4646/frame/dev/polkadot/@v0.9.37/user_i' from origin 'null' has been blocked by CORS policy: Response to preflight request doesn't pass access control check: No 'Access-Control-Allow-Origin' header is present on the requested resource. If an opaque response serves your needs, set the request's mode to 'no-cors' to fetch the resource with CORS disabled.
VM25:2942          POST http://localhost:4646/frame/dev/polkadot/@v0.9.37/user_i net::ERR_FAILED
users2 @ VM25:2942
(anonymous) @ VM25:21410
VM20:2728 TypeError: Failed to fetch
    at users2 (<anonymous>:2942:28)
    at <anonymous>:21410:21
```

ryanleecode:
Resolves #793

### DIFF
--- a/providers/frame/common.ts
+++ b/providers/frame/common.ts
@@ -58,23 +58,17 @@ export const createUsers ${
 }
 
 export async function handleCount(request: Request, cache: { count: number }): Promise<Response> {
-  switch (request.method) {
-    case "POST": {
-      const body = await request.json()
-      $.assert($.field("count", $.u32), body)
-      const { count } = body
-      const index = cache.count
-      const newCount = index + count
-      if (newCount < DEFAULT_TEST_USER_COUNT) cache.count = newCount
-      else throw new Error("Maximum test user count reached")
-      return new Response(JSON.stringify({ index }), {
-        status: 200,
-        headers: { "Content-Type": "application/json" },
-      })
-    }
-    default:
-      return new Response(null, { status: 200 })
-  }
+  const body = await request.json()
+  $.assert($.field("count", $.u32), body)
+  const { count } = body
+  const index = cache.count
+  const newCount = index + count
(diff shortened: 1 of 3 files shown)


providers/frame/common.ts
```diff
@@ -58,17 +58,23 @@ export const createUsers ${
...
-  const { count } = body
-  const index = cache.count
-  const newCount = index + count
-  if (newCount < DEFAULT_TEST_USER_COUNT) cache.count = newCount
-  else throw new Error("Maximum test user count reached")
-  return new Response(JSON.stringify({ index }), {
-    status: 200,
-    headers: { "Content-Type": "application/json" },
-  })
+  switch (request.method) {
```
tjjfvi, line 61: Why was this change made?

ryanleecode: When it does cors it sends a OPTION request to this endpoint. that request doesnt have a body so it would just crash

tjjfvi: In that case, can you handle OPTIONS requests in `handleCors`?

ryanleecode: Wouldn't it still crash since this handler would trigger before `handleCors`?

ryanleecode: I guess it would be caught by the handleError we would just ignore the response

tjjfvi: > Wouldn't it still crash since this handler would trigger before `handleCors`?

I was suggesting that you check in `handleCors` if the method is `OPTIONS` before calling the inner handler

tjjfvi (approved):

Issue #811, opened 2023-03-27
constraint-related changes to metadata shape


harrysolovay:
Do we need to represent `extrinsic` and system events storage in a special way such that we can more easily build constraints?

tjjfvi:
For my future reference: `| { type: Exclude<X["type"], ...> }`

tjjfvi:
```ts
type Constraint<C extends Chain> = {
  someUnionType: $.Decodec<
    | { type: "foo", abc: 123 }
    | { type: Exclude<$.Native<C["someUnionType"]>["type"], "foo"> }
    // | { type: string but not "foo" }
  >
}
```

PR #843, opened 2023-04-04
feat: ink contract events

### LINKED ISSUE #570: ink contract event decoding
Will be simpler once we can make use of codegen (aka. pending #517).

kratico:
resolve #570 #845

- Implement Ink contract emitted event decoding.
- Upgrade to Ink v4 metadata
- Change example contract to [Ink ERC-20 example](https://github.com/paritytech/ink-examples/blob/bc6b5c1dd7e3b5af38b7795b2cfe6ff6b83c466d/erc20/lib.rs#L4)


patterns/ink/events.ts
```diff
@@ -1,5 +1,19 @@
 import { AccountIdRune, ApplyExtrinsicEventPhase, Rune, RunicArgs } from "../../mod.ts"
 
+export interface ContractsRuntimeEvent {
+  phase: ApplyExtrinsicEventPhase
+  event: {
+    type: "Contracts"
+    value: {
+      contract?: Uint8Array
```
kratico, line 8: Not every contract runtime event has the `contract` attribute

See https://github.com/paritytech/substrate/blob/ae83a672f2bbe6a4f7391f6a5b2b6fd7ad4f8651/frame/contracts/src/lib.rs#L747-L816

examples/ink/interact.eg.ts
```diff
@@ -51,7 +55,11 @@ await contract
   .sent()
   .dbgStatus("Flip:")
   .inBlockEvents()
-  .pipe(contract.filterContractEvents)
+  .pipe((events) =>
+    Rune.resolve(events).map((events) => events.filter((e) => RuntimeEvent.isContracts(e.event)))
+  )
+  // .pipe(contract.filterContractEvents)
```
kratico, line 61: @harrysolovay  Pattern development question
What would be better?
- filter using the `RuntimeEvent.isContracts`
- filter hardcoding the `Contracts` value (coming from [`construct_runtime!`](https://github.com/paritytech/substrate-contracts-node/blob/fbc28a7ad4b4a58bc9d8bc848f5f3c46018ec81a/runtime/src/lib.rs#L386)) in the `InkRune`

patterns/ink/InkRune.ts
```diff
@@ -86,18 +87,24 @@ export class InkRune<out C extends Chain, out U>
...
       )
   }
 
+  // TODO: protect / consider whether this should be exposed
+  filterContractEmittedEvents = <X>(...[events]: RunicArgs<X, [events: Event[]]>) => {
+    return this.filterContractEvents(events).map((events) => events.filter(isContractEmitted))
+  }
+
   // TODO: improve
   decodeErrorEvent = <X>(...[failRuntimeEvent]: RunicArgs<X, [any]>) => {
```
harrysolovay, line 91: Should we rename this to `unhandleFailed` to keep it somewhat consistent with the extrinsic error unhandling? The difference ofc would be that one would pipe into this method (instead of it being accessible on an `ExtrinsicEventsRune`). Thoughts?

harrysolovay (changes requested):
To keep consistency between `$event` and our error-related code, let's move [the error codec](https://github.com/paritytech/capi/blob/feat/ink-events/patterns/ink/InkRune.ts#L92-L98) into the `InkMetadataRune`, with the name `$error`.

patterns/ink/InkMetadataRune.ts
```diff
@@ -35,14 +35,45 @@ export class InkMetadataRune<out U> extends Rune<InkMetadata, U> {
...
+          $.variant(
+            event.label,
+            $.object(...event.args.map((arg) => $.field(arg.label, codecs[arg.type.type]!))),
+          )
+        ),
+      )
+    )
+    .into(CodecRune)
+
+  $error<C extends Chain, U, X>(
```
harrysolovay, line 60: On second thought... is this any different from [`unhandleFailed`](https://github.com/paritytech/capi/blob/794edd47ee685b6046676eea88173606793f12d2/fluent/ExtrinsicsEventsRune.ts#L14)?

harrysolovay: Might have implemented this before we had `unhandleFailed`. If they are in fact the same, then we can likely scrap this, and `into` the `ExtrinsicEventsRune`.

harrysolovay: Might have implemented this before we had `unhandleFailed`. If they are in fact the same, then we can likely scrap this, and `into` the `ExtrinsicEventsRune`.

kratico: I'm not very familiar with that API.
I'll investigate

patterns/ink/events.ts
harrysolovay, line 1: We'll clean this up once we solve constraints / #811

harrysolovay (dismissed):

examples/ink/interact.eg.ts
```diff
@@ -2,13 +2,12 @@
...
- *
- * The Ink patterns simplify the reading of contract instance state and events, as well as
+ * @description The Ink patterns simplify the reading of contract instance state and events, as well as
  * the submission of transactions.
  */
 
-import { assertNotEquals } from "asserts"
-import { alice } from "capi"
+import { assert } from "asserts"
+import { $, alice, bob } from "capi"
```
tjjfvi, line 10: Should use test users

examples/ink/interact.eg.ts
```diff
@@ -26,36 +25,53 @@ if (!address) {
...
+)
+console.log(events)
+
 // Retrieve the final state.
 const finalState = await state.run()
-console.log("final state:", finalState)
+console.log("Alice final balance:", finalState)
 
-assertNotEquals(initialState, finalState)
+assert(initialState > finalState)
```
tjjfvi, line 77:
```suggestion
assert(finalState < initialState)
```

examples/ink/interact.eg.ts
harrysolovay: This is one of those unique cases where a constant test user is preferable: the contract is deployed from Alice, hence only Alice has funds to transfer. We could do this with `createUser`, but it would involve merging the deploy and interact examples. I'm not opposed to this, but let's perhaps do this at a later date.

tjjfvi: In that case, can you put that context in a comment?

tjjfvi (approved):

PR #851, opened 2023-04-05
some tweaks

harrysolovay:


examples/xcm/reserve_transfer.eg.ts
```diff
@@ -1,62 +1,49 @@
...
+/**
+ * @title XCM Reserve Asset Transfer
+ * @stability nearing
+ * @description Perform an XCM reserve asset transfer, in which two chains, which
+ * do not trust one another, rely on a third chain to store assets and facilitate
+ * the exchange.
+ */
+
+import { alice, bob, hex, Rune, ValueRune } from "capi"
+import { signature } from "capi/patterns/signature/statemint.ts"
```
kratico, line 10: it seems that statemine `signature` is generic over the 3 chains Rococo, Statemine and Trappist

It would be better to rename the file

harrysolovay: Same thing is true for the signature in `polkadot.ts`, which applies to Westend, misc. What is the *main* chain bearing this signature?

kratico: I copied from `polkadot.ts`

There is something that we may need to tackle later, that is
- the `tip` could be a native asset
- or, the `tip` could be an asset from the assets pallet in statemine

And that affects the extra arguments in the signature data

harrysolovay: That's fine. We can adjust later on. `statemine.ts` would likely make for a better pattern file name, yes?

kratico: ok

kratico (approved):

PR #867, opened 2023-04-10
feat: flat codegen

### LINKED ISSUE #849: Typescript autocompletion is laggy in VSCode
This is probably related to an update in the last 2 weeks

### LINKED ISSUE #864: rework codegen to stave off LSP file count limit
```
Hit the language server document preload limit of 1000 file system entries. You may want to use the "deno.enablePaths" configuration setting to only have Deno partially enable a workspace.
```

tjjfvi:
- Resolves #864
- Resolves #849

This exposes all types from the codegen root sans namespaces. When type names conflict, relevant module sections are prepended to disambiguate (e.g. `PalletBalancesEvent` and `XcmV2Junctions`).

Additionally, this makes `@capi/foo` an alias for `@capi/foo/mod.js`, so we can use the same import paths for both deno and node.

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -56,8 +56,8 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       - run: deno run -A _tasks/use_remote.ts
       - run: deno task run _tasks/star.ts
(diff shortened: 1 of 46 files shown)


harrysolovay (approved):
Just incredible. Huge improvement.

PR #870, opened 2023-04-11
feat: archive devnet stdout/stderr on workflow failure

kratico:
Archive devnet `stdout`/`stderr` if the workflow fails

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -63,9 +63,7 @@ jobs:
         if: failure()
         with:
           name: cache-target
-          path: |
-            target/**/stderr
-            target/**/stdout
+          path: target
 
   sync:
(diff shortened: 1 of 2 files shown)


.github/workflows/checks.yml
```diff
@@ -122,3 +136,11 @@ jobs:
...
       - run: deno run -A _tasks/use_remote.ts
       - run: deno task test:examples:deno
+      - name: Archive target
+        uses: actions/upload-artifact@v3
+        if: failure()
+        with:
+          name: examples-deno-target
+          path: |
+            target/**/stderr
+            target/**/stdout
```
tjjfvi, line 146:
```suggestion
          path: target
```

tjjfvi (approved):

PR #815, opened 2023-03-27
docs: NFTs pallet example

nythrox:
Resolves #596

PR #878, opened 2023-04-12
feat: add rococo + westmint network

kratico:


harrysolovay (approved):
LGTM

kratico:
> If possible, I think we should try to have as few networks as possible – can you merge this network with the other rococo network we have?

@tjjfvi there is an issue having statemine and westmint in the same network.
I'll start to debug that soon.
In the meantime, this is a workaround for #815

tjjfvi (approved):
Ah, I see

PR #908, opened 2023-04-18
chore: remove usages of approve in multisig examples

ryanleecode:
## Summary

`approveAsMulti` isn't required for an E2E mutlisig flow as it doesn't count towards the number of approvers defined by the threshold. Its a soft approval where it will show up on chain but is actually meaningless. This change makes our examples only use ratify so developers won't get confused about the difference between `ratify` and  `approve`.

https://wiki.polkadot.network/docs/learn-account-multisig#example-using-multi-signature-accounts


tjjfvi:
Can this be closed in favor of #914?

PR #909, opened 2023-04-18
refactor: make dev users non-chain-specific

harrysolovay:
This change exposes the `createUsers` util through `capi` instead of through chain-specific codegen (the dev users are not chain-specific, after all). Also, given that we use the term "dev" in the context of dev nets, let's do the same for test users. The creation util is now exposed through `capi` itself, so let's rename `createUsers` to `createDevUsers`; hopefully this makes clear that its not to be used outside of the context of Capi devnets.


devnets/devnets_env.ts
tjjfvi, line 1: I think this should be renamed to devnetsUrl.ts

examples/dev/test_users.eg.ts
tjjfvi, line 1: Should be renamed

server/codegenHandler.ts
```diff
@@ -212,30 +212,16 @@ export function createCodegenHandler(dataCache: CacheBase, tempCache: CacheBase)
...
-import * as C from "./capi.js"
+    `import * as C from "./capi.js"
 
 export const connect = C.${connection.type}.bind(${JSON.stringify(connection.discovery)})
-
-${
-      connection.type === "DevnetConnection"
-        ? `export const createUsers = C.testUserFactory(${JSON.stringify(connection.discovery)})`
-        : ""
-    }
```
tjjfvi, line 224: Yay

kratico (dismissed):
LGTM

tjjfvi (approved):

PR #911, opened 2023-04-18
feat: add startNetworkForMetadata

kratico:


devnets/startNetwork.ts
```diff
@@ -17,6 +17,59 @@ export interface NetworkChain {
...
+          const binary = await resolveBinary(config.binary, signal)
+          const chain = await spawnChain(
+            path.join(tempDir, name),
+            binary,
+            await createCustomChainSpec(
+              path.join(tempDir, name),
+              binary,
+              config.chain,
+              () => {},
+            ),
```
tjjfvi, line 54: I think this can just be
```suggestion
            config.chain,
```

devnets/startNetwork.ts
```diff
@@ -17,6 +17,59 @@ export interface NetworkChain {
...
+      path.join(tempDir, "relay"),
+      relayBinary,
+      relaySpec,
+      1,
+      [],
+      relayBinary,
+      signal,
+    ),
+    paras: Object.fromEntries(
+      await Promise.all(
```
tjjfvi, line 43: Since the paras no longer need to be spawned after the relay, these two `await`ed functions should be done in parallel

tjjfvi (changes requested):

devnets/startNetwork.ts
kratico: we need the raw chain spec without the boot nodes.
Nevertheless, I will add a `createRawChainSpec` that doesn't create the intermediate chain spec to customize the genesis.

tjjfvi (approved):
:tada:

tjjfvi:
@kratico Anything else you want to do before merge?

PR #912, opened 2023-04-18
fix: lazy shims

tjjfvi:
Should fix capi in the browser.

### DIFF
--- a/_tasks/dnt.ts
+++ b/_tasks/dnt.ts
@@ -102,6 +102,7 @@ await Promise.all([
         name: "ws",
         version: "8.13.0",
       },
+      "./deps/shims/shim-deno.ts": "@deno/shim-deno",
       "./deps/shims/upgradeWebSocket.ts": "./deps/shims/upgradeWebSocket.node.ts",
       "./deps/shims/register.ts": "./deps/shims/register.node.ts",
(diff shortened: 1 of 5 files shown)


ryanleecode (approved):

_tasks/dnt.ts
```diff
@@ -136,8 +137,21 @@ await Promise.all([
...
-  "target/npm/esm/main.js",
-  (await Deno.readTextFile("target/npm/esm/main.js"))
-    .replace(/^#!.+/, "#!/usr/bin/env -S node --loader ts-node/esm"),
-)
+await Promise.all([
+  editFile(
+    "target/npm/esm/main.js",
+    (content) =>
+      content
+        .replace(/^#!.+/, "#!/usr/bin/env -S node --loader ts-node/esm"),
```
harrysolovay, line 145: Should we add `ts-node` as a peer dependency?

tjjfvi: I'm not sure if there's a way to add a peer dev dependency, which is what we'd want here.

tjjfvi: Yes, but I believe that will force downstream users to add `ts-node` as a non-dev dependency, when it should be a dev dependency.

harrysolovay (approved):

PR #919, opened 2023-04-21
refactor: rework FRAME codegen

### LINKED ISSUE #901: codegen pallet bindings on `ChainRune` subtype
We currently codegen pallet bindings, which are accessible from the root mod. Ie.

```ts
import { Balances } from "@capi/my-chain"

const call = Balances.transfer({ /* ... */ })
```

There's an issue with this however: we may want to specify the discovery value / how to connect at runtime.

---

We can do this if we utilize `chain`.

```ts
import { chain as chain_ } from "@capi/my-chain"
import { SmoldotConnection } from "capi"

const chain = _chain.with(SmoldotConnection.bind({ relayChainSpec }))

const call = chain.extrinsic(Rune.object({
  type: "Balances",
  value: Rune.object({
    type: "transfer",
    // ...
  })
}))
```

This experience is different from that of utilizing the generated bindings. It's not the "happy path."

---

Alternatively, we could expose a means of mutating the internally-held connection state.

```ts
import { setConnection } from "@capi/my-chain"
import { SmoldotConnection } from "capi"

setConnection(SmoldotConnection.bind({ relayChainSpec }))
```

However, this too is not ideal.

---

Per an offline Capi team convo yesterday, we're thinking of changing the look of the codegen, such that the experience would be as follows.

```ts
import { chain } from "@capi/my-chain"

const call = chain.Balances.transfer({ /* ... */ })
```

There's an issue with this: what if a pallet name conflicts with that of a property of chain? Perhaps...

```ts
import { chain } from "@capi/my-chain"

const call = chain.pallets.Balances.transfer({ /* ... */ })
```

I'd personally prefer not to place pallet bindings under yet another prop. That being said, I'm unsure of whether there are better options.

---

One more thing to note: this approach would be consumable within pattern libraries, which is not true of the current pallet bindings.

---

@kratico @ryanleecode @tjjfvi –– thoughts?

harrysolovay:
Resolves #901


codegen/frameCodegen.ts
```diff
@@ -0,0 +1,199 @@
...
+
+  files.set("mod.d.ts", MOD)
+  files.set("mod.js", MOD)
+
+  files.set(
+    "metadata.d.ts",
+    `
+      ${IMPORTS_COMMON}
+      export const metadata: metadata
+      export type metadata = ${typeCodegen.print(metadata)}
```
tjjfvi, line 27:
```suggestion
      export const metadata: ${typeCodegen.print(metadata)}
```

codegen/frameCodegen.ts
```diff
@@ -0,0 +1,199 @@
...
+    `,
+  )
+
+  files.set(
+    "chain.d.ts",
+    `
+      ${IMPORTS_COMMON}
+      import { metadata } from "./metadata.js"
+
+      export interface ${chainIdent} extends C.Chain<metadata> {}
```
tjjfvi, line 145:
```suggestion
      export interface ${chainIdent} extends C.Chain<typeof metadata> {}
```

examples/watch.eg.ts
```diff
@@ -6,18 +6,18 @@
...
 /// can be treated as reactive and (B) is a dependent of a "timeline" associated
 /// with Polkadot's block production.
-const now = Timestamp.Now.value(undefined, chain.latestBlockHash)
-
-let i = 0
+const now = chain.Timestamp.Now.value(undefined, chain.latestBlockHash)
 
 /// Use the `watch` method to retrieve an async iterable, which will
 /// gather and yield the `collection`-described data upon new blocks.
+let i = 0
```
tjjfvi, line 19: Perhaps here we should use `hide-next-line` to hide the `i`-related logic

tjjfvi (changes requested):
For all the other codegen, we don't indent the template tags, e.g.
```ts
function foo() {
  return `
console.log("Hello World")
`
}
```
Even though the code is formatted by dprint, it's nice for it to be less adhoc-ly indented before formatting. Can you update `frameCodegen.ts` to follow this convention?

harrysolovay:
> For all the other codegen, we don't indent the template tags

I think it helps with legibility. Certainly makes it easier for newcomers.

codegen/frameCodegen.ts
```diff
@@ -0,0 +1,199 @@
...
+
+      export class ${chainIdent}ChainRune<U> extends C.ChainRune<${chainIdent}, U> {
+        static override from(connect: (signal: AbortSignal) => C.Connection): ${chainIdent}Rune
```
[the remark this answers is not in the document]
harrysolovay: Nice catch

examples/watch.eg.ts
harrysolovay: Moved it to the line before instead and added a comment about its purpose.

tjjfvi:
Unlike the rest of our formatting, though, dprint doesn't handle it, and so the indentation can get out of sync with the surrounding code.

codegen/frameCodegen.ts
tjjfvi: I believe there's only one place where we do this, and `typeof metadata` seems much more semantic than having a lower-cased `metadata` type.

codegen/frameCodegen.ts
```diff
@@ -0,0 +1,199 @@
...
+  files.set(
+    "chain.d.ts",
+    `
+      ${IMPORTS_COMMON}
+      import { metadata } from "./metadata.js"
+
+      export interface ${chainIdent} extends C.Chain<metadata> {}
+
+      export class ${chainIdent}ChainRune<U> extends C.ChainRune<${chainIdent}, U> {
+        static override from(connect: (signal: AbortSignal) => C.Connection): ${chainIdent}ChainRune<U>
```
tjjfvi, line 148: `U` isn't defined at the static level
```suggestion
        static override from(connect: (signal: AbortSignal) => C.Connection): ${chainIdent}ChainRune<never>
```

tjjfvi (approved):

Issue #624, opened 2023-02-21
runions


harrysolovay:
`UnionRune` misc,

```ts
declare const result: {
  ok: string
  error?: never
} | {
  ok?: never
  error: number
}

const value = Rune
  .constant(result)
  .into(UnionRune)
  .discriminate((value) => {
    if (value.ok) return Rune.constant("YES")
    return Rune.constant("NO")
  })
```

`TaggedUnionRune`

```ts
declare const tagged: {
  type: "a"
  value: string
} | {
  type: "b"
  value: number
} | {
  type: "c"
  value: boolean
}

const value = Rune
  .constant(u)
  .into(TaggedUnionRune)
  .match("type", {
    a({ value }) {},
    b({ value }) {},
    c({ value }) {},
  })
```

PR #930, opened 2023-04-25
feat: improve ink! contract instantiate

### LINKED ISSUE #932: Instantiate Ink! contract DX
The current `InkMetadataRune` implementation supports instantiating a contract by requiring the user to provide the contract metadata and code.
However, if the code has already been deployed, the contract could be instantiated with the contract code hash (which is in the metadata).
Finally, a contract could also be instantiated just with a `<contract>.contract` file,  which is the metadata file with the hex encoded contract code in a `.wasm` attribute. 

Given the above scenarios, What would be a better DX?
- make `code` optional in `InkMetadataRune.instantiation(...)`
- throw an error if `code` is not provided and the contract code hash is not found in the deployed contracts
- add a `InkMetadataRune.instantiationWithCode(...)`
- other

Bellow is a list of scenarios to cover

| Contract deployed? | Metadata Format | Code provided? | Result | Preferred Pallet Contract Call
|--------|--------|--------|--------|--------|
| ❌  | `.json` | ❌  | Error | N/A |
| ❌  | `.json` | ✅  | ✅  | `.instantiateWithCode(...)` |
| ❌  | `.contract` | ❌  | ✅  | `.instantiateWithCode(...)` |
| ❌  | `.contract` | ✅   | ✅  | `.instantiateWithCode(...)` |
| ✅   | `.json` | ❌  | ✅  | `.instantiate(...)` |
| ✅   | `.json` | ✅  | ✅  | `.instantiate(...)` |
| ✅   | `.contract` | ❌  | ✅  | `.instantiate(...)` |
| ✅   | `.contract` | ✅   | ✅  | `.instantiate(...)` |

Note: if `.instantiateWithCode(...)` is called and the contract code was already deployed, the end result is the same but it looks to be less efficient because sending the code for an already deployed contract is unnecessary.

kratico:
Resolves #932


patterns/ink/InkMetadataRune.ts
```diff
@@ -89,7 +89,30 @@ export class InkMetadataRune<out U> extends Rune<InkMetadata, U> {
...
+    //     code
+    //       ? sourceHash
+    //       : chain.pallet("Contracts").storage("CodeStorage").valueRaw(
+    //         sourceHash as $.Native<Chain.Storage<C, "Contracts", "CodeStorage">["key"]>,
+    //       )
+    //         .unhandle(undefined)
+    //         .map(() => sourceHash)
+    //         .rehandle(undefined, () => Rune.constant(new NotDeployedError()))
+    //         .unhandle(NotDeployedError)
+    //   )
```
kratico, line 115: @tjjfvi I'd like to throw a `NotDeployedError` if the contract has not been deployed.
How would you branch this logic with `Rune`s?
Right now, the logic assumes that there is a contract deployed for the given source hash.

harrysolovay:
We'll revisit #932 after #624 and #811.

PR #935, opened 2023-04-26
feat: add devnet custom genesis, aura and grandpa support

### LINKED ISSUE #889: Devnet add support configure pallet_session, pallet_aura and pallet_grandpa at genesis
Not every runtime uses `pallet_session`, some runtimes use `pallet_aura` and `pallet_grandpa`.
Then, some runtimes configure `pallet_session` under `session`, `palletSession` or `authorMapping`.
Furthermore, `pallet_session` is generic over the `SessionKeys`.
For example
```rust
// rococo
impl_opaque_keys! {
	pub struct SessionKeys {
		pub grandpa: Grandpa,
		pub babe: Babe,
		pub im_online: ImOnline,
		pub para_validator: Initializer,
		pub para_assignment: ParaSessionInfo,
		pub authority_discovery: AuthorityDiscovery,
		pub beefy: Beefy,
	}
}

// polkadot
impl_opaque_keys! {
	pub struct SessionKeys {
		pub grandpa: Grandpa,
		pub babe: Babe,
		pub im_online: ImOnline,
		pub para_validator: Initializer,
		pub para_assignment: ParaSessionInfo,
		pub authority_discovery: AuthorityDiscovery,
	}
}
```

Shall we leave it as is?

kratico:
Resolves #889 

**How to test?**

- for `aura` and `grandpa`, compile substrate `node-template`
- execute the following script
```typescript
import { bins, net } from "./mod.ts"

const controller = new AbortController()

Deno.addSignalListener("SIGTERM", () => controller.abort())
Deno.addSignalListener("SIGINT", () => controller.abort())

const bin = bins({
  nodeTemplate: "/Users/parity/ghq/github.com/paritytech/substrate/target/release/node-template",
})

const relay = net.dev({
  bin: bin.nodeTemplate,
  chain: "local",
  customize(chainSpec) {
    // @ts-ignore
    chainSpec.genesis.runtime.sudo.key = "5FHneW46xGXgs5mUiveU4sbTyGBzmstUspZC92UhjJM694ty"
  },
})
relay.name = "relay"

console.log(await relay.spawnNet(controller.signal, await Deno.makeTempDirSync({ dir: "./tmp" })))
```
- check the logs `./tmp/cbfda8c3/relay/alice/stderr` and validate that blocks are produced and finalized
- check the sudo key is `5FHneW46xGXgs5mUiveU4sbTyGBzmstUspZC92UhjJM694ty` (or `await chain.Sudo.Key.value()`)


kratico:
@harrysolovay @tjjfvi 
Given that the pallet name could be aliased to any string, Shall I implement an option to pass an optional genesis config in `DevChainConfig`?
This genesis config will be merged with chain spec genesis config.

For example, Moonbean manages consensus with their `pallet-author-mapping` and sets an [`AuthorMapping`](https://github.com/PureStake/moonbeam/blob/0933fe3db3a658383e9bbe433c57869bb5955370/runtime/moonbeam/src/lib.rs#L1343) attribute in `construct_rutime`.
In the `capi.config.ts`, we could have something like
```
{
  binary: ...,
  chain: "...",
  genesis: { ... },
}
```

harrysolovay:
That seems like a good path, but not necessarily expressing it in the TS. What's the most likely user story? Given that devs will generate the chainspec for their target chain, maybe they could pass the entire chainspec str, from which we'd extract the genesis config? Any simpler way to go about this?

kratico:
> That seems like a good path, but not necessarily expressing it in the TS. What's the most likely user story? Given that devs will generate the chainspec for their target chain, maybe they could pass the entire chainspec str, from which we'd extract the genesis config? Any simpler way to go about this?

For this story, I would just merge a custom genesis with the default genesis generated from the binary chain spec (for example, `polkadot build-spec --chain rococo-local`).

On the other hand, if devs pass a chainspec, I would just use that chainspec, add the bootnodes and use the genesis from that chainspec. For this, I would assume that the dev has already customized the genesis for their needs.

devnets/Config.ts
```diff
@@ -25,11 +25,13 @@ export interface DevChainConfig {
 
   chain: string
   nodes?: number
+  genesis?: Record<string, unknown>
```
tjjfvi, line 28: I don't think this should be configurable in the config unless we have a really solid use-case (that can't be solved by something else)

nets/DevRelaySpec.ts
```diff
@@ -39,15 +39,16 @@ export class DevRelaySpec extends DevNetSpec {
     return createCustomChainSpec(tempDir, binary, this.chain, (chainSpec) => {
       const genesisConfig = getGenesisConfig(chainSpec)
       if (parachainInfo.length) {
-        genesisConfig.paras.paras.push(
-          ...parachainInfo.map((
-            { id, genesis },
-          ): GenesisConfig["paras"]["paras"][number] => [id, [...genesis, true]]),
+        genesisConfig.paras?.paras.push(
+          ...parachainInfo.map(({ id, genesis: [state, wasm] }) =>
```
harrysolovay, line 43: - Can we add an explicit return type and remove the `as NonNullable<GenesisConfig["paras"]>["paras"][number]`?
- Can we expose this type as named from within `DevNetSpec`? `ChainSpecGenesisConfigParasEntry` perhaps?

nets/chain_spec/addAuthorities.ts
```diff
@@ -1,24 +1,31 @@
...
+        babe: srAccount,
+        im_online: srAccount,
+        para_validator: srAccount,
+        para_assignment: srAccount,
+        authority_discovery: srAccount,
+        beefy: ecAccount,
+      },
+    ])
+  } else if (genesisConfig.aura && genesisConfig.grandpa) {
+    genesisConfig.aura.authorities = authorities.slice(0, count).map(({ srAccount }) => srAccount)
```
harrysolovay, line 24: Let's extract `authorities.slice(0, count)` to a preceding local & reuse it in the following two lines.

harrysolovay (changes requested):

nets/DevRelaySpec.ts
tjjfvi: You could just write `[id, [state, wasm, true]] satisfies Narrow`

tjjfvi (approved):

harrysolovay (approved):

PR #936, opened 2023-04-27
feat: use polkadot/polkadot-parachain patched

kratico:
Companion PR https://github.com/paritytech/capi-binary-builds/pull/36

Use the `polkadot` and `polkadot-parachain` patched version with 2s and 4s block times.

image

FYI, I did see an improvement in individual example runs but I did not see an improvement in the `deno task test:eg` timings.

### DIFF
--- a/import_map.json
+++ b/import_map.json
@@ -1,6 +1,6 @@
 {
   "imports": {
-    "@capi/": "http://localhost:4646/01a459230d968c6e/"
+    "@capi/": "http://localhost:4646/c203f477e2a88c37/"
   },
   "scopes": {
     "examples/": {
--- a/nets.ts
+++ b/nets.ts
@@ -1,8 +1,8 @@
 import { bins, net } from "./mod.ts"
 
 const bin = bins({
-  polkadot: ["polkadot", "v0.9.38"],
-  polkadotParachain: ["polkadot-parachain", "v0.9.380"],
+  polkadot: ["polkadot-fast", "v0.9.38"],
+  polkadotParachain: ["polkadot-parachain-fast", "v0.9.380"],
   substrateContractsNode: ["substrate-contracts-node", "v0.24.0"],
   trappistPolkadot: ["polkadot", "v0.9.37"],
   trappistPolkadotParachain: ["polkadot-parachain", "v0.9.370"],


capi.config.ts
```diff
@@ -1,7 +1,7 @@
 import { binary, Config } from "./mod.ts"
 
-const polkadot = binary("polkadot", "v0.9.38")
-const polkadotParachain = binary("polkadot-parachain", "v0.9.380")
+const polkadot = binary("polkadot", "v0.9.38-patched")
+const polkadotParachain = binary("polkadot-parachain", "v0.9.380-patched")
```
kratico, line 4: `v0.9.38-dev` might be better

WDYT?

kratico:
> Any reason not to patch with 1s and 1s instead of 2s and 4s?

I did a couple of tests and there is no improvement with smaller values.
I guess, it's because there are more forks with lower values and it takes longer to finalize blocks.
Nevertheless, I think that babe/aura/grandpa pallets could be tweaked to get shorter block times.

harrysolovay:
13s is still a win! Not quite instantaneous, but hopefully there's a path to get there.

server/devnetsHandler.ts
```diff
@@ -7,7 +7,7 @@ import * as f from "./factories.ts"
 const rDevnetsApi = /^\/devnets\/([\w-]+)$/
 
 export function createDevnetsHandler(
-  devnetTempDir: string,
+  devnetsTempDir: string,
```
harrysolovay, line 10:
```suggestion
  devnetTempDir: string,
```

server/devnetsHandler.ts
```diff
@@ -31,7 +31,7 @@ export function createDevnetsHandler(
     if (net instanceof DevNetSpec) {
       const network = await networkMemo.run(
         net.relay,
-        () => net.relay.spawnNet(signal, devnetTempDir),
+        () => net.relay.spawnNet(signal, devnetsTempDir),
```
harrysolovay, line 34:
```suggestion
        () => net.relay.spawnNet(signal, devnetTempDir),
```

harrysolovay (approved):

tjjfvi (approved):

PR #944, opened 2023-04-28
chore: fix deno workflow errors

ryanleecode:
Pins the deno version to `1.32.5` because 1.33 has introduced many breaking changing and we are unable to pinpoint whether workkflow errors are because of capi or because of a bug in the latest upgrade.

Added an issue to keep track of when we need to upgrade: https://github.com/paritytech/capi/issues/945

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -26,7 +26,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: denoland/setup-deno@9db7f66e8e16b5699a514448ce994936c63f0d54 # v1.1.0
         with:
-          deno-version: v1.x
+          deno-version: v1.32.x
       - run: deno lint
 
   spellcheck:
@@ -45,7 +45,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: denoland/setup-deno@9db7f66e8e16b5699a514448ce994936c63f0d54 # v1.1.0
         with:
-          deno-version: v1.x
+          deno-version: v1.32.x
       - uses: actions/cache@v3
         with:
           path: ~/.cache/deno
@@ -76,7 +76,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: denoland/setup-deno@9db7f66e8e16b5699a514448ce994936c63f0d54 # v1.1.0
         with:
-          deno-version: v1.x
+          deno-version: v1.32.x
       - run: deno task sync --check
       - name: Archive target
         uses: actions/upload-artifact@v3
@@ -97,7 +97,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: denoland/setup-deno@9db7f66e8e16b5699a514448ce994936c63f0d54 # v1.1.0
         with:
-          deno-version: v1.x
+          deno-version: v1.32.x
       - uses: actions/cache@v3
         with:
           path: ~/.cache/deno
@@ -116,7 +116,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: denoland/setup-deno@9db7f66e8e16b5699a514448ce994936c63f0d54 # v1.1.0
         with:
-          deno-version: v1.x
+          deno-version: v1.32.x
       - uses: actions/cache@v3
         with:
           path: ~/.cache/deno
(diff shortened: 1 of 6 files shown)


_tasks/udd.ts
```diff
@@ -15,7 +15,8 @@ for await (
 await Promise.all(pending)
 
 async function udd(filePath: string) {
-  await Deno
-    .run({ cmd: ["deno", "run", "-A", "https://deno.land/x/udd@0.8.2/main.ts", filePath] })
-    .status()
+  await new Deno.Command("deno", {
+    args: ["run", "-A", "https://deno.land/x/udd@0.8.2/main.ts", filePath],
+  }).spawn()
+    .status
```
harrysolovay, line 21: Nit: can be on line 20

_tasks/udd.ts
```diff
@@ -15,7 +15,8 @@ for await (
 await Promise.all(pending)
 
 async function udd(filePath: string) {
-  await Deno
-    .run({ cmd: ["deno", "run", "-A", "https://deno.land/x/udd@0.8.2/main.ts", filePath] })
-    .status()
+  await new Deno.Command("deno", {
+    args: ["run", "-A", "https://deno.land/x/udd@0.8.2/main.ts", filePath],
+  }).spawn()
```
kratico, line 20: Given that the child process is not used, I think we just need `.output()`

deps/shims/command.node.ts
```diff
@@ -1,3 +1,4 @@
+// deno-lint-ignore-file
```
kratico, line 1: I'd add `// deno-lint-ignore no-deprecated-deno-api` to the 2 `Deno.run` calls

deps/shims/Deno.node.ts
```diff
@@ -1,3 +1,4 @@
+// deno-lint-ignore-file
```
kratico, line 1: I think this is not needed

.github/workflows/checks.yml
```diff
@@ -133,7 +133,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: denoland/setup-deno@9db7f66e8e16b5699a514448ce994936c63f0d54 # v1.1.0
         with:
-          deno-version: v1.x
+          deno-version: v1.32.x
```
kratico, line 136: I think we only need `1.32.x` only for the examples.
For the rest, we could leave `1.x`

deps/shims/command.node.ts
ryanleecode: Ill get another warning if i do that B/c we still on older version. Maybe we can defer this change then?

`Ignore for code "no-deprecated-deno-api" was not used.deno-lint(ban-unused-ignore)`

.github/workflows/checks.yml
ryanleecode: Lint as well since the deno version on our machines will mismatch the CI

deps/shims/command.node.ts
kratico: apologies, I did my review using `v1.33.x`

_tasks/udd.ts
```diff
@@ -15,7 +15,7 @@ for await (
 await Promise.all(pending)
 
 async function udd(filePath: string) {
-  await Deno
-    .run({ cmd: ["deno", "run", "-A", "https://deno.land/x/udd@0.8.2/main.ts", filePath] })
-    .status()
+  await new Deno.Command("deno", {
+    args: ["run", "-A", "https://deno.land/x/udd@0.8.2/main.ts", filePath],
+  }).spawn().output()
```
kratico, line 20: this should be without `.spawn`

_tasks/udd.ts
```diff
@@ -15,7 +15,7 @@ for await (
 await Promise.all(pending)
 
 async function udd(filePath: string) {
-  await Deno
-    .run({ cmd: ["deno", "run", "-A", "https://deno.land/x/udd@0.8.2/main.ts", filePath] })
-    .status()
+  await new Deno.Command("deno", {
```
kratico, line 18: Nit, `Deno.execPath()` instead of `deno`

kratico (approved):
LGTM 🚀

tjjfvi (approved):

PR #960, opened 2023-05-03
chore: clean up weights + estimates

harrysolovay:
- Renames `estimateFee` to `weight` (which is what it actually is)
- Adds an `estimate` method to `ExtrinsicRune` which actually returns the estimate (a `bigint`)

**NOTE:** the `TransactionPaymentApi_query_weight_to_fee` endpoint is currently Westend-only, but will supposedly make its way into Polkadot (correct @statictype ?).


fluent/ExtrinsicRune.ts
```diff
@@ -64,18 +64,29 @@ export class ExtrinsicRune<out C extends Chain, out U> extends PatternRune<Chain
...
+  weightRaw() {
     const arg = Rune
       .fn(concat)
       .call(this.extrinsic, this.extrinsic.access("length").map((n) => $.u32.encode(n)))
       .map(hex.encodePrefixed)
-    const data = this.chain.connection
+    return this.chain.connection
       .call("state_call", "TransactionPaymentApi_query_info", arg)
       .map(hex.decode)
+      .map((v) => v.slice(0, 6))
```
harrysolovay, line 75: The `TransactionPaymentApi_query_info` endpoint returns a heavily-padded hex str. If we don't strip the excess before passing to other calls, we get a runtime panic.

peetzweg: Odd. Maybe we should add a comment in the code to prevent confusion in the future?

Can we be sure it is always just 7 bytes?

harrysolovay: > Can we be sure it is always just 7 bytes?

Should be consistent (as it's a u128 / 2^7)

kratico (dismissed):
LGTM

examples/tx/weights_&_estimates.ts
```diff
@@ -0,0 +1,33 @@
...
+
+/// Gather the weight and estimate.
+const collection = await Rune
+  .object({
+    weight: call.weight(),
+    estimate: call.estimate(),
+  })
+  .run()
+
+/// Ensure the data is of the expected shape
```
harrysolovay, line 25:
```suggestion
/// Ensure the data is of the expected shape.
```

examples/tx/weights_&_estimates.ts
```diff
@@ -0,0 +1,33 @@
...
+/// Gather the weight and estimate.
+const collection = await Rune
+  .object({
+    weight: call.weight(),
+    estimate: call.estimate(),
+  })
+  .run()
+
+/// Ensure the data is of the expected shape.
+console.log(collection)
```
peetzweg, line 26: Is this meant to be still in the code?

examples/tx/weights_&_estimates.ts
```diff
@@ -0,0 +1,33 @@
+/**
+ * @title Weights & Estimates
```
peetzweg, line 2: Btw never seen an ampersand in a source code filename. Are we sure this will not cause any errors down the line?

`examples/tx/weights_&_estimates.ts`

peetzweg: Never mind, just realized it's an example file.

examples/tx/weights_&_estimates.ts
peetzweg: Never mind, just realized it's an example file.

fluent/ExtrinsicRune.ts
```diff
@@ -59,13 +59,29 @@ export class ExtrinsicRune<out C extends Chain, out U> extends PatternRune<Chain
       .fn(concat)
       .call(extrinsic, extrinsic.access("length").map((n) => $.u32.encode(n)))
       .map(hex.encodePrefixed)
-    const data = this.chain.connection
+    // The TransactionPaymentApi_query_info endpoint returns a heavily-padded hex str.
+    // If we don't strip the excess before passing to other calls, we get a runtime panic.
+    return this.chain.connection
       .call("state_call", "TransactionPaymentApi_query_info", arg)
```
tjjfvi, line 65: (`TransactionPaymentApi_query_info` doesn't directly return a weight, it returns a `RuntimeDispatchInfo`)[https://github.com/paritytech/substrate/blob/master/frame/transaction-payment/rpc/runtime-api/src/lib.rs#L32]

So it's not a "heavily-padded hex str", but rather has other data. It should be decoded with a `$.tuple($weight, $dispatchClass, $balance)`.

The initial `$weight` won't be consistently 6 bytes, as it is `$.tuple($.compact($.u128), $.compact($.u128))`, so the size varies based on the values.

examples/tx/weights_&_estimates.ts
tjjfvi, line 1: Should be `.eg.ts`

tjjfvi: Also, the `&` in the name is odd – I think `weights_and_estimates` would be better.

fluent/ExtrinsicRune.ts
```diff
@@ -59,13 +59,30 @@ export class ExtrinsicRune<out C extends Chain, out U> extends PatternRune<Chain
...
       .call(extrinsic, extrinsic.access("length").map((n) => $.u32.encode(n)))
       .map(hex.encodePrefixed)
-    const data = this.chain.connection
+    const info = this.chain.connection
       .call("state_call", "TransactionPaymentApi_query_info", arg)
       .map(hex.decode)
     return this.chain.metadata
-      .access("paths", "sp_weights::weight_v2::Weight")
-      .map(($c) => $.field("weight", $c))
+      .access("types", "DispatchInfo")
```
tjjfvi, line 66: These `types` are pretty unstable; you should use the corresponding `paths` codec instead.

tjjfvi (approved):

Issue #968, opened 2023-05-04
dnt wrong module and types in package.json


ryanleecode:
Using b3f0c68073e791497850831d043a673390cc70e7 with https://github.com/paritytech/capi/pull/967 applied,

the output package json for dnt is 

```json
"module": "./esm/crypto/Sr25519.js",
"types": "./types/crypto/Sr25519.d.ts",
"bin": {
  "capi": "./esm/main.js"
},
```

when it should be

```json
"module": "./esm/mod.js",
"types": "./types/mod.d.ts",
"bin": {
  "capi": "./esm/main.js"
},
```

kratico:
this should be fixed now

PR #967, opened 2023-05-04
fix: dnt shims

kratico:
Fix `dnt` shims by excluding
- files from `target|_tasks|examples`
- `deps/shims/ws.ts` and `deps/shims/shim-deno.ts` from the entry points

This PR also fixes `capi sync --package-json package.json` which is needed for https://github.com/paritytech/capi-vite-example/pull/1

PR #969, opened 2023-05-05
fix: dnt package json

ryanleecode:
temporary work around for https://github.com/paritytech/capi/issues/968 while we investigate root cause.


kratico:
Thanks @ryanleecode 
This should be fixed with these lines https://github.com/paritytech/capi/pull/967/files#diff-7293772ad7f24dfa2389450fb5b78ae7c9d76bb8b3f02be62c92a77f22eed9e3R76-R79

apparently, the first entry point will be set as `"module"` in the `package.json`

ryanleecode:
Yes that seems to have fixed it. Well done :smiley_cat:

PR #983, opened 2023-05-10
feat: net target / connection specificity

### LINKED ISSUE #925: devnet <-> prod net toggling story
How will we allow both Node and Deno users to easily switch––without touching their code––between targeting a development vs. production network?

Currently, one needs to update their code (specifically import specifiers, the chain rune subtype instance name (as of #923) and any references to that name).

```diff
- import { Polkadot } from "@capi/polkadot"
+ import { PolkadotDev } from "@capi/polkadot-dev"

- Polkadot.System
+ PolkadotDev.System
```

This isn't the smoothest of development stories.

I propose that we allow specificity of `targets` in top-level `NetworkConfig`s. This would often be used to pair production networks with their devnet counterparts.

```ts
import { binary, Config, NetworkConfig } from "capi"

const polkadotDev: NetworkConfig = {
  binary: binary("polkadot", "v0.9.40"),
  chain: "polkadot-dev",
}

export const config: Config = {
  server: "https://capi.dev/",
  networks: {
    polkadotDev,
    polkadot: {
      url: "wss://rpc.polkadot.io/",
      version: "v0.9.40",
      targets: {
        dev: polkadotDev,
      }
    },
  },
}
```

After syncing this config, one could pass `--target dev` to Capi's `serve` command. This would tell the codegen to produce a `connection.js` reflecting the `DevConnection`.

This would also open the door for us to validate prod-dev chain feature parity. We could produce console warnings about inconsistencies between the default and target-listed metadata.

harrysolovay:
Resolves #925

Enables hot-swapping of underlying connection via `nets.ts` setup + `--target` arg passed to serve.

```diff
export const polkadot = net.ws({
  url: "wss://rpc.polkadot.io/",
+ targets: { dev: polkadotDev },
})
```

Then...

```
capi serve --target dev -- deno run -A ./my_script.ts
```


rune/ValueRune.ts
```diff
@@ -18,7 +18,7 @@ type Access<T, K extends keyof T, _K = NonIndexSignatureKeys<T>> = [
 /** @ts-ignore: assume it's a valid key */
 type GetPath<T, P> = P extends [infer K, ...infer Q] ? GetPath<T[K], Q> : T
 
-type EnsurePath<T, P> = never extends P ? P extends [infer K, ...infer Q] ? 
+type EnsurePath<T, P> = never extends P ? P extends [infer K, ...infer Q] ?
```
harrysolovay, line 21: I think this has to do with a dprint plugin update

Readme.md
```diff
@@ -100,30 +103,19 @@ const accounts = await polkadot.System.Account.entries({ count: 10 }).run()
...
-    .run()
-```
-
-To run code that depends on a devnet, use the `serve` command, followed by a
-"--" and your devnet-using command. In this case, we'll run the script
-(`main.js`) with Node.JS.
+During development, we may want to swap out the underlying connection with that
+of a devnet. This can be achieved with the `serve` command, followed by target
+specificity (in our case `dev`), a "--" and your devnet-using command. In this
+case, we'll run the script (`main.js`) with Node.JS.
```
tjjfvi, line 109:
```suggestion
During development, we may want to swap out the underlying connection with that
of a devnet. This can be achieved via targets — by specifying alternate targets
in your `nets.ts` file, you can switch to them by wrapping your command with
`capi serve --target someTarget --`. For example:
```

tjjfvi, line 109:
```suggestion
During development, we may want to swap out the underlying connection with that
of a devnet. This can be achieved via targets — by specifying alternate targets
in your `nets.ts` file, you can switch to them by wrapping your command with
`capi serve --target someTarget --`. For example:
```

server/CodegenSpec.ts
```diff
@@ -8,6 +8,7 @@ const $codegenEntry = $.taggedUnion("type", [
     $.field("metadataHash", $.sizedUint8Array(64)),
     $.field("chainName", $.str),
     $.optionalField("connection", $connectionSpec),
```
tjjfvi, line 10: This should no longer be optional
```suggestion
    $.field("connection", $connectionSpec),
```

tjjfvi (changes requested):

Readme.md
```diff
@@ -43,14 +43,17 @@ import { bins, net } from "capi"
...
 // A Polkadot development network
 export const polkadotDev = net.dev({
   bin: bin.polkadot,
   chain: "polkadot-dev",
 })
+
+// The Polkadot relay chain
+export const polkadot = net.ws({
+  url: "wss://rpc.polkadot.io/",
+  targets: { dev: polkadotDev },
```
ryanleecode, line 55: should we do this in the capi `nets.ts` file?

server/client/detectConnect.ts
```diff
@@ -0,0 +1,20 @@
...
+    const target = targets[targetName]
+    if (target) spec = target
+  }
+  switch (spec.type) {
+    case "WsConnection":
+      return WsConnection.bind(spec.discovery)
+    case "DevnetConnection":
+      return DevnetConnection.bind(spec.discovery)
+  }
+}
```
ryanleecode, line 20: Not a fan of parameter reassignment so something like this could work better.

```suggestion
export function detectConnect(
  spec: ConnectionSpec,
  targets: Record<string, ConnectionSpec>,
): Connect {
  const targetName = Deno.env.get("CAPI_TARGET") ?? ""
  return bindConnection(targets[targetName] ?? spec)
}

function bindConnection(spec: ConnectionSpec) {
  switch (spec.type) {
    case "WsConnection":
      return WsConnection.bind(spec.discovery)
    case "DevnetConnection":
      return DevnetConnection.bind(spec.discovery)
  }
}
```

ryanleecode (changes requested):

Readme.md
harrysolovay: No, as it would require separate server instances for differently-targeted examples.

ryanleecode (dismissed):

nets.ts
```diff
@@ -59,7 +59,3 @@ export const rococoDevXcmStatemine = rococoDevXcm.parachain({
   chain: "statemine-local",
   id: 1000,
 })
-
-export const polkadotFromMetadata = net.metadata({
-  metadata: await Deno.readFile("examples/raw_rpc/metadata"),
```
tjjfvi, line 64: We can delete this file now, right?

tjjfvi (approved):

PR #997, opened 2023-05-12
chore: update examples with unhandleFailed

ryanleecode:


harrysolovay:
We'll tackle the error-handling story more holistically following #624. Let's close this for now.

PR #999, opened 2023-05-14
fix: default to finalized block hash

### LINKED ISSUE #239: Defaulting to the best block is insecure
I see several examples showing that it is possible to call `read()` in order to read a storage value from, as far as I understand, the best block.

If the JSON-RPC server is a light node, then this is completely insecure. **A light client doesn't offer any guarantee about the validity of the best block**. Only the finalized block can be trusted. This is a fundamental property of Substrate-based light clients and will not change in the future.

For this reason, I don't think that it is a good idea to default to reading from the best block.

However, users might still be interested in reading a storage value or watching a storage value of the best block in order to show the changes that will likely be part of the finalized block.
In other words, a storage value in the best block should be seen as "pending changes that might end up not being applied" while storage values in the finalized block should be seen as "has actually made it into the chain".

harrysolovay:
Fixes #239

### DIFF
--- a/fluent/ChainRune.ts
+++ b/fluent/ChainRune.ts
@@ -87,8 +87,7 @@ export class ChainRune<out C extends Chain, out U> extends Rune<C, U> {
   blockHash<X>(...[blockHash]: RunicArgs<X, [blockHash?: string]>) {
     return Rune
       .resolve(blockHash)
-      .unhandle(undefined)
-      .rehandle(undefined, () => this.connection.call("chain_getFinalizedHead"))
+      .handle(undefined, () => this.connection.call("chain_getFinalizedHead"))
       .into(BlockHashRune, this)
   }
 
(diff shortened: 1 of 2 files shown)


fluent/ChainRune.ts
```diff
@@ -84,11 +84,10 @@ export class ChainRune<out C extends Chain, out U> extends Rune<C, U> {
     .unsafeAs<string>()
     .into(BlockHashRune, this)
 
-  blockHash<X>(...[blockHash]: RunicArgs<X, [blockHash?: string]>) {
+  resolveBlockHash<X>(...[blockHash]: RunicArgs<X, [blockHash?: string]>) {
```
harrysolovay, line 87:
```suggestion
  blockHash<X>(...[blockHash]: RunicArgs<X, [blockHash?: string]>) {
```

tjjfvi (approved):

PR #1011, opened 2023-05-16
chore: CLI detect version

### LINKED ISSUE #859: auto-detect cli version and use it to inform default for server

harrysolovay:
Resolves #859


main.ts
```diff
@@ -1,14 +1,25 @@
...
   .command("bin", bin)
   .command("sync", sync)
   .command("serve", serve)
   .parse(Deno.args)
+
+function version() {
+  try {
+    return detectVersion()
+  } catch (_e) {
+    return new TextDecoder().decode(
```
ryanleecode, line 21: so this would show up as `main`?

server/detectVersion.ts
```diff
@@ -0,0 +1,9 @@
+const rCapiVersion = /^https:\/\/(?:capi.dev\/|deno.land\/x\/capi)@(.+?)\//
+
+export function detectVersion(): string {
+  const version = rCapiVersion.exec(import.meta.url)?.[0]
+  if (!version) {
+    throw new Error("Could not detect version from URL. Please specify via `--version` flag.")
```
ryanleecode, line 6: if the catch block in `version` doesn't make use of these errors arent the error messages here moot?

main.ts
harrysolovay: If you run it in a clone of main, yea.

server/detectVersion.ts
harrysolovay: No, we want this error message to occur in the case of running the sync command of a non-dist Capi without `--version`

cli/resolveNets.ts
```diff
@@ -5,26 +5,24 @@ import { NetSpec } from "../nets/mod.ts"
...
-    for (const p of ["nets.ts", "nets.js"]) {
+  const resolvedNetsPath = await resolveNetsPath(maybeNetsPath)
+  const nets = await import(path.toFileUrl(resolvedNetsPath).toString())
+  $.assert($nets, nets)
+  for (const key in nets) nets[key]!.name = key
+  return nets
+}
+
+async function resolveNetsPath(maybeNetsPath?: string): Promise<string> {
+  for (const p of [maybeNetsPath, "nets.ts", "nets.js"]) {
```
tjjfvi, line 16: If the nets path is specified, we shouldn't also check `nets.ts` and `nets.js`

server/detectVersion.ts
```diff
@@ -0,0 +1,9 @@
+const rCapiVersion = /^https:\/\/(?:capi.dev\/|deno.land\/x\/capi)@(.+?)\//
+
+export function detectVersion(): string {
+  const version = rCapiVersion.exec(import.meta.url)?.[1]
+  if (!version) {
+    throw new Error("Could not detect version from URL. Please specify via `--version` flag.")
```
tjjfvi, line 6: This should be handled downstream, in the sync command; rather than throwing here, it will probably be nicer to return undefined. Also, there is no `--version` flag.

main.ts
tjjfvi, line 21: This should only be done in deno, and if `import.meta.url` is a file url. I.e., it should be moved to `detectVersion`.

main.ts
```diff
@@ -1,14 +1,25 @@
...
+      new Deno.Command("git", { args: ["branch", "--show-current"] }).outputSync().stdout,
```
tjjfvi, line 22: This should be the current abbreviated sha.

tjjfvi (changes requested):

cli/sync.ts
```diff
@@ -39,7 +40,13 @@ async function runSync({
 }: RunSyncOptions, runtime: string) {
   const netSpecs = await resolveNets(netsFile)
   const devnetTempDir = await tempDir(out, "devnet")
-  const baseUrl = await syncNets(server, devnetTempDir, netSpecs)
+  const version = detectVersion()
+  if (!version) throw new Error("Please re-run `sync` with \`--server\` specified.")
```
tjjfvi, line 44:
```suggestion
  if (!server && !version) throw new Error("Could not auto-detect version; please re-run `sync` with \`--server\` specified.")
```

server/detectVersion.ts
```diff
@@ -0,0 +1,11 @@
+const rCapiVersion = /^https:\/\/(?:capi.dev\/|deno.land\/x\/capi)@(.+?)\//
+
+export function detectVersion(): string | undefined {
+  const match = rCapiVersion.exec(import.meta.url)?.[1]
+  if (match) return match
+  const sha = new TextDecoder().decode(
+    new Deno.Command("git", { args: ["rev-parse", "HEAD"] }).outputSync().stdout,
+  )
+  if (sha.length) return sha.slice(0, 8)
```
tjjfvi, line 9:
```suggestion
  if (new URL(import.meta.url).protocol === "file:") {
    const sha = new TextDecoder().decode(
      new Deno.Command("git", { args: ["rev-parse", "HEAD"] 
 }).outputSync().stdout,
    )
    if (sha.length) return sha.slice(0, 8)
  }
```

tjjfvi (changes requested):

server/detectVersion.ts
```diff
@@ -0,0 +1,13 @@
+const rCapiVersion = /^https:\/\/(?:capi.dev\/|deno.land\/x\/capi)@(.+?)\//
+
+export function detectVersion(): string | undefined {
+  const match = rCapiVersion.exec(import.meta.url)?.[1]
+  if (match) return match
+  if (new URL(import.meta.url).protocol === "file:") {
+    const sha = new TextDecoder().decode(
+      new Deno.Command("git", { args: ["rev-parse", "HEAD"] }).outputSync().stdout,
```
tjjfvi, line 8: The cwd here should be set based on `import.meta.url`

tjjfvi (approved):

PR #1013, opened 2023-05-17
chore: rename page-retrieval `count` to `limit`

harrysolovay:
"Limit" more accurately describes the prop.


tjjfvi (approved):

ryanleecode (changes requested):
this needs to be changed too or no?

https://github.com/paritytech/capi/blob/c6730261db3015cd9df7a790d2458c11b3563e20/rpc/known/childstate.ts#L33

ryanleecode (approved):

PR #1019, opened 2023-05-18
feat: optional stash for multisigrune

ryanleecode:
Adds an optional stash for Multisigrune


tjjfvi (commented):
Can you clarify what this is for?

ryanleecode:
> Can you clarify what this is for?

Apps team wanted this for thier MVP flow

harrysolovay:
It seems this change is primarily about the state codec. Instead of altering `$multisig`/`Multisig`, it'd be better for users to define and utilize a new codec. For instance...

```ts
$.tuple($multisig, $.sizedUint8Array(32))
```

PR #1025, opened 2023-05-19
feat: CLI `init` command

### LINKED ISSUE #940: CAPI CLI DX
Improve CLI DX by supporting
- extra args, like `--help`
- subcommands help
- subcommands args help
- other...

harrysolovay:
Resolves #940

In a Deno project, simply do the following.

```sh
deno run -A https://deno.land/x/capi/main.ts init
```

Or in a Node project

```sh
npm i capi
npx capi init
npm run capi:sync
```

**How to test this locally with Node?**

```sh
# in Capi
deno task dnt
cd target/npm
npm pack
# in a Node project
npm i /path/to/target/npm/capi-v0.0.0-local.tgz
npx capi init
npm run capi:sync
```
Then, validate that
- `nets.ts` has been created
- `@capi/polkadot` and `@capi/polkadot-dev` have been added to `package.json` dependencies
- `capi:sync` and `capi:serve` npm scripts have been added

### DIFF
--- a/_tasks/dnt.ts
+++ b/_tasks/dnt.ts
@@ -131,6 +131,7 @@ await Promise.all([
   fs.copy("LICENSE", path.join(outDir, "LICENSE")),
   fs.copy("Readme.md", path.join(outDir, "Readme.md")),
   fs.copy("server/static/", path.join(outDir, "esm/server/static/")),
+  writeTsLoader(path.join(outDir, "ts-loader.js")),
 ])
 
 await Promise.all([
@@ -143,7 +144,7 @@ await Promise.all([
     "target/npm/esm/main.js",
     (content) =>
       content
-        .replace(/^#!.+/, "#!/usr/bin/env -S node --loader ts-node/esm"),
+        .replace(/^#!.+/, "#!/usr/bin/env -S node --loader capi/ts-loader"),
   ),
   editFile(
     "target/npm/esm/_dnt.shims.js",
@@ -151,8 +152,27 @@ await Promise.all([
       content
         .replace(/"@deno\/shim-deno"/g, `"./deps/shims/Deno.node.js"`),
   ),
+  editFile(
+    "target/npm/package.json",
+    (content) => {
+      const packageJson = JSON.parse(content)
+      packageJson.exports["./ts-loader"] = "./ts-loader.js"
+      return JSON.stringify(packageJson, undefined, 2)
+    },
+  ),
 ])
 
 async function editFile(path: string, modify: (content: string) => string) {
   await Deno.writeTextFile(path, modify(await Deno.readTextFile(path)))
+}
+
+async function writeTsLoader(path: string) {
+  const content = `
+import tsNode from "ts-node"
+
+const service = tsNode.create({ transpileOnly: true, compilerOptions: { module: "ESNext" } })
+
+export const { resolve, load, getFormat, transformSource } = tsNode.createEsmHooks(service)
+`
+  await Deno.writeTextFile(path, content)
 }
--- a/_tasks/dnt.ts
+++ b/_tasks/dnt.ts
@@ -134,7 +134,6 @@ await Promise.all([
   fs.copy("LICENSE", path.join(outDir, "LICENSE")),
   fs.copy("Readme.md", path.join(outDir, "Readme.md")),
   fs.copy("server/static/", path.join(outDir, "esm/server/static/")),
-  writeTsLoader(path.join(outDir, "ts-loader.js")),
 ])
 
 await Promise.all([
@@ -147,7 +146,7 @@ await Promise.all([
     "target/npm/esm/main.js",
     (content) =>
       content
-        .replace(/^#!.+/, "#!/usr/bin/env -S node --loader capi/ts-loader"),
+        .replace(/^#!.+/, "#!/usr/bin/env -S node --loader ts-node/esm"),
   ),
   editFile(
     "target/npm/esm/_dnt.shims.js",
@@ -155,20 +154,8 @@ await Promise.all([
       content
         .replace(/"@deno\/shim-deno"/g, `"./deps/shims/Deno.node.js"`),
   ),
-  editFile(
-    "target/npm/package.json",
-    (content) => {
-      const packageJson = JSON.parse(content)
-      packageJson.exports["./ts-loader"] = "./ts-loader.js"
-      return JSON.stringify(packageJson, undefined, 2)
-    },
-  ),
 ])
 
 async function editFile(path: string, modify: (content: string) => string) {
   await Deno.writeTextFile(path, modify(await Deno.readTextFile(path)))
-}
-
-async function writeTsLoader(path: string) {
-  await Deno.writeTextFile(path, `export * from "ts-node/esm"`)
 }
--- a/_tasks/dnt.ts
+++ b/_tasks/dnt.ts
@@ -170,12 +170,5 @@ async function editFile(path: string, modify: (content: string) => string) {
 }
 
 async function writeTsLoader(path: string) {
-  const content = `
-import tsNode from "ts-node"
-
-const service = tsNode.create({ transpileOnly: true, compilerOptions: { module: "ESNext" } })
-
-export const { resolve, load, getFormat, transformSource } = tsNode.createEsmHooks(service)
-`
-  await Deno.writeTextFile(path, content)
+  await Deno.writeTextFile(path, `export * from "ts-node/esm"`)
 }
--- a/_tasks/dnt.ts
+++ b/_tasks/dnt.ts
@@ -56,12 +56,15 @@ await Promise.all([
       description: "Capi is a framework for crafting interactions with Substrate chains",
       license: "Apache-2.0",
       repository: "github:paritytech/capi",
-      dependencies: Object.fromEntries(
-        Object.keys(nets).map((key) => {
-          const name = normalizePackageName(key)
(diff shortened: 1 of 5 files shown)


cli/init.ts
```diff
@@ -0,0 +1,94 @@
...
+  .action(runInit)
+
+async function runInit() {
+  const packageJsonPath = path.join(Deno.cwd(), "package.json")
+  if (await fs.exists(packageJsonPath)) return await runInitNode(packageJsonPath)
+  let denoJsonPath = path.join(Deno.cwd(), "deno.json")
+  if (await fs.exists(denoJsonPath)) return await runInitDeno(denoJsonPath)
+  denoJsonPath += "c"
+  if (await fs.exists(denoJsonPath)) return await runInitDeno(denoJsonPath)
+  throw new Error("Could not find neither a `package.json` nor `deno.json`/`deno.jsonc`.")
```
tjjfvi, line 19:
```suggestion
  throw new Error("Could not find a `package.json` nor a `deno.json`/`deno.jsonc`.")
```

cli/init.ts
```diff
@@ -0,0 +1,94 @@
...
+
+async function runInitDeno(denoJsonPath: string) {
+  const denoJson = parse(Deno.readTextFileSync(denoJsonPath))
+  assertManifest(denoJson)
+  const tasks = denoJson.tasks ??= {}
+  tasks["capi"] = "deno run -A http://deno.land/x/capi/main.ts"
+  tasks["capi:sync"] = "deno task capi sync node"
+  tasks["capi:serve"] = "deno task capi serve"
+  Deno.writeTextFileSync(denoJsonPath, JSON.stringify(denoJson, null, 2))
+  await Promise.all([netsInit("https://deno.land/x/capi/mod.ts", true), updateGitignore()])
```
tjjfvi, line 52: Should also be pinned

tjjfvi (changes requested):

cli/init.ts
```diff
@@ -0,0 +1,97 @@
...
+  assertManifest(packageJson)
+  const isTs = confirm("Are you using TypeScript?")
+  const devDependencies = packageJson.devDependencies ??= {}
+  if (isTs && !devDependencies["ts-node"]) {
+    const addTsNode = confirm(
+      `To define your net specs in TypeScript, we'll need to install the \`ts-node\` loader. Can we add it to your \`package.json\`?`,
+    )
+    if (addTsNode) {
+      devDependencies["ts-node"] = "*"
+    }
```
kratico, line 34: if `ts-node` is not installed, `capi` can not be run

```
> capi init

(node:40288) ExperimentalWarning: Custom ESM Loaders is an experimental feature and might change at any time
(Use `node --trace-warnings ...` to show where the warning was created)

node:internal/process/esm_loader:42
      internalBinding('errors').triggerUncaughtException(
                                ^
Error [ERR_MODULE_NOT_FOUND]: Cannot find package 'ts-node' imported from /Users/parity/Code/4fun/harry-test/
```
after manually installing `npm i ts-node`, I get

```
    const isTs = confirm("Are you using TypeScript?");
                 ^

ReferenceError: confirm is not defined
```

cli/sync.ts
```diff
@@ -7,84 +7,95 @@ import { normalizePackageName } from "../util/mod.ts"
...
-    default: "target/capi",
-  })
-  .option("-s, --server <server:string>", "")
-  .option(
-    "--runtime-config <runtimeConfig:string>",
-    "the import_map.json or package.json file path",
-  )
-  .action(runSync)
+  .command("node", syncNode)
+  .command("deno", syncDeno)
```
kratico, line 26: whit the subcommand approach, running `capi sync` does nothing.
Perhaps it should print a help message

kratico (changes requested):

cli/init.ts
```diff
@@ -0,0 +1,97 @@
...
+  }
+  const dependencies = packageJson.dependencies ??= {}
+  dependencies.capi = detectVersion() ?? "*"
+  const scripts = packageJson.scripts ??= {}
+  scripts["capi:sync"] = "capi sync node"
+  scripts["capi:serve"] = "capi serve"
+  Deno.writeTextFileSync(packageJsonPath, JSON.stringify(packageJsonPath, null, 2))
```
kratico, line 41: it should be `JSON.stringify(packageJson, null, 2)` with the updated dependencies

cli/init.ts
```diff
@@ -0,0 +1,97 @@
...
+  await Promise.all([netsInit("capi", isTs), updateGitignore()])
```
kratico, line 42: also, package json needs `"type": "module"`

cli/sync.ts
kratico: or we could have 2 commands `sync-deno` and `sync-node`; or go back to the runtime argument
```
.arguments("<runtime:runtime>")

File: cli/init.ts (hunk 3 above)
Comment (harrysolovay) [reply cli/init.ts]: I addressed this feedback, with the exception of overriding / erroring on tsconfig specifics. Not sure how we should tackle this... a bit of a can of worms.

File: cli/sync.ts (hunk 4 above)
Comment (harrysolovay) [reply cli/sync.ts:26]: > with the subcommand approach, running capi sync does nothing.

Is there an easy way to get it to print out the help text? Surprised cliffy doesn't do this by default.

File: cli/init.ts (hunk 3 above)
Comment (kratico) [reply cli/init.ts]: Perhaps running Capi without `ts-node` so we don't need a `tsconfig.json`

File: cli/sync.ts (hunk 4 above)
Comment (kratico) [reply cli/sync.ts:26]: adding an `.action` before the subcommand seems to do it
```ts
  .description("Sync net specs and update your manifest")
  .action(function() {
    console.log("help")
    this.showHelp()
  })
  .command("node", syncNode)
  .command("deno", syncDeno)
```

_tasks/dnt.ts
```diff
@@ -148,16 +149,35 @@ await Promise.all([
...
   await Deno.writeTextFile(path, modify(await Deno.readTextFile(path)))
 }
+
+async function writeTsLoader(path: string) {
+  const content = `
+import tsNode from "ts-node"
+
+const service = tsNode.create({ transpileOnly: true, compilerOptions: { module: "ESNext" } })
+
+export const { resolve, load, getFormat, transformSource } = tsNode.createEsmHooks(service)
```
kratico, line 180: with this preconfigured `ts-loader`, the `tsconfig.json` is not needed to parse the `nets.ts`

tjjfvi: The `tsconfig` could specify many important things; I don't think we should hard-code this. I like the `capi/ts-loader` approach, but I think the file should just be
```suggestion
export * from "ts-node/esm"
```

harrysolovay: This is a slippery slope. The dev may have configured TS such that they expect to be able to use other language features in their `nets.ts` (ostensibly for loading in bins / instantiating netspecs to export). The user will be frustrated if their tsconfig is ignored / if the loader results in different behavior. On the flipside, we could load in their tsconfig and override `module`... but this could be problematic as well.

@tjjfvi said the following:

> The tsconfig could specify many important things; I don't think we should hard-code this.

I think I agree with this. We'll offer good debugging info in the case that they've misconfigured.

tjjfvi: > Also, replacing with
> 
> ```
> export * from "ts-node/esm"
> ```
> 
> is the same as executing `target/npm/esm/main.js` with
> 
> ```
> #!/usr/bin/env -S node --loader ts-node/esm
> ```

Except that it doesn't require downstream users to install ts-node themselves

kratico: ok, I'll keep the custom loader with `export * from "ts-node/esm"`

and, I'll add a check in the `init` to require a `tsconfig.json` with `compilerOptions.module: ESNext`

keep in mind that the `tsconfig.json` will be required even if this is pure JS proyect.

@harrysolovay @tjjfvi Are you fine with the above?

tjjfvi: > `compilerOptions.module: ESNext`

We don't require this as of #1061 

> keep in mind that the `tsconfig.json` will be required even if this is pure JS project.

Ah, true, I forgot that we support `nets.js` files. Perhaps then, we do want a custom loader, which, if a `tsconfig.json` is present, uses `ts-node/esm`, and otherwise is a no-op.

tjjfvi: Let's defer making this loader better and keep it with `--loader ts-node/esm` for now

kratico: for `--loader ts-node/esm` to work with the `main` Capi CLI entry and a `nets.ts` file, the following is required
- the `tsconfig.json` should exist
- `compilerOptions.module` needs to be `ES*` or `Node*`
- `compilerOptions.moduleResolution` needs to be `node*` or `bundler`

if that's ok, I'll add the checks to `capi init`

tjjfvi: I think that's out of scope for `capi init`, especially since we'll be improving the loader soon.

kratico: the above is challenging if the `tsconfig.json` [extends](https://www.typescriptlang.org/tsconfig#extends) from another config

kratico: Found another issue parsing the `tsconfig.json`.
`JSON.parse` does not work when there are comments.

tjjfvi (approved):

_tasks/dnt.ts
kratico: documented concerns in https://github.com/paritytech/capi/issues/1064

kratico (approved):

PR #1026, opened 2023-05-19
ci: build examples separately with dnt

harrysolovay:
This way, the example tests will depend on the same pkg that is ultimately distributed.


ryanleecode:
superseding with #1031

PR #1028, opened 2023-05-23
chore(deps): use smoldot 1.0

lrazovic:
After migrating from https://github.com/paritytech/smoldot to https://github.com/smol-dot/smoldot, the package name has changed from `smoldot` to `smoldot2`.
This PR removes the deprecated package and replaces it with the new version.

### DIFF
--- a/_tasks/dnt.ts
+++ b/_tasks/dnt.ts
@@ -91,13 +91,13 @@ await Promise.all([
         name: "scale-codec",
         version: "0.11.2",
       },
-      "https://deno.land/x/smoldot@light-js-deno-v0.7.6/index-deno.js": {
-        name: "@substrate/smoldot-light",
-        version: "0.7.6",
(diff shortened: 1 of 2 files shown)


kratico:
@lrazovic your changes are good but the `deno-deploy` does not deploy it.
I'll keep investigating how to configure it properly.

FYI, I created a [PR](https://github.com/paritytech/capi/pull/1039) to double check how works when the deploy succeeds

lrazovic:
Thanks for the update @kratico. The culprit seems: https://github.com/paritytech/capi/blob/d9852dc803831ddde4aaae3e601f929637d580f7/server/capi.dev/delegator.ts#L108 It doesn't find any deployments so the `_getDeploymentUrl` never returns a valid url and after 1 minute of trying the Action times out

tjjfvi (approved):
Thanks for the contribution!

Looks like our CI doesn't work properly with forks (#1056), but we've tested this locally and it works like a charm, so I'll force-merge it.

PR #1031, opened 2023-05-24
feat: examples node testing

ryanleecode:
Adds testing capi in node w/ eg-ts.

Metadata example is excluded because the output exceeds Node V8 max string length.


deno.jsonc
```diff
@@ -34,7 +34,8 @@
     "dnt": "deno task run _tasks/dnt.ts",
     "test": "deno task capi serve -- deno test -A -L=info --ignore=target --parallel -r=http://localhost:4646/",
     "test:update": "deno task test -- --update",
-    "test:eg": "deno task run https://deno.land/x/egts@v0.1.0-beta.3/test.ts 'examples/**/*.eg.ts'",
+    "test:eg:deno": "deno task run https://deno.land/x/egts@v0.1.0-beta.5/main.ts test deno 'examples/**/*.eg.ts'",
```
harrysolovay, line 37: Let's have a `test:eg` task with which we can reuse `deno task run https://...`

import_map_examples.json
harrysolovay, line 1: Should this go inside of `_tasks`, given that it's utilized solely within the examples DNT build?

ryanleecode: I moved it into tasks and replaced the importMap field with `importMap: new URL("./import_map_examples.json", import.meta.url).toString(),` and updated the import map to 

```json
{
  "scopes": {
    "examples/": {
      "asserts": "../deps/std/testing/asserts.ts"
    }
  }
}
```

but it doesn't seem to work 

`Error [ERR_MODULE_NOT_FOUND]: Cannot find package 'asserts' imported from /workspaces/capi/target/npm/capi-examples/esm/examples/rune/intro.eg.js`

_tasks/dnt_import_map.json
tjjfvi, line 1: This doesn't seem to be used currently, but I think this should be used instead of the `import_map_examples.json` (which clutters the root directory).

deno.jsonc
```diff
@@ -34,7 +34,9 @@
     "dnt": "deno task run _tasks/dnt.ts",
     "test": "deno task capi serve -- deno test -A -L=info --ignore=target --parallel -r=http://localhost:4646/",
     "test:update": "deno task test -- --update",
-    "test:eg": "deno task run https://deno.land/x/egts@v0.1.0-beta.3/test.ts 'examples/**/*.eg.ts'",
+    "test:eg": "deno task run https://deno.land/x/egts@v0.1.0-beta.5/main.ts",
+    "test:eg:deno": "deno task test:eg test deno 'examples/**/*.eg.ts'",
+    "test:eg:node": "deno task test:eg test node 'target/npm/capi-examples/esm/examples/**/!(metadata)*.eg.js'",
```
tjjfvi, line 39: Rather than having this exclude hardcoded here, perhaps the `@test_skip` directive can take a scope, like `@test_skip node`.

deps/shims/Deno.node.ts
```diff
@@ -18,6 +19,9 @@ let _deno = {
     },
     toObject: () => ({ ...process.env }),
   },
+  readFileSync: (path: string | URL): Uint8Array =>
+    Buffer.from(fs.readFileSync(path, "binary"), "binary"),
```
tjjfvi, line 23: `fs.readFileSync` already returns a `Buffer`, so there's no need to use `Buffer.from` here.

tjjfvi (changes requested):

_tasks/dnt_import_map.json
ryanleecode: Can u try this patch, its not working for me

```patch
diff --git a/_tasks/dnt.ts b/_tasks/dnt.ts
index 84fa661..dd6bdea 100755
--- a/_tasks/dnt.ts
+++ b/_tasks/dnt.ts
@@ -234,7 +234,7 @@ await build({
       version: "10.7.2",
     },
   },
-  importMap: "import_map_examples.json",
+  importMap: new URL("./import_map_examples.json", import.meta.url).toString(),
   outDir: examplesOutDir,
   scriptModule: false,
   declaration: false,
diff --git a/import_map_examples.json b/_tasks/import_map_examples.json
similarity index 85%
rename from import_map_examples.json
rename to _tasks/import_map_examples.json
index d4fe332..d0f34e9 100644
--- a/import_map_examples.json
+++ b/_tasks/import_map_examples.json
@@ -1,7 +1,7 @@
 {
   "scopes": {
     "examples/": {
-      "asserts": "./deps/std/testing/asserts.ts"
+      "asserts": "../deps/std/testing/asserts.ts"
     }
   }
 }

```

examples/dev/metadata.eg.ts
```diff
@@ -8,6 +8,7 @@
  * even plain-text descriptions. This metadata serves as the basis off of which we
  * generate chain-specific APIs. Unless you are building an advanced Capi-based library,
  * chances are that you don't need to work with the metadata directly.
+ * @test_skip node
```
harrysolovay, line 11:
```suggestion
 * @test_skip
```

Let's actually skip this in all test envs

tjjfvi (dismissed):

.github/workflows/release.yml
```diff
@@ -23,11 +23,11 @@ jobs:
       - name: Build NPM Package
         run: deno task dnt --version ${{steps.get_tag_version.outputs.tag_version}} --server ${{steps.get_tag_version.outputs.tag_version}}
       - run: npm pack
```
tjjfvi, line 25: Now that the dnt task runs pack, this step shouldn't be necessary.

ryanleecode: does this look right

```patch
diff --git a/.github/workflows/release.yml b/.github/workflows/release.yml
index 15ec856..aca45c6 100644
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -22,12 +22,10 @@ jobs:
           registry-url: "https://registry.npmjs.org"
       - name: Build NPM Package
         run: deno task dnt --version ${{steps.get_tag_version.outputs.tag_version}} --server ${{steps.get_tag_version.outputs.tag_version}}
-      - run: npm pack
-        working-directory: "./target/npm/capi"
       - uses: actions/upload-artifact@v3
         with:
           name: package
-          path: "./target/npm/capi/*.tgz"
+          path: "./target/npm/artifacts/capi*.tgz"
       - uses: octokit/request-action@v2.x
         if: startsWith(github.ref, 'refs/tags/')
         with:
```

import_map_examples.json
tjjfvi: I believe the scope needs to be `"../examples/"`

.github/workflows/release.yml
tjjfvi: Sure

harrysolovay (approved):
Well done!

tjjfvi (approved):

harrysolovay (approved):

harrysolovay (approved):

tjjfvi (approved):

harrysolovay (approved):

kratico (approved):
🚀 🎉

PR #1035, opened 2023-05-31
fix: codegen ts keyword

### LINKED ISSUE #1033: Sync with a local substrate faulty codegen
When running capi sync with a local substrate generates a faulty codegen

script from package.json
```
"sync": "capi sync node --runtime-config package.json --server https://capi.dev"
```

nets.ts
```
export const local = net.ws({
  url: 'ws://127.0.0.1:9944',
});
```

https://capi.dev/@v0.1.0-beta.40/7d699789a4e7565d/local.tar
```
Error: Expected '{', got 'namespace' at types.d.ts:6063:8

  export namespace PalletCoreFellowshipCall {
         ~~~~~~~~~
    at Object.formatText (https://deno.land/x/dprint@0.2.0/mod.ts:246:17)
    at file:///src/server/codegenHandler.ts:170:39
    at eventLoopTick (ext:core/01_core.js:181:11)
    at async Promise.all (index 0)
    at async file:///src/server/codegenHandler.ts:156:7
    at async S3Cache._getRaw (file:///src/util/cache/s3.ts:18:19)
    at async handleTarball (file:///src/server/codegenHandler.ts:149:21)
    at async Server.<anonymous> (file:///src/server/errorHandler.ts:6:14)
    at async Server.#respond (https://deno.land/std@0.186.0/http/server.ts:299:18)
   ```
   
   Capi: beta 40
   Node: 20

### PR DESCRIPTION
**Pull request by kratico**
Resolves #1033 

Note:
I noticed that sometimes a request to `http://localhost:4646/hash/some-chain/types.d.ts` runs out of memory when `shiki` is doing the code rendering

```
<--- Last few GCs --->

[13812:0x140018000]   419196 ms: Mark-Compact 1398.0 (1425.3) -> 1396.8 (1425.3) MB, 310.00 / 0.00 ms  (average mu = 0.095, current mu = 0.001) allocation failure; scavenge might not succeed
[13812:0x140018000]   419508 ms: Mark-Compact 1397.8 (1425.3) -> 1396.8 (1425.3) MB, 312.58 / 0.00 ms  (average mu = 0.050, current mu = 0.001) allocation failure; scavenge might not succeed


<--- JS stacktrace --->


#
# Fatal JavaScript out of memory: Ineffective mark-compacts near heap limit
#

### DIFF
--- a/util/normalize.ts
+++ b/util/normalize.ts
@@ -21,7 +21,54 @@ export function normalizeVariableName(name: string) {
   return normalizeIdent(name).replace(/^./, (x) => x.toLowerCase())
 }
 
-const keywords = ["import"]
+// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#reserved_words
+const keywords = [
+  "break",
+  "case",
+  "catch",
+  "class",
+  "const",
+  "continue",
+  "debugger",
+  "delete",
+  "do",
+  "else",
+  "export",
+  "extends",
+  "false",
+  "finally",
+  "for",
+  "function",
+  "if",
+  "import",
+  "in",
+  "instanceof",
+  "new",
+  "null",
+  "return",
+  "super",
+  "switch",
+  "this",
+  "throw",
+  "true",
+  "try",
+  "typeof",
+  "var",
+  "void",
+  "while",
+  "with",
+  "let",
+  "static",
+  "yield",
+  "await",
+  "enum",
+  "implements",
+  "interface",
+  "package",
+  "private",
+  "protected",
+  "public",
+]
 function normalizeKeyword(ident: string) {
   return keywords.includes(ident) ? ident + "_" : ident
 }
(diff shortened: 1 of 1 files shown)


ryanleecode (approved):

util/normalize.ts
```diff
@@ -18,3 +20,8 @@ export function normalizeTypeName(name: string) {
 export function normalizeVariableName(name: string) {
   return normalizeIdent(name).replace(/^./, (x) => x.toLowerCase())
 }
+
+const keywords = ["import"]
```
tjjfvi, line 24:
```suggestion
// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#reserved_words
const keywords = [
  "break",
  "case",
  "catch",
  "class",
  "const",
  "continue",
  "debugger",
  "delete",
  "do",
  "else",
  "export",
  "extends",
  "false",
  "finally",
  "for",
  "function",
  "if",
  "import",
  "in",
  "instanceof",
  "new",
  "null",
  "return",
  "super",
  "switch",
  "this",
  "throw",
  "true",
  "try",
  "typeof",
  "var",
  "void",
  "while",
  "with",
  "let",
  "static",
  "yield",
  "await",
  "enum",
  "implements",
  "interface",
  "package",
  "private",
  "protected",
  "public",
]
```

tjjfvi (changes requested):

tjjfvi (approved):

PR #1039, opened 2023-06-01
fix: deployment

kratico:
This is a copy from https://github.com/paritytech/capi/pull/1028 to debug deployments.

It seems that external PRs are not deployed by `deno-deploy`

@harrysolovay @tjjfvi Would you know how to trigger an on-demand deploy for external contributions?

### DIFF
--- a/_tasks/dnt.ts
+++ b/_tasks/dnt.ts
@@ -91,13 +91,13 @@ await Promise.all([
         name: "scale-codec",
         version: "0.11.2",
       },
-      "https://deno.land/x/smoldot@light-js-deno-v0.7.6/index-deno.js": {
-        name: "@substrate/smoldot-light",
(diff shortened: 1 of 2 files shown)


ryanleecode:
Its probably set to only parity org to prevent spam. otherwise they could open 100000 PRs without our oversight and use up our deno deploy cost/resources

kratico:
That's correct.
I guess that we may need to run deno deploy with https://deno.com/deploy/docs/deployctl instead of using the Github integration.

I'm not sure if there a better way to run the workflow when we approve a run from a public fork https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks

Thoughts?

harrysolovay:
Closing, as this doesn't address the root problem described in #1056

Issue #928, opened 2023-04-24
CAPN: support substrate new CLI args 


kratico:
https://github.com/paritytech/substrate/pull/13384

The CLI options that are removed/or renamed are:

1.  --rpc--max--payload (replaced by --rpc--max-request-size and --rpc-max-response-size)
3. --ws--max--out-buffer-capacity
4. --ws-external (replaced by --rpc_external)
5. --unsafe--ws--external (replaced by --unsafe-rpc-external)
6. --ipc--path 
7. --ws-port (replaced by --rpc--port)
8. --ws-max--connections (replaced by --rpc--max--connections)
9. --rpc-http (replaced by --rpc-addr)
10. --rpc-ws (replaced by --rpc-addr)

PR #1040, opened 2023-06-01
feat: add support for updated substrate cli args

kratico:
Resolves https://github.com/paritytech/capi/issues/928

### DIFF
--- a/nets/DevNetSpec.ts
+++ b/nets/DevNetSpec.ts
@@ -118,7 +118,7 @@ export async function spawnDevNet({
       chainSpecPath,
       "--port",
       `${httpPort}`,
-      "--ws-port",
+      await getBinaryWsPortCliArg(binary),
       `${wsPort}`,
     ]
     if (bootnodes) {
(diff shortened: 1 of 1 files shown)


nets/DevNetSpec.ts
```diff
@@ -170,3 +170,8 @@ async function spawnNode(tempDir: string, binary: string, args: string[], signal
     }
   })
 }
+
+async function getBinaryWsPortCliArg(binary: string): Promise<string> {
```
harrysolovay, line 174:
```suggestion
async function wsPortArgName(binary: string): Promise<string> {
```

nets/DevNetSpec.ts
```diff
@@ -118,7 +118,7 @@ export async function spawnDevNet({
       chainSpecPath,
       "--port",
       `${httpPort}`,
-      "--ws-port",
+      await getBinaryWsPortCliArg(binary),
```
harrysolovay, line 121: Can await this within a `Promise.all` above (should perhaps also contain the two `getFreePort` calls)

harrysolovay (approved):

PR #1045, opened 2023-06-06
fix: MultisigRune - `depositor` as `Uint8Array`

peetzweg:
Changes type of `depositor` from `bigint` to `Uint8Array` as it's an `AccountId`.

I think this is the wrong type here in capi. It's supposed to be an `AccountId` so it should also be a Uint8Array, like everywhere else.

Screenshot 2023-06-06 at 16 16 09

https://github.com/paritytech/substrate/blob/master/frame/multisig/src/lib.rs#LL107C1-L119C2


kratico (approved):
LGTM

Issue #980, opened 2023-05-08
capi sync package json insertion ordering should be alphabetical


ryanleecode:
When running `capi sync --package-json package.json`, any dependencies are added to the bottom of the `dependencies {}` block, this should be alphabetically instead. 

Current:

```json
"dependencies": {
  "@headlessui/react": "^1.7.11",
  "@preact/signals": "^1.1.3",
  "@talisman-connect/wallets": "^1.1.3",
  "capi": "0.1.0-beta.37",
  "dotenv": "^16.0.3",
  "preact": "^10.12.1",
  "react-hook-form": "^7.43.3",
  "react-router-dom": "^6.8.1",
  "scale-codec": "0.11.2",
  "tailwind-merge": "^1.10.0",
  "vite-tsconfig-paths": "^4.0.5",
  "zod": "^3.21.4",
  "@capi/westend": "https://capi.dev/9fccbde64e0e4b10/westend.tar"
}
```

Desired:

```json
"dependencies": {
  "@capi/westend": "https://capi.dev/9fccbde64e0e4b10/westend.tar",
  "@headlessui/react": "^1.7.11",
  "@preact/signals": "^1.1.3",
  "@talisman-connect/wallets": "^1.1.3",
  "capi": "0.1.0-beta.37",
  "dotenv": "^16.0.3",
  "preact": "^10.12.1",
  "react-hook-form": "^7.43.3",
  "react-router-dom": "^6.8.1",
  "scale-codec": "0.11.2",
  "tailwind-merge": "^1.10.0",
  "vite-tsconfig-paths": "^4.0.5",
  "zod": "^3.21.4"
}
```

PR #1046, opened 2023-06-06
chore: sort sync package json deps

ryanleecode:
https://github.com/paritytech/capi/issues/980


cli/sync.ts
```diff
@@ -71,6 +71,13 @@ async function runSync({
           delete packageJson.dependencies[packageName]
         }
       }
+      packageJson.dependencies = Object.keys(packageJson.dependencies).sort().reduce(
```
harrysolovay, line 74: How about using `Object.entries`, sorting and then using `Object.fromEntries`?

harrysolovay (approved):

tjjfvi (approved):

PR #1047, opened 2023-06-07
docs: multisig

ryanleecode:
multisig pattern docs


peetzweg:
Appreciate any docs for capi, but any reasoning why [docs.capi.dev](https://github.com/paritytech/docs.capi.dev) got abandoned? IMHO these READMEs are not accessible and visible tucked away in this repo. 🤷

harrysolovay:
> any reasoning why [docs.capi.dev](https://github.com/paritytech/docs.capi.dev) got abandoned?

Think again @peetzweg ;)

@ryanleecode can you please integrate any new ideas from your MD file into the multisig example `///` comments themselves?

harrysolovay:
Seems this file contains two key insights:

1. the multisig representation can be serialized/reified
2. the structure of the virtual multisig rune

Given that we're going to be doing a lot of work on the virtual multisig rune, perhaps we should hold off on 2.

As for 1: it would be great if there was instead an example that illustrated this flow. @ryanleecode can you please create the following two examples: #1124, #1125

PR #1059, opened 2023-06-12
fix: `tryGetEnv` instead of direct access

### LINKED ISSUE #1058: replace `Deno.env.get` with a node-shimmed `tryGetEnv`

harrysolovay:
Fixes #1058


util/tryGetEnv.node.ts
```diff
@@ -0,0 +1,9 @@
+declare global {
+  interface Process {
+    env: Record<string, string>
+  }
+  export const process: Process
+}
+export function tryGetEnv(name: string): string | undefined {
+  return process.env?.[name]
```
tjjfvi, line 8:
```suggestion
  return typeof process !== "undefined" ? process.env?.[name] : undefined
```

util/tryGetEnv.node.ts
```diff
@@ -0,0 +1,9 @@
+declare global {
+  interface Process {
+    env: Record<string, string>
+  }
+  export const process: Process
```
tjjfvi, line 5:
```suggestion
  export const process?: Process
```

server/client/detectServer.ts
```diff
@@ -1,5 +1,7 @@
+import { tryGetEnv } from "../../util/mod.ts"
```
kratico, line 1: Importing this from `util/mod.ts` will bring into a web-bundle `util/compression.ts`.
Then, `util/compression.ts` will bring `import { Buffer } from "../deps/std/streams.ts"`

I'm not sure if `compression.ts` should be excluded from `util/mod.ts`, or, if this import should be re-written to `import { tryGetEnv } from "../../util/tryGetEnv.ts"`

util/tryGetEnv.node.ts
```diff
@@ -0,0 +1,9 @@
+declare global {
+  interface Process {
+    env: Record<string, string>
+  }
+  export const process: Process | undefined
+}
+export function tryGetEnv(name: string): string | undefined {
+  return typeof process !== "undefined" ? process.env?.[name] : undefined
```
harrysolovay, line 8: @tjjfvi why this instead of `process?.env?.[name]`?

tjjfvi: `process?.env?.[name]` will error if `process` is not defined, but this will not.

Issue #1067, opened 2023-06-13
node errors in `Trace.ts`


ryanleecode:
Likely introduced in #1062

```
Failed (7/37) "target/npm/capi-examples/esm/examples/tx/balances_transfer.eg.js"
file:///home/runner/work/capi/capi/target/npm/capi-examples/node_modules/capi/esm/rune/Trace.js:65
    globalThis.addEventListener("error", (e) => {
               ^

TypeError: globalThis.addEventListener is not a function
(Use `node --trace-uncaught ...` to show where the exception was thrown)
```

Suggested fix:

```patch
diff --git a/deps/shims/Deno.node.ts b/deps/shims/Deno.node.ts
index 2a66799..501c093 100644
--- a/deps/shims/Deno.node.ts
+++ b/deps/shims/Deno.node.ts
@@ -3,6 +3,7 @@ import { Command } from "./command.node.ts"
 
 declare const process: { env: Record<string, string> }
 let _deno = {
+  isShim: true,
   errors: {
     NotFound: class NotFound extends Error {},
     AlreadyExists: class AlreadyExists extends Error {},
diff --git a/rune/Trace.ts b/rune/Trace.ts
index 8786ecd..2fe2168 100644
--- a/rune/Trace.ts
+++ b/rune/Trace.ts
@@ -70,7 +70,7 @@ Error.prepareStackTrace = (err, sites) => {
 }
 
 // Force deno to use the custom prepareStackTrace when an error isn't handled
-if ("Deno" in globalThis) {
+if ("Deno" in globalThis && !("isShim" in globalThis["Deno"])) {
   globalThis.addEventListener("error", (e) => {
     handler(e.error)
   })
```

After applying patch there is new error

```
Failed (34/37) "target/npm/capi-examples/esm/examples/misc/indices.eg.js"
file:///home/ryan/Documents/Repositories/capi/target/npm/capi-examples/node_modules/capi/esm/rune/Trace.js:163
        if (site.functionName()) {
                 ^

TypeError: site.functionName is not a function
(Use `node --trace-uncaught ...` to show where the exception was thrown)

Node.js v20.1.0
```

PR #1068, opened 2023-06-13
fix: revert "feat: rework trace (#1062)"

ryanleecode:
This reverts commit 47c4292b367a7c1dc920070eff194642a6b8db7b.

See #1067


harrysolovay:
The trace changes are improvements... I don't think reverting is the right move vs. addressing what's broken. @tjjfvi please prioritize #1067.

Issue #1053, opened 2023-06-09
Bundle size too large due to dev users


statictype:
~3.5MB of the minified and gzipped production bundle size comes from dev users public keys artifact, without actually importing any dev users in the project. Looks like a regression of https://github.com/paritytech/capi/issues/951

The reasoning behind this approach of creating 100k users for testing is not clear to me.

<img width="2831" alt="Screenshot 2023-06-09 at 13 17 47" src="https://github.com/paritytech/capi/assets/42985153/6de085c4-1e3a-45d9-8e71-265c6984a4e0">

tjjfvi:
On another note, you might be happy to hear that the second-largest rectangle there (`capi_crypto_wrappers.generated.js`) will be gone as of #1060

statictype:
i just wonder what was the issue you encountered which led you to conclude there is a need for many different users.
the default dev accounts seem to be sufficient for all testing needs i know of.
is restarting the dev chain not possible? or transferring smaller amounts to prevent running out of balance? set the initial balance of default dev accounts to max `u128`? sudo set balance before all tests?

tjjfvi:
If two different tests running in parallel use the same test users, they will attempt to send two extrinsics with the same nonce, which will result in one of the two failing.

statictype:
finding a way to increment the nonce sounds like a better way to go than generating 100k test users, imho
i am sure this is possible.

Issue #1038, opened 2023-06-01
Improve web bundler integration


kratico:
Using Capi with a web bundler requires extra configuration to shim/polyfill the Capi dnt npm module.
One reason is that dnt targets `node`.
Another reason is that `capi` have the `capi` `cli` (`serve`/`sync`) in the same npm module.

Perhaps, there could be 2 npm packages `capi` and `capi-cli` (as suggested by @peetzweg).

For example, for `capi@0.1.0-beta40` and `vite` this config is used

```
import { defineConfig } from "vite";

export default defineConfig({
  build: {
    target: "esnext",
    rollupOptions: {
      shimMissingExports: true,
    },
  },
  optimizeDeps: {
    esbuildOptions: { target: "es2022" },
  },
  define: {
    "process.env.CAPI_SERVER": process.env.CAPI_SERVER,
    "process.env.CAPI_TARGET": process.env.CAPI_TARGET,
  },
});
```

Note: for a vite prod build, the `process` global var needs to be shimmed differently because [vite define](https://vitejs.dev/config/shared-options.html#define) define entries as globals during dev but it statically replace them during build.

Something similar is required for `webpack`

```
import webpack from 'webpack'

export default {
  // ...
  resolve: {
    fallback: {
      crypto: 'crypto-browserify',
      stream: 'stream-browserify',
    }
  },
  externals: {
    "node:net": {},
    "node:stream": {},
    "node:fs/promises": {},
  },
  experiments: {
    topLevelAwait: true,
  },
  plugins: [
    new webpack.DefinePlugin({
      process: {
        env: {
          CAPI_SERVER: JSON.stringify(process.env.CAPI_SERVER),
          CAPI_TARGET: JSON.stringify(process.env.CAPI_TARGET),
        }
      }
    }),
  ],
};

```

kratico:
I would say that `dnt` transpiles TS and rewrites imports.
That seems to be fine.

I think that the issue is that the main Capi entry point is transitive depedent on a some `node` requires.
As far as I know, those `node` requires are used in the Capi CLI (`sync`/`serve`).
So the solution could be to identify that shared dependency and make sure that it doesn't land in the main Capi entry point dependencies.

So far, based on the `vite` build output, it seems that the dependencies are
```
node:http > Missing export "createServer" has been shimmed in module "__vite-browser-external".
node:net > Missing export "createConnection" has been shimmed in module "__vite-browser-external".
[...]

PR #1070, opened 2023-06-14
feat: improve bundle size

kratico:
Closes https://github.com/paritytech/capi/issues/1053 https://github.com/paritytech/capi/issues/1038

This PR improves 
- the bundle size for web
- `webpack` configuration because `externals["node:*"]` is no longer required
- `vite` configuration because `build.rollupOptions.shimMissingExports` is no longer required  

Webpack Bundle Analyzer before and after this change

image

image

`vite-bundle-visualizer` after this change

image

### DIFF
--- a/mod.ts
+++ b/mod.ts
@@ -6,6 +6,7 @@ if (CAPI_MARKER in globalThis) {
 
 export * as $ from "./deps/scale.ts"
 export { BitSequence } from "./deps/scale.ts"
+export * from "./server/client/createDevUsers.ts"
 export * from "./server/client/detectConnect.ts"
 
 // moderate --exclude main.ts nets.ts util server
--- a/mod.ts
+++ b/mod.ts
@@ -6,14 +6,13 @@ if (CAPI_MARKER in globalThis) {
 
 export * as $ from "./deps/scale.ts"
 export { BitSequence } from "./deps/scale.ts"
-export * from "./server/client/mod.ts"
+export * from "./server/client/detectConnect.ts"
 
 // moderate --exclude main.ts nets.ts util server
 
 export * from "./crypto/mod.ts"
 export * from "./fluent/mod.ts"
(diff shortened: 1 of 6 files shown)


mod.ts
```diff
@@ -6,14 +6,13 @@ if (CAPI_MARKER in globalThis) {
...
 
 // moderate --exclude main.ts nets.ts util server
 
 export * from "./crypto/mod.ts"
 export * from "./fluent/mod.ts"
 export * from "./frame_metadata/mod.ts"
-export * from "./nets/mod.ts"
 export * from "./rpc/mod.ts"
 export * from "./rune/mod.ts"
 export * from "./scale_info/mod.ts"
```
kratico, line 18: `./nets/mod.ts` and `./server/client/mod.ts` have functions that close over the `devUserPublicKeysData`.

I believe that the main `esm` entry should be optimized for the browser.
If so, `./nets/mod.ts` is only needed for the `nets.ts` file (which runs in node/deno) and the `import` could import `capi/nets` instead of `capi`.

Then, from `./server/client/mod.ts` only `detectConnect.ts` is needed for the browser by the target chain `connection.js`.

@harrysolovay @tjjfvi WDYT?

mod.ts
```diff
@@ -6,14 +6,14 @@ if (CAPI_MARKER in globalThis) {
 
 export * as $ from "./deps/scale.ts"
 export { BitSequence } from "./deps/scale.ts"
-export * from "./server/client/mod.ts"
+export * from "./server/client/createDevUsers.ts"
+export * from "./server/client/detectConnect.ts"
 
 // moderate --exclude main.ts nets.ts util server
```
tjjfvi, line 12:
```suggestion
// moderate --exclude main.ts nets.ts util server nets
```

tjjfvi (approved):
Looks good, but this won't successfully merge until #1080 is merged

PR #1075, opened 2023-06-14
fix: get rid of chainspec artifact in smoldot example

harrysolovay:
Will help us avoid crossing the deno.land/x size limit

### DIFF
--- a/examples/smoldot/fetch_chainspec.eg.ts
+++ b/examples/smoldot/fetch_chainspec.eg.ts
@@ -14,5 +14,5 @@ const chainSpec = await polkadot.connection.call("sync_state_genSyncSpec", true)
 /// We'll print out the chainspec here. This can be written into a file for later use.
 console.log(chainSpec)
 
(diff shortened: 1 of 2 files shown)


examples/smoldot/fetch_chainspec.eg.ts
```diff
@@ -12,4 +12,7 @@ import { polkadot } from "@capi/polkadot"
 const chainSpec = await polkadot.connection.call("sync_state_genSyncSpec", true).run()
 
 /// We'll print out the chainspec here. This can be written into a file for later use.
-console.log(JSON.stringify(chainSpec, null, 2))
+console.log(chainSpec)
+
+// Let's export it for use by the `smoldot.eg.ts` example.
```
tjjfvi, line 17:
```suggestion
/// Let's export it for use by the `smoldot.eg.ts` example.
```

tjjfvi (approved):

PR #1086, opened 2023-06-15
fix: make npm pkg nets importable

harrysolovay:
I think it makes sense for a file to take precedence over a same-named dir's `mod.ts`, hence, let's just exclude the `nets.ts` from being matched by the fs walker.


kratico (approved):
LGTM

tjjfvi (approved):

PR #1115, opened 2023-06-29
fix: capi node on TS/JS & CJS/ESM

tjjfvi:
One big change: to properly support CJS, we need to emit *only* CJS, as only one instance of Capi should ever be loaded.

Additionally, this has a custom require-hook & esm-loader setup that only invokes ts-node if resolving a `nets.ts`.

Resolves #1112


tjjfvi:
@ryanleecode Can you confirm this works in Vite / misc workspace setups?

harrysolovay (approved):
LGTM

PR #1117, opened 2023-06-29
docs: misc. tsdoc-ing

harrysolovay:
A first pass at adding tsdoc comments to various symbols


fluent/EventsRune.ts
```diff
@@ -26,37 +28,3 @@ export interface FinalizationEventPhase {
 export interface InitializationEventPhase {
   type: "Initialization"
 }
-
-// TODO: delete this
-export type SystemExtrinsicFailedEvent = Event<{
```
harrysolovay, line 31: Moved this out of `EventsRune.ts` given that we only utilize within `ExtrinsicEventsRune.ts`

harrysolovay: I'll be cleaning this up further in a subsequent PR

PR #1156, opened 2023-07-10
feat: examples with ts-node [don't merge]

gaboesquivel:
this is a work in progress


ryanleecode:
Hi Gabo, thank you for your PR. I've converted this to a draft for now since its a WIP. What is the motivation for this PR? If it is to run examples with node, we already have a workflow running in CI that does that.

https://github.com/paritytech/capi/blob/main/.github/workflows/checks.yml#L153

Issue #1163, opened 2023-07-11
rework node compatibility message


tjjfvi:
See #1162 

Should be changed to:
> Note: The minimum supported Node version is 20.3.1. We prioritize browser support, and thus we require the [Web Crypto API](https://nodejs.org/docs/latest-v20.x/api/webcrypto.html#web-crypto-api) to be accessible from `globalThis.crypto`. Node versions 18 and below only expose this api from `require("node:crypto").webcrypto`. 
>
> If you need to use Capi with a lower version of Node, you can shim `globalThis.crypto`:
> ```ts
> globalThis.crypto = require("node:crypto").webcrypto
> ```

Should be updated in both the README and docs.capi.dev

PR #1164, opened 2023-07-11
fix: node compatibility docs

ryanleecode:
#1163

Omitted "and below", since if we go below node version 14 this api doesn't exist.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -17,7 +17,9 @@ interactions without compromising either performance or ease of use.
 npm i capi
 ```
 
-> Note: Capi depends on the standard Web Crypto API (Node 20.3.1 and above). See
+> Note: Capi depends on the standard
+> [Web Crypto API](https://nodejs.org/docs/latest-v20.x/api/webcrypto.html#web-crypto-api)
+> (**Node 20.3.1** and above). See
 > [shimming instructions](https://docs.capi.dev/setup/#web-crypto-api).
 
 <details>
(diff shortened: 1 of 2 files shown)


Readme.md
```diff
@@ -17,9 +17,18 @@ interactions without compromising either performance or ease of use.
 npm i capi
 ```
 
-> Note: The minimum supported Node version is 20, as we require the
+> Note: The minimum supported Node version is 20.3.1. We prioritize browser
```
harrysolovay, line 20: While important, this is somewhat cluttering the readme's beginning. Anything we can do to make it less intrusive? Can we try placing within a detail element, with the text "Using with Node 20.3 and below" ?

vjjft: I suggest:

`> Note: The minimum supported Node version is 20.3.1, as Capi requires the standard Web Crypto API.`

harrysolovay: Agreed.

Let's include that shim line in the docs' setup page.

```ts
globalThis.crypto = require("node:crypto").webcrypto
```

We can link to it from an additional sentence of this note: `See instructions for shimming in older versions of Node.`

Thoughts?

ryanleecode: Do we like it as a details block here?

https://github.com/paritytech/docs.capi.dev/pull/40

harrysolovay: Yes we do! We can link to it as https://docs.capi.dev/setup/#requirements (updated your PR to have a section)

harrysolovay: Just pushed the following. Any tweaks we should make?

Screenshot 2023-07-13 at 8 11 16 PM

Issue #1125, opened 2023-07-03
multisig serialize+hydrate example


harrysolovay:
An example in which the multisig state is serialized and then hydrated (used to create a new multisig rune)

Can be commented like [the offline signing example](https://github.com/paritytech/capi/blob/main/examples/sign/offline.eg.ts).

ryanleecode:
this should be blocked by https://github.com/paritytech/capi/issues/976

PR #1165, opened 2023-07-11
chore: multisig serialize and hydrate example

ryanleecode:
Closes: #1125

### DIFF
--- a/examples/multisig/hydrate.eg.ts
+++ b/examples/multisig/hydrate.eg.ts
@@ -13,24 +13,23 @@ import { MultisigRune } from "capi/patterns/multisig"
 
 const { alexa, billy, carol } = await createDevUsers()
 
-const multisigParams = {
+/// Create the multisig.
+const multisig = MultisigRune.from(polkadotDev, {
   signatories: [alexa, billy, carol].map(({ publicKey }) => publicKey),
   threshold: 2,
-}
-/// Create the multisig. Extract the hex.
-const multisig = MultisigRune.from(polkadotDev, multisigParams)
+})
 
-assertEquals(await multisig.run(), multisigParams)
-
+/// Serialize the multisig's state into a hex for later use.
 const hex = await multisig.hex.run()
+
 /// Save `hex` however you'd like (i.e. writing to disk, etc.).
 save(hex)
 
-// Hydrate the multisig
+// Hydrate a multisig with the hex from before.
 const hydratedMultisig = MultisigRune.fromHex(polkadotDev, hex)
 
 // Test hydration
-assertEquals(await hydratedMultisig.run(), multisigParams)
+assertEquals(await multisig.run(), await hydratedMultisig.run())
 
 // hide-start
 // The following noop is solely for explanation. Swap this out with your
--- a/examples/multisig/hydrate.eg.ts
+++ b/examples/multisig/hydrate.eg.ts
@@ -2,8 +2,7 @@
  * @title Multisig Hydration
  * @description Creates a multisig and encodes it into a hex string. This hex
  * string is then used to hydrate a new MultisigRune. This is useful if you
- * have created a multisig and want to use it a later date or you want to send
- * it other signatories.
+ * want to use it a later date or if you want to send it to the other signatories.
  */
 
 import { polkadotDev } from "@capi/polkadot-dev"
@@ -11,6 +10,7 @@ import { assertEquals } from "asserts"
 import { createDevUsers } from "capi"
 import { MultisigRune } from "capi/patterns/multisig"
 
+/// Create three dev users. These users will serve as signatories of the multisig.
 const { alexa, billy, carol } = await createDevUsers()
 
 /// Create the multisig.
@@ -25,14 +25,15 @@ const hex = await multisig.hex.run()
 /// Save `hex` however you'd like (i.e. writing to disk, etc.).
 save(hex)
 
-// Hydrate a multisig with the hex from before.
+/// Hydrate a multisig with the hex from before.
 const hydratedMultisig = MultisigRune.fromHex(polkadotDev, hex)
 
-// Test hydration
+/// Ensure the initially-constructed multisig state is identical to that of
+/// the hydrated multisig.
 assertEquals(await multisig.run(), await hydratedMultisig.run())
 
 // hide-start
 // The following noop is solely for explanation. Swap this out with your
-// own signed-hex-representation-consuming code.
+// own hex-representation-consuming code.
 function save(_hex: string) {}
 // hide-end
--- a/examples/multisig/hydrate.eg.ts
+++ b/examples/multisig/hydrate.eg.ts
@@ -1,8 +1,8 @@
 /**
  * @title Multisig Serialization and Hydration
- * @description Creates a multisig and encodes it into a hex string. This hex
- * string is then used to hydrate a new MultisigRune. This is useful if you
- * want to use it at a later date or if you want to send it to the other signatories.
+ * @description Create a multisig and encode its state into a hex string. Then
+ * use this hex string to hydrate a new MultisigRune. This is useful if you
+ * want to share the multisig setup info with the other signatories or a database.
  */
 
 import { polkadotDev } from "@capi/polkadot-dev"
--- a/examples/multisig/hydrate.eg.ts
+++ b/examples/multisig/hydrate.eg.ts
@@ -1,5 +1,5 @@
 /**
- * @title Multisig Hydration
+ * @title Multisig Serialization and Hydration
  * @description Creates a multisig and encodes it into a hex string. This hex
  * string is then used to hydrate a new MultisigRune. This is useful if you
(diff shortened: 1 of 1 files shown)


examples/multisig/hydrate.eg.ts
```diff
@@ -0,0 +1,36 @@
...
+/// Save `hex` however you'd like (i.e. writing to disk, etc.).
+save(hex)
+
+// Hydrate the multisig
+const multisig = MultisigRune.fromHex(polkadotDev, hex)
+
+const { signatories, threshold } = await multisig.into(ValueRune).run()
+
+// Test hydration
+assertEquals(signatories.length, 3)
```
harrysolovay, line 28: You could do an `assertEquals` on the initial (ran) and then the hydrated (ran) multisigs

examples/multisig/hydrate.eg.ts
```diff
@@ -0,0 +1,39 @@
...
+import { assertEquals } from "asserts"
+import { createDevUsers } from "capi"
+import { MultisigRune } from "capi/patterns/multisig"
+
+const { alexa, billy, carol } = await createDevUsers()
+
+const multisigParams = {
+  signatories: [alexa, billy, carol].map(({ publicKey }) => publicKey),
+  threshold: 2,
+}
```
harrysolovay, line 19: Let's inline this in the `MultisigRune.from` call

examples/multisig/hydrate.eg.ts
```diff
@@ -0,0 +1,39 @@
...
+/// Create the multisig. Extract the hex.
```
harrysolovay, line 20:
```suggestion
/// Create the multisig.
```

examples/multisig/hydrate.eg.ts
```diff
@@ -0,0 +1,39 @@
...
+const multisig = MultisigRune.from(polkadotDev, multisigParams)
+
+assertEquals(await multisig.run(), multisigParams)
```
harrysolovay, line 23:
```suggestion
```

examples/multisig/hydrate.eg.ts
```diff
@@ -0,0 +1,39 @@
...
+
```
harrysolovay, line 24:
```suggestion
/// Serialize the multisig's state into a hex for later use.
```

examples/multisig/hydrate.eg.ts
```diff
@@ -0,0 +1,39 @@
...
+const hex = await multisig.hex.run()
```
harrysolovay, line 25:
```suggestion
const hex = await multisig.hex.run()

```

examples/multisig/hydrate.eg.ts
```diff
@@ -0,0 +1,39 @@
...
+/// Save `hex` however you'd like (i.e. writing to disk, etc.).
+save(hex)
+
+// Hydrate the multisig
```
harrysolovay, line 29:
```suggestion
// Hydrate a multisig with the hex from before.
```

examples/multisig/hydrate.eg.ts
```diff
@@ -0,0 +1,39 @@
...
+const hydratedMultisig = MultisigRune.fromHex(polkadotDev, hex)
+
+// Test hydration
+assertEquals(await hydratedMultisig.run(), multisigParams)
```
harrysolovay, line 33:
```suggestion
assertEquals(await multisig.run(), await hydratedMultisig.run())
```

harrysolovay (changes requested):

examples/multisig/hydrate.eg.ts
```diff
@@ -0,0 +1,38 @@
...
+ * have created a multisig and want to use it a later date or you want to send
+ * it other signatories.
+ */
+
+import { polkadotDev } from "@capi/polkadot-dev"
+import { assertEquals } from "asserts"
+import { createDevUsers } from "capi"
+import { MultisigRune } from "capi/patterns/multisig"
+
+const { alexa, billy, carol } = await createDevUsers()
```
harrysolovay, line 14:
```suggestion
/// Create three dev users. These users will serve as signatories of the multisig.
const { alexa, billy, carol } = await createDevUsers()
```

examples/multisig/hydrate.eg.ts
```diff
@@ -0,0 +1,38 @@
...
+  threshold: 2,
+})
+
+/// Serialize the multisig's state into a hex for later use.
+const hex = await multisig.hex.run()
+
+/// Save `hex` however you'd like (i.e. writing to disk, etc.).
+save(hex)
+
+// Hydrate a multisig with the hex from before.
```
harrysolovay, line 28:
```suggestion
/// Hydrate a multisig with the hex from before.
```

examples/multisig/hydrate.eg.ts
```diff
@@ -0,0 +1,38 @@
...
+const hydratedMultisig = MultisigRune.fromHex(polkadotDev, hex)
+
+// Test hydration
```
harrysolovay, line 31:
```suggestion
/// Ensure the initially-constructed multisig state is identical to that of the hydrated multisig
```

examples/multisig/hydrate.eg.ts
```diff
@@ -0,0 +1,39 @@
+/**
+ * @title Multisig Hydration
```
harrysolovay, line 2:
```suggestion
 * @title Multisig Serialization and Hydration
```

examples/multisig/hydrate.eg.ts
```diff
@@ -0,0 +1,39 @@
...
+ * @description Creates a multisig and encodes it into a hex string. This hex
+ * string is then used to hydrate a new MultisigRune. This is useful if you
+ * want to use it at a later date or if you want to send it to the other signatories.
```
harrysolovay, line 5:
```suggestion
 * @description Create a multisig and encode its state into a hex string. Then use this
 * hex string to hydrate a new MultisigRune. This is useful if you want to share the
 * multisig setup info with the other signatories or a database.
```

harrysolovay: Final nit 😆

ryanleecode: Wouldn't it be better to keep the third person singulars? It sounds more descriptive -- without it they sound like instructions.

harrysolovay: Imo this / the imperative style is preferable

harrysolovay: I suppose we should decide on a consistent approach language we use, as with tsdocs... I could see it both ways. Hmm... what are your thoughts on why we would prefer the first way vs. the imperative way?

ryanleecode: > **Creates** a multisig and **encodes** its state into a hex string. Then **uses** this hex string to hydrate a new MultisigRune. This is useful if you want to share the multisig setup info with the other signatories or a database.

If we use third person singulars, it sounds like a proper description because its describing what the example is doing in the present tense.

Where as the latter sounds more like a set of instructions. For example if, I just add "please" in front:

> **Please** create a multisig and encode its state into a hex string. Then use this hex string to hydrate a new MultisigRune. This is useful if you want to share the multisig setup info with the other signatories or a database.

but this wouldn't work if they were third person singulars

> Please creates a multisig and encodes its state into a hex string. Then uses this hex string to hydrate a new MultisigRune. This is useful if you want to share the multisig setup info with the other signatories or a database.

ryanleecode: I can see the argument for the instructions style if we want to view the examples as a "problem-solution" format .

ryanleecode: Actually we'll just use the imperative format since everywhere else its already like that

examples/multisig/hydrate.eg.ts
```diff
@@ -0,0 +1,39 @@
+/**
+ * @title Multisig Serialization and Hydration
+ * @description Create a multisig and encode its state into a hex string. Then
+ * use this hex string to hydrate a new MultisigRune. This is useful if you
```
harrysolovay, line 4:
```suggestion
 * use this hex string to hydrate a new multisig. This is useful if you
```

examples/multisig/hydrate.eg.ts
```diff
@@ -0,0 +1,39 @@
...
+/// Create the multisig.
+const multisig = MultisigRune.from(polkadotDev, {
+  signatories: [alexa, billy, carol].map(({ publicKey }) => publicKey),
+  threshold: 2,
+})
+
+/// Serialize the multisig's state into a hex for later use.
+const hex = await multisig.hex.run()
+
+/// Save `hex` however you'd like (i.e. writing to disk, etc.).
```
harrysolovay, line 25:
```suggestion
/// Save `hex` however you'd like (for instance, writing it to a database).
```

harrysolovay (approved):

PR #1176, opened 2023-07-13
feat: add `ready` method to `ExtrinsicStatusRune` + processing stages example

harrysolovay:
@SaltyCucumber

Allows one to attain a promise, which upon `ready` resolves to `true` (or unhandles with a `NeverReadyError`).

```ts
const sent = westendDev.Balances
  .transfer({
    value: 12345n,
    dest: billy.address,
  })
  .signed(signature({ sender: alexa }))
  .sent()

/// Attach callbacks to the pending status-specific promises.
sent.ready().run().then(() => console.log(`Ready`))
sent.inBlock().run().then((hash) => console.log(`In block ${hash}`))
sent.finalized().run().then((hash) => console.log(`Finalized ${hash}`))
```

### DIFF
--- a/examples/tx/processing_stages.eg.ts
+++ b/examples/tx/processing_stages.eg.ts
@@ -1,0 +1,26 @@
+/**
+ * @title Handling Different Extrinsic Statuses
+ * @description Get promises corresponding to specific extrinsic statuses.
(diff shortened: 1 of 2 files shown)


SaltyCucumber:
I vote for having this ready method + not making transactionStatuses private (just in case)

